PR #25, opened 2024-11-15
feature: add parameters to subscription

elle-le:
https://github.tools.sap/CoLa/support/issues/59

## Status
- PR was not abandoned on purpose, but prio shifts
- no concrete concern on implementation side

## Outstanding
- [ ] outstanding: End to End test
- [ ] perform a local-kind test to create a resource with this feature


elle-le:
things needs to be done:
- extend e2e test for this feature
- perform a local-kind test to create a resource with this feature

mirzakopic:
I had a look at the e2e tests while they were running. The import is failing and the subscription never materialized inside the subaccount. I will try to run this locally and see what is going on. 
@elle-le Did you run tests locally to confirm it works for you? Just checking if this is possible flakiness.

elle-le:
> I had a look at the e2e tests while they were running. The import is failing and the subscription never materialized inside the subaccount. I will try to run this locally and see what is going on. @elle-le Did you run tests locally to confirm it works for you? Just checking if this is possible flakiness.

I was not able to run locally as my IDE could not pick up the .env file that I provided and was complaining of missing env

internal/utils.go
```diff
@@ -123,6 +123,26 @@ type clusterYaml struct {
 	Server              string `yaml:"server"`
 }
 
+// UnmarshalRawParameters produces a map structure from a given raw YAML/JSON input
+func UnmarshalRawMapParameters(in []byte) (map[string]map[string]interface{}, error) {
```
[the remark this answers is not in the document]
gergely-szabo-sap: No, in golang `map[string]map[string]interface{}` and `map[string]interface{}` are two different types with no possible type conversion. See [Go Playground](https://go.dev/play/p/Pr6-f93PMYY).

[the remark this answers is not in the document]
sdischer-sap: Ah ok I see the generated client requires you to have that type right. 
I thought its just about marshalling it from json, which would just work fine with this more generic type. But yes if you need to put it in a variable of that type that wouldn't work.

apis/account/v1alpha1/subscription_types.go
```diff
@@ -27,6 +28,9 @@ type SubscriptionParameters struct {
 	// PlanName to subscribe to, empty plannames are shown as "default" in cockpit, use "" instead
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="planName can't be updated once set"
 	PlanName string `json:"planName"`
+	// Subscription parameters allows you to add additional parameters
+	// +kubebuilder:validation:Optional
+	SubscriptionParameters runtime.RawExtension `json:"subscriptionParameters"`
```
maximilianbraun, line 33:
```suggestion
	SubscriptionParameters runtime.RawExtension `json:"parameters"`
```

maybe would fit better?

maximilianbraun: are

gergely-szabo-sap: Thank you, @maximilianbraun, for the comment.

The change is implemented in https://github.com/SAP/crossplane-provider-btp/pull/25/commits/145114d49d978f3c653de67c7ecefac2166de268

internal/clients/subscription/subscription_test.go
```diff
@@ -103,6 +104,14 @@ func TestSubscriptionApiHandler_CreateSubscription(t *testing.T) {
 				appName: "name1",
 				CreateSubscriptionRequestPayload: saas_client.CreateSubscriptionRequestPayload{
 					PlanName: internal.Ptr("plan2"),
+					SubscriptionParams: map[string]map[string]interface{}{
```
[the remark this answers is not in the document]
gergely-szabo-sap: Thank you, @maximilianbraun, for the comment.

This made me research, and I believe the generated BTP API code is faulty. I delivered a fix in https://github.com/SAP/crossplane-provider-btp/pull/25/commits/0ebd19a6014d0ba0584d0956bfcd383bedb167b1

test/e2e/testdata/crs/subscription/create_flow/subscription.yaml
```diff
@@ -7,5 +7,6 @@ spec:
   forProvider:
     appName: auditlog-viewer
     planName: free
+    parameters: {"key1": "value1","key2": {"key3":"value3"}}
```
maximilianbraun, line 10: i would suggest using yaml, i think this is more natural

sdischer-sap (approved):
lgtm

PR #29, opened 2024-11-18
feat: add `servicePlanUniqueIdentifier` to entitlement resource

lasse-sap:
This PR adds the `servicePlanUniqueIdentifier` parameter to the entitlement resource so that services with different host data centers can be selected, even when they have ambiguous names and plans. 

Added unit test and cleaned up Makefile.


internal/controller/account/entitlement/entitlement_test.go
```diff
@@ -90,6 +90,43 @@ func TestObserve(t *testing.T) {
 				err: errors.New(errKubeAPI),
 			},
 		},
+		"Simple Case, no unique identifier passed, api error for ambigous names": {
+			args: args{
+				kube: &test.MockClient{
+					MockStatusUpdate: noopStatusUpdate, MockList: test.NewMockListFn(nil, ListEntitlements(entitlement(withServiceName("hana-cloud"), withUniqueServicePlanIdentifier("a")), entitlement(withServiceName("hana-cloud"), withUniqueServicePlanIdentifier("b")))),
+				},
+				client: fake.MockClient{MockDescribeCluster: func(ctx context.Context, input v1alpha1.Entitlement) (*entitlement2.Instance, error) {
+					return nil, errors.New(errClientAPI)
```
sdischer-sap, line 99: This is not testing the right logic. 
Basically you just verify here that "if error comes back from client, return it from controller", which is the same as the test case above.

Since you test the internal logic in the client, I guess its find to not have additional tests here. The test cases here cover more of the interaction with other components.

sdischer-sap: Perfect would be to test the Create to make sure this new attribute is passed to the api client. But since we do not have any tests other then observe yet that would be more effort, so from my side no need to do that right now.

sdischer-sap (changes requested):
Just a bit more testing would be great. 

But already great work!

sdischer-sap (approved):
Thanks for adding those, lgtm now

PR #52, opened 2024-12-02
fix(servicemanager): improve Secret-Data error handling

MarlenKoch:
## Problem:
The error message `BindingCredentials can't be created from invalid secret data` covers too many different scenarios. 

https://github.com/SAP/crossplane-provider-btp/blob/84d2f41137efe91403b676f83524acaf14c7004a/internal/clients/servicemanager/api_plan_resolver.go#L45-L47

## Solution:
This PR introduces more specific error messages that will be returned. It improves the user experience so that users can figure out the cause of the issue.


sdischer-sap (approved):
LGTM! Good work!

PR #53, opened 2024-12-05
Provider secret special chars

MarlenKoch:
### Problem:

When the provider secret contains a special char (like "), the error message is the following: 
`cannot create new Service: failed to get BTP accounts service client.: invalid character '9' after object key:value pair`

https://github.com/SAP/crossplane-provider-btp/blob/main/btp/cisclient.go#L236-L254

### Solution:
This PR introduces a special error message in case the user credentials json can't be parsed. 

Since there are two different secrets, one of which already had a more specific error message, I added the new error message to the user credentials (and not the cis-provider-secret).


internal/clients/tfclient/tfclient.go
```diff
@@ -135,7 +136,7 @@ func TerraformSetupBuilderNoTracking(version, providerSource, providerVersion st
 
 		var userCredential btp.UserCredential
 		if err := json.Unmarshal(ServiceAccountSecretData, &userCredential); err != nil {
-			return ps, err
+			return ps, errors.Wrap(err, errCouldNotParseUserCredential)
```
sdischer-sap, line 139: Just seeing that now sorry. 
There is actually another function called `TerraformSetupBuilder` above which does pretty much the same thing. 
Probably makes sense to combined those, but for now lets just add the same error message on parsing there like here.

I will create a follow up issue for adding test cases for that.

sdischer-sap (changes requested):
Just one minor addition required. Other then that looks really good. Nice work!

sdischer-sap (approved):
lgtm, thank you!

maximiliantech (approved):
Thanks for your contribution! 🚀

PR #59, opened 2024-12-11
feat: add crd for service broker

MarlenKoch:
### Feature:
Adding a servicebroker in a subaccount: The btp terraform provider recently added a btp_subaccount_service_broker resource. Using Upjet we can provide a crossplane Kind that leverages the same capabilities.

https://registry.terraform.io/providers/SAP/btp/latest/docs/resources/subaccount_service_broker


test/e2e/subaccount_service_broker_v1alpha1_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"testing"
+
+	"github.com/sap/crossplane-provider-btp/apis/account/v1alpha1"
+	"github.com/crossplane-contrib/xp-testing/pkg/resources"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+func Test_SubaccountServiceBroker_v1alpha1(t *testing.T) {
+	t.Skip("skipping test until the feature is implemented")
```
sdischer-sap, line 34: Could you please change the message here to something like:
"Requires a working service broker to test, won't be fully implemented until promoting to v1beta1"

sdischer-sap: We discussed yesterday to create a follow up for e2e tests, but after thinking about it I think we can wait with that for now until we see more traction on the resource, so lets put a message like that to make our intentions  clear

test/e2e/subaccount_service_broker_v1alpha1_test.go
```diff
@@ -0,0 +1,61 @@
...
+	t.Parallel()
+
+	resource := resources.ResourceTestConfig{
+		Kind: "SubaccountServiceBroker",
+		Obj:  &v1alpha1.SubaccountServiceBroker{},
```
sdischer-sap, line 39: @MarlenKoch looks like there is some compiler error here coming from the generated code. 

Could you check whats wrong here by verifying against the other tests?

sdischer-sap (changes requested):
Great work. Requires some minor fixes to make the build succeed though (see comments).

test/e2e/subaccount_service_broker_v1alpha1_test.go
```diff
@@ -0,0 +1,57 @@
+/*
```
sdischer-sap, line 1: Sorry should have been more clear 
I meant removing this e2e tag just temporarily for you locally to see compiler errors. 
I needs to be there for the for the code to properly work I think.

sdischer-sap (changes requested):

sdischer-sap (approved):

PR #62, opened 2024-12-12
Bugfix: entitlements multiplan

sdischer-sap:
**Description**

This fixes a bug that caused multiple entitlements with same `serviceName` but different `planName` to get stuck in unhealthy state. It was caused by some false error handling in the observation logic. 

Along with the fix I refactored and documented the related code a little to make it more readable since it was highly confusing. Also cleaned  out some duplicated controller logic.

Unittests were added to properly validate the lookup of assigned services.


sdischer-sap:
> Is this the right behaviour? I would expect the controller to return an error in case the ServicePlan of an Entitlement could not be found, am I missing something here?

Any logic being touched here refers to the `assignedServices`. So that is just checking if the servicePlan is already set for the subaccount. The logic about what servicePlans are available to the subaccount for assignment are part of this `entitledServices`.

lasse-sap (approved):
Thanks for clarifying! LGTM

PR #105, opened 2025-02-11
chore: cleanup unused Dockerfile

mirzakopic:
#### Small cleanup task

The file in `cluster/Dockerfile`, is not needed. I think this can be removed as it was probably used in the past but left behind. 

As far as I can see only the Dockerfiles in `cluster/images` are actually required. 

Building locally without that Dockerfile results in byte-for-byte identical image. I have re-tagged the images that were build without the 3rd Docker file with `:without`. See image below for analysis with `container-diff`

image


sdischer-sap (approved):
lgtm

PR #106, opened 2025-02-13
Fix: Continous updates on KymaEnvironment due to breaking provisioning API change 

sdischer-sap:
There was a change in the provisioning api, which leads to unknown parameters not being returned in a GET call. This lead to continous drift detection and therefor continous update calls to patch this specific operator label that we previously used.

This PR removes maintenance of that label. It has outgrown its use anyway, so removing that wasn't really a much of a question.


maximilianbraun (approved):
lgtm

PR #111, opened 2025-02-20
feat: remove reconcilation of initialOrgManagers in CloudfoundryEnvironement

### LINKED ISSUE #78: [FEATURE] Remove reconcilation of `initialOrgManagers` in `CloudfoundryEnvironement`
**Is your feature request related to a problem? Please describe.**
Other then the name suggest the initialOrgManagers field in the cf environment CR, is being reconciled.

It acts as a single source of truth, meaning it removes all users that are not listed there from being a manager and thus conflicts with usages of the [OrgMember](https://pages.github.tools.sap/cloud-orchestration/browser/Providers/provider-cloudfoundry/cloudfoundry.btp.orchestrate.cloud.sap/orgmembers/v1alpha1) kind.

Also the [guide](https://pages.github.tools.sap/cloud-orchestration/docs/sap-services/btp-services/cloudfoundry/org_space#special-note-about-assigning-org-managers) reflects a different behaviour.

**Describe the solution you'd like**
The reconcilation logic of that field should be removed. The users should be assigned the role only on initial creation, there should be no drift detection after that. Ideally this is also prevented on schema level. It should be also part of the in-code documentation, so that it shows up in the CRD Browser as well.

** Follow Ups **
- [ ] Update guide to reflect those changes

dev-laky:
Closes #78


sdischer-sap (commented):
Nice work! lgtm in general 

Two points require fixing:
- crs from test should not be pushed
- there is a cf e2e test, where users are updated, this part of the test needs to be removed as well

apis/environment/v1alpha1/cfenvironment_types.go
```diff
@@ -44,6 +44,7 @@ func (u *User) String() string {
 // CfEnvironmentParameters are the configurable fields of a CloudFoundryEnvironment.
 type CfEnvironmentParameters struct {
 	// A list of users (with username/email and origin) to assign as the Org Manager role
+	// Cannot be updated after creation --> initial creation only
```
sdischer-sap, line 47: Sorry missed that before, just one more change request ;) 
Please add some XValidation validation directive for preventing updates, like here:
https://github.com/SAP/crossplane-provider-btp/blob/main/apis/account/v1alpha1/subaccount_types.go#L42

This will deny an update on that field during apply, so the user will immediatly see and error message.

sdischer-sap (changes requested):

sdischer-sap (approved):
lgtm, thanks for adding those changes

sdischer-sap:
@dev-laky feel free to merge ;)

Issue #73, opened 2024-12-18
Add InstanceName and BindingName for CloudManagement CRD


sdischer-sap:
**Problem Statement**
The `Cloudmanagement` resource currently uses `medatdata.name` as a name in BTP. 
Since the metadata should be related only to k8s internal things thats not a good design. 

 
**Solution Proposal**

PR #115, opened 2025-02-27
feat: add `serviceInstanceName` and `serviceBindingName` for CloudManagement CRD

PedroKlein:
### This PR adds the following changes

- New version of `CloudManagement` CRD with `serviceInstanceName` and `serviceBindingName`
- Adds the Update method for `serviceInstanceName` field

### Behaviour and backward compatibility 

With the crd on `v1beta1`, a `CloudManagement` MR can now define the `serviceInstanceName` and `serviceBindingName` apart from the `metadata.name` of the resource, as mapped by issue #73. The logic behind setting the name of the instance or binding is the following:

- if `serviceInstanceName` or `serviceBindingName` exists, use it to define the name of the corresponding external resource.
- if empty, use a default name (like `ServiceManager` does).
- if it is from `v1alpha1`, uses `metadata.name` for both instance name and binding name. (maintains backward compatibility)

It was also added the `Update` method for triggering the change in the instance name of `cloud-management`. This does have some implications on  backward compatibility, since changing the `metadata.name` on an eventual adoption of the resource can now update the name on the external one. The old behaviour was an error on adoption that forced `metadata.name` to be the same as the instance name on BTP.


sdischer-sap:
@PedroKlein I am really impressed by your PR. 
You really thought all the scenarios through and implemented exactly as I would. 
The only thing I think that is off is the tricky migration scenario you mentioned already. I tried sorting my thoughts with this diagram here. 
Untitled-2024-09-04-1328

This check for versions alpha vs. beta I think wouldn't work here, but I will check. If not I am having another idea that might help, but I need to confirm that first.

PedroKlein:
Hey @sdischer-sap, let me see if I get it right. On the diagram, the flow with a possible problem would be the case were I have a resource from `v1alpha1`, then I upgrade the provider and also upgrade the version the CR is using?

If so, there is indeed a unanticipated behaviour on provider update. I assumed that if I had a `v1alpha1` resource already up, I would get this version when accessing its data on the controller with `cm.TypeMeta.APIVersion`, but it seems that since the struct is now from `apisv1beta1.CloudManagement`, it will always come with the version set to `v1beta1`. So the logic to maintain the same behaviour by checking the version doesn't work! 

The migration still happens sucessfully because it renames the resource to `managed-cloud-management`.

### My test to simulate the provider upgrade

1. Apply old CRD with only `v1alpha1`
2. Run the old controller
3. Apply the `v1alpha1` cloudmanagement resource -> service instance name equals `metadata.name`
4. Stop the controller
5. Apply the new CRD (with `v1alpha1` and `v1beta1`) 
6. Run the new controller
7. Check resource events -> service instance name is now `managed-cloud-management`
```
Events:
  Type    Reason                   Age                  From                     Message
  ----    ------                   ----                 ----                     -------
  Normal  PendingExternalResource  2m (x4 over 2m12s)   managed/cloudmanagement  Waiting for external resource existence to be confirmed
  Normal  CreatedExternalResource  93s (x2 over 2m16s)  managed/cloudmanagement  Successfully requested creation of external resource
  Normal  UpdatedExternalResource  5s                   managed/cloudmanagement  Successfully requested update of external resource
```

So we can actually remove the logic for using `metadata.name` if it is from `v1alpha1`, since it is not working, and with the current behaviour it would update all old resources to have the name `managed-cloud-management`. The binding will keep the old name, since it doesn't have an update method.

sdischer-sap:
@PedroKlein Exactly, thank you for that investigation. That was exactly what I meant.
So yes the question is if we want to accept this "breaking change" of changing the instance name on provider upgrade. I mean it shouldn't actually break anything, so maybe thats just ok the way it is. 
Let me discuss that with some colleagues and maybe do some quick check whether I find a quick solution for it. I guess we can decide based on the required efforts, since the implications shouldn't be too bad here.

sdischer-sap:
@PedroKlein hope you don't mind. I've taken the liberty to commit to your PR.

So I found a way to achieve the behaviour I've scribbled down before (see [commit](https://github.com/SAP/crossplane-provider-btp/pull/115/commits/7d76f2bec8936753211b28c985e3dffc26763b54)). So basically I prevent the update in case of not specified instanceName in case it matches the metadata name. I am scratching my head about this one for a while, but I think that should only match this use case of upgrading a previously existing serviceinstance during a provider upgrade. 

Honestly I am not really sure if I like that approach, because it feels a bit hacky, but I guess I like updating instances based on a provider upgrade even less ;) Let me know what you think about that one. 

I've you feel up for it maybe you can also work on the e2e tests. Testing this migration case would be overkill I guess, but maybe you change the existing test to: 
1. create serviceinstance without name -> check for default
2. udate with instanceName -> check its changed

But feel free to deny that one. I know setting up e2e tests can be painful and running them usually takes quite some time as well.

PedroKlein:
@sdischer-sap Sure! Thanks for already implementing the solution! I will try doing the E2E to learn more about it as well, so I will take a little more time if that isn't a problem.

sdischer-sap:
@PedroKlein cool, no worries about timing. Take the time you need. 
For running e2e tests a little hint. Running the whole suite is very time consuming (around 70mins). To make it easier consider running it like this:
`go test ./test/e2e -run "^TestCloudManagementCreationFlow$" -tags=e2e -v`

Should make your feedback loop a lot faster ;)

PedroKlein:
Hey @sdischer-sap, I was only able to work on it this week. I've adjusted some unit tests and added the e2e ones. I notice that there was a [Instance](https://github.com/PedroKlein/crossplane-provider-btp/blob/fix-cloudmanagement-name/apis/account/v1beta1/cloudmanagement_types.go#L73) and [Binding](https://github.com/PedroKlein/crossplane-provider-btp/blob/fix-cloudmanagement-name/apis/account/v1beta1/cloudmanagement_types.go#L104) structs not in use, so I've filled them for more detailed status on the CR to make the e2e test easier.

internal/clients/cis/cis_tf_client_test.go
```diff
@@ -13,92 +13,129 @@ import (
...
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
 			resources, err := NewTfClient(
-				&ExternalConnectorFake{
-					tc.instanceConnectorMock,
-				},
-				&ExternalConnectorFake{
-					tc.bindingConnectorMock,
-				},
+				&ExternalConnectorFake{tc.instanceConnector},
```
sdischer-sap, line 120: I like this approach to keep the mocks lightweight as fns and wire it here.

test/e2e/cloudmanagement_test.go
```diff
@@ -27,228 +22,85 @@ import (
...
+
 				return ctx
 			},
 		).Assess(
-		"Properly delete all resources", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
-			// k8s resource cleaned up?
-			cm := &v1alpha1.CloudManagement{}
-			MustGetResource(t, cfg, cisCreateName, nil, cm)
-			AwaitResourceDeletionOrFail(ctx, t, cfg, cm, wait.WithTimeout(time.Minute*10))
+		"Check CloudManagement Resource is fully created with default values", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
```
sdischer-sap, line 67: perfect ;)

sdischer-sap (approved):
I am super sorry for having left your PR hanging for such a long time. There was a lot going on around ServiceInstances and other priorities unfortunately.  
Great work! Nothing to complain here at all!

We have a requirement to have import testing available for resources in v1beta1, but I will create a follow up issue for that one.

PR #122, opened 2025-03-28
fix: cannot create CF-ENV

dev-laky:
This PR fixes the bug which occurred after the merge of #111:
- could not create cf-env anymore because of `nil pointer dereference` in `createClientWithType` method

### Before
Used `atProvider` to initiate the client, but during create this will not be existing (at least for initial creation of resource)

### After
Using `org` data directly from the created instance, not from `status`


sdischer-sap (approved):
lgtm

Issue #81, opened 2024-12-18
[FEATURE] Add dedicated name for CF environment


sdischer-sap:
**Is your feature request related to a problem? Please describe.**
When creating a CF Environment via the BTP Provider, the 'name' attribute from 'metadata' is being used as the CF 'org-name'. This is a problem, because this 'org-name' must be unique within probably the whole global account. 
Therefore the [tutorial](https://pages.github.tools.sap/cloud-orchestration/docs/sap-services/btp-services/account-managment/environments) is not working properly out of the box.  

This requires a more solid concept first, but we can already describe here what the expectation is. 

**Describe the solution you'd like**
- [x] Removed binding between 'org-name' and the 'metadata' 'name'
- [x] Create new parameter/uuid for defining the 'org-name' and instance name 
- [x] Analysis of existing tests (Unit and Integration)
- [x] make sure schema is compatible and existing resources can be "upgraded"

**Additional context**
We might need to experiment with the api a bit to better understand whats the difference of org-name, instanceName and environment name is and how to set it.

vickysuch:
we could add orgName to forProvider, if not exist then append either random string or subaccountRef

sdischer-sap:
Just had a call with vicky and we agreed on this data model changes:

![Image](https://github.com/user-attachments/assets/6246fa28-360b-41cc-8916-72b47a8c63e8)

vickysuch:
PR: 
https://github.com/SAP/crossplane-provider-btp/pull/125
https://github.tools.sap/cloud-orchestration/cloud-orchestration.github.tools.sap/pull/880

sdischer-sap:
Released a while ago already

PR #125, opened 2025-04-03
Feature enable specify cf org

vickysuch:
## Related Issue
#81 
## Behaviour Notes
- added two property OrgName and EnvironmentName for both the cf org and cf instance
- after created, the external name will be set to the crated cf org name
- if OrgName is ommited, default value will be subaccount id + "-" + CR.name
- if EnvironmentName is ommited, then will send null value to API, default of API value will be used
- if both OrgName and EnvName is ommited, the new created org will be subaccount id + "-" + CR.name (note: this is different to the current behavior, where the org name will be the cr name)
- when Observe, filter if there's cf org name equal to CR.name(to be compatible) or equal to OrgName or equal to the defualt Org name(subaccount id + "-" + CR.name)


sdischer-sap:
@vickysuch is that one good to be reviewed? 
If so could you add some description to it?

vickysuch:
image

sdischer-sap:
Thanks, those detailed behaviour notes are great for future reference. I actually plan to add some more permanent docs for changes like that and I will later ask you to put them there. 

However when starting to review I noticed that the PR is not in line with the main branch unfortunately. There have been some changes merged meanwhile. Could you please update the branch. Hope the conflicts are not too massive here...

internal/controller/environment/cloudfoundry/cfenvironment.go
```diff
@@ -118,18 +119,21 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 	if err != nil {
 		return managed.ExternalObservation{}, err
 	}
+	if instance != nil  && *instance.State == v1alpha1.InstanceStateOk && *env.ExternalName(instance) != meta.GetExternalName(cr) {
```
sdischer-sap, line 122: That means we are changing the external-name for previously existing resources right?

Do we need to treat it that way? I guess the lookup is also able to lookup via legacy metadata.name right?

vickysuch: this is to set the external name correctly, and to keep it consistent when if the external name is provided incorrectly. The previous existing resource should not be changed since the org name should also equal to the CR's name, so it should be the same after changed to the org name?

btp/cisclient.go
```diff
@@ -293,50 +293,85 @@ func (c *Client) UpdateKymaEnvironment(ctx context.Context, environmentInstanceI
 
 func (c *Client) CreateCloudFoundryOrg(
 	ctx context.Context, instanceName string, serviceAccountEmail string, resourceUID string,
```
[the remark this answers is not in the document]
vickysuch: good point, it's actually not needed anymore, will remove it

internal/controller/environment/cloudfoundry/cfenvironment.go
sdischer-sap: Sorry I don't think I get it. 
Could you give me an example of when and how changing the external-name here is required?

vickysuch: when the CF org is created, originally the external name is set to the CR.name, since we want to change it to the real CF org name, so I added the coding here. However only based on condition when the instance already created and status is ok, meanwhile when the external name doesnt match the real org name. (we dont want to change it everytime observe).

vickysuch: Yes if the external-name provided is the org name in CF env, we also wont change the external name. Because then this condition env.ExternalName(instance) != meta.GetExternalName(cr)  will be false, so we wont change it. 
Why I add it here is after we did the creation of  CF org, then the process ends. It comes to another Observe, then here we check to set the external name. 
(We still will look up from the external name provided by the user. )

sdischer-sap: Ahhh so actually this is the regular setting of the external-name after a creation. 
Why don't you do that part in the Create(). Thats usually how its done. You also don't need to manually kube.Update() there because crossplane will take care of that.

vickysuch: ah ok, because there was originally a setExternalName in the Observe thought this is where I should do it, will check to do this in the Create method

sdischer-sap: Thank you, yes sorry wasn't aware of the previous implementation. But yes lets use this occasion to align with general best practices here.

sdischer-sap (approved):
Sorry for the back and forth. This is great now ;) 
Thanks for the effort!

PR #138, opened 2025-04-10
feat: add workflow to publish tage images of a branch

GenosseOtt:


.github/workflows/release-candidate.yaml
```diff
@@ -0,0 +1,76 @@
+# Release builds and publish a new release.
+
+# Secrets:
+# TEST_REGISTRY_USER: registry user for the container registry
+# TEST_REGISTRY_PASSWORD: password for container registry
+# Vars:
+# REGISTRY_URL: registry url for the container registry
+
+name: Publish Release Candidate
```
sdischer-sap, line 9: @maximilianbraun do you see a reason why we shouldn't still require all the other QA stages like e2e and unittests.

Just asking, I can add it if you you agree.

maximilianbraun: maybe you want to test something or debug something. Then you just want to publish the images somewhere. However, we could as well think this whole problem differently and say: instead people will publish the images on their GH username. Since publishing the RC under this org has a different trust level.

maximilianbraun: For OSPO the prefix is enough.

maximilianbraun (approved):

PR #143, opened 2025-04-15
switch to shared workflows

GenosseOtt:


maximilianbraun (approved):
lgtm, but i think we need to test it out, there are some secrets used in the shared workflow which mostlikely are not acceisble

maximilianbraun:
Merged. No e2e tests necessay.

PR #146, opened 2025-04-17
add basic testDelete for Subaccount

vickysuch:


internal/controller/account/subaccount/subaccount_test.go
```diff
@@ -722,6 +723,79 @@ func TestCreate(t *testing.T) {
...
+				cr: NewSubaccount("unittest-sa", WithStatus(v1alpha1.SubaccountObservation{SubaccountGuid: internal.Ptr("123"), })),
+				mockClient: &MockSubaccountClient{
+					returnSubaccount: &accountclient.SubaccountResponseObject{Guid: "123"},
+					mockDeleteSubaccountExecute: func(r accountclient.ApiDeleteSubaccountRequest) (*accountclient.SubaccountResponseObject, *http.Response, error) {
+						return &accountclient.SubaccountResponseObject{Guid: "123", State: "Deleting"}, &http.Response{StatusCode: 200}, nil
+					},
+				},
+			},
+			want: want{
+				err: errors.New("Deletion Pending: Current status: Deleting"),
```
sdischer-sap, line 749: I tripped over that one during our call. 
It actually looks like there is an implementation error which returns an error here always in the first reconcile run. In the next one it will resolve itself. I have actually seen this error occasionally in my cluster, but thought its an API hickup since it resolved itself. 

So yes your testing here is great, but can you create a follow up issue to fix the implementation and maybe put a comment here that this is actually wrong and requires fixing?

internal/testutils/error_helper_utils.go
```diff
@@ -6,6 +6,7 @@ import (
 
 // ContainsError While testing there is no point in mimicking wrapped error hierarchies, but we do want to distinguish check
 // whether an error is part of the stacktrace
+// when error message itself contains :, the split contains logic is not enough, add compare of error message itself
```
sdischer-sap, line 9: Yes actually I think with newer go versions some error.Is function has been introduced that makes this manually check pointless, but thats good enough for now as well.

sdischer-sap (approved):
LGTM, 
just put a comment to create a follow up issue to fix the error we found here.

Looks like working on the tests here was paid out already ;)

internal/controller/account/subaccount/subaccount_test.go
vickysuch: ah ok sure will do so, created an issue here: https://github.com/SAP/crossplane-provider-btp/issues/155,  was also wondering whats the reason to return error

Issue #83, opened 2024-12-18
[Task] Test Coverage `Subaccount`


sdischer-sap:
**Context**
The `Subaccount` was one of the first resources we added. We learned a lot since then. It could use some refactoring as well as Upgrading Test Coverage before we actually promote it to new api versions.

**Scope**
As a first step we should fully cover the controller with unittests. 
Currently there is still tests missing for:
- `Connect()`

**Stretch Goals**
The implementation isn't separated into Controller and Client, which makes it harder to maintain. Eventually we should invest into refactoring here. This can be picked up as part of this task or created as follow up.

sdischer-sap:
## Status Update

@vickysuch added the tests for the Delete() as part of #146. 
So only thing left to do for this issue is the tests for Connect()

LeonardoAgustoni:
@sdischer-sap @GenosseOtt Is this ready for work? How can I assign it to me?

sdischer-sap:
Glad to here you are interested. Yes its ready and I assigned you now. 
Think you as an external contributor can't directly assign, but feel free to just drop a comment as you did here

PR #170, opened 2025-05-09
chore: add more coverage to subaccount resource

LeonardoAgustoni:
Still about #83, just adding more coverage and making small adjustments to some tests and logic.

With this in place, I guess it makes sense to move forward with the splitting the Subaccount resource in Client and Controller.


internal/controller/account/subaccount/subaccount.go
```diff
@@ -196,9 +196,6 @@ func needsUpdate(desired apisv1alpha1.SubaccountSpec, actual apisv1alpha1.Subacc
 	if !reflect.DeepEqual(&cleanedDesired.DisplayName, cleanedActual.DisplayName) {
 		return true
 	}
-	if !reflect.DeepEqual(&cleanedDesired.Region, cleanedActual.Region) {
-		return true
-	}
```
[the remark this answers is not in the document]
sdischer-sap: Changing the subdomain is actually not supported by the API: https://api.sap.com/api/APIAccountsService/path/updateSubaccount

When reworking the resource and making it ready for `v1beta1` probably we should prevent changing the spec on schema level. But yes for now removing that is a good catch!

internal/controller/account/subaccount/subaccount_test.go
```diff
@@ -50,6 +61,146 @@ func TestObserve(t *testing.T) {
...
+		"FindSubaccountError": {
+			reason: "Get Subaccount error should reset remote state",
+			args: args{
+				cr: NewSubaccount("unittest-sa"),
+				mockAPIClient: &MockSubaccountClient{
+					returnErr: errors.New("some error"),
+				},
+			},
+			want: want{
+				o: managed.ExternalObservation{ResourceExists: false},
```
sdischer-sap, line 73: ideally we can also verify that this very error is returned from the reconciler

internal/controller/account/subaccount/subaccount_test.go
```diff
@@ -50,6 +61,146 @@ func TestObserve(t *testing.T) {
...
+			},
+		},
+		"NeedsCreationInDifferentRegion": {
```
sdischer-sap, line 76: Thats technically probably the correct thing that is happening right now, but its not really the intended use case for us, so I wouldn't test that. Since those tests also act as a developer documentation of expected behaviour, this might limit us later to do some adaptations.

internal/controller/account/subaccount/subaccount_test.go
```diff
@@ -50,6 +61,146 @@ func TestObserve(t *testing.T) {
...
+							},
+						},
+					},
+				},
+			},
+			want: want{
+				o: managed.ExternalObservation{ResourceExists: false},
+			},
+		},
+		"NeedsCreationWithDifferentSubdomain": {
```
sdischer-sap, line 112: same here

internal/controller/account/subaccount/subaccount_test.go
```diff
@@ -741,10 +1065,12 @@ func TestConnect(t *testing.T) {
...
 
 	tests := map[string]struct {
-		args args
-		want want
+		reason string
+		args   args
+		want   want
 	}{
 		"NilResource": {
+			reason: "Expect error if used with another resource type",
```
sdischer-sap, line 1073: nice additions!

internal/tracking/test/noopblocked.go
```diff
@@ -0,0 +1,34 @@
...
+	"context"
+
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	providerv1alpha1 "github.com/sap/crossplane-provider-btp/apis/v1alpha1"
+)
+
+// NoOpReferenceResolverTracker with blocked delete for testing purposes
+type NoOpReferenceResolverTrackerBlocked struct {
```
sdischer-sap, line 13: Could you maybe refactor a little to use the existing `NoOpReferenceResolverTracker` and pass in a bool for the only different outcome in the `DeleteShouldBeBlocked` call?

sdischer-sap (changes requested):
Again very nice contribution. Very thorough testing. 
Have a couple of small change requests.

internal/tracking/test/noopblocked.go
LeonardoAgustoni: Yes, makes sense, will do that!

internal/controller/account/subaccount/subaccount_test.go
LeonardoAgustoni: > Since those tests also act as a developer documentation of expected behaviour

Totally agree with this sentence! Didn't knew this wasn't actually an expected behavior.
Will remove this test.

If I understood correctly, the Region and Subdomain are necessary fields to find the subaccount via the BTP API, since the `findSubaccount` method only accepts an GUID, meaning we must first fetch all subaccounts, and filter them by region and subdomain.

If so, I agree with you that they should not be editable, because changing them will probably trigger a creation of a new subaccount, and maybe move resources to places they shouldn't be.

internal/controller/account/subaccount/subaccount_test.go
LeonardoAgustoni: Will also remove this one.

internal/controller/account/subaccount/subaccount_test.go
LeonardoAgustoni: Didn't quite get this one. Could give me some more details on how to achieve this?

sdischer-sap: Sorry my bad. The checking for errors generally is in place, so I guess my confusion about this is why an error is treated as "not found" here. So I guess this is not your fault, but something we probably should change. 

If the api returns an error I would assume it should be returned from the controller rather then just treating it as not found.

sdischer-sap (commented):
Again very nice work!
Looks like your work here helped already finding yet another unexpected behaviour of the controller. With that I would suggest to change it according to our expectation.

internal/controller/account/subaccount/subaccount_test.go
LeonardoAgustoni: Done @sdischer-sap , thanks for the review again.

internal/controller/account/subaccount/subaccount.go
```diff
@@ -112,13 +116,17 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
...
+) error {
+	subaccount, err := c.findBTPSubaccount(ctx, desiredState)
+	if err != nil {
+		resetRemoteState(desiredState)
+		return err
+	}
 	if subaccount == nil {
 		resetRemoteState(desiredState)
-		return
+		return errors.New(errSubaccountNotFound)
```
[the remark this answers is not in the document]
LeonardoAgustoni: Makes sense, so the `Not Found` actually means that the resource is still probably pending creation.

sdischer-sap (approved):
Perfect, thank you!

PR #171, opened 2025-05-09
fix: breaker to ignore non-diff

maximilianbraun:


internal/controller/environment/kyma/kymaenvironment_test.go
```diff
@@ -570,13 +566,13 @@ func TestUpdateCircuitBreakerStatus(t *testing.T) {
 				},
 				desired:    anything,
 				current:    anything,
-				diff:       "",
+				diff:       "some-diff",
```
sdischer-sap, line 569: funny that it was even tested this way ;)

sdischer-sap (approved):

internal/controller/environment/kyma/kymaenvironment.go
```diff
@@ -245,6 +245,9 @@ func updateCircuitBreakerStatus(cr *v1alpha1.KymaEnvironment, desired any, curre
 	if cr.Status.RetryStatus == nil {
 		cr.Status.RetryStatus = &v1alpha1.RetryStatus{}
 	}
+	if diff == "" {
```
sdischer-sap, line 248: So this means we keep the Retryer data including the diff up until the next update arrives. 
Is that what we want? 
I am not sure if thats not confusing. 
Either way I think we need to have another test case for that "getting back to no diff behaviour".

sdischer-sap (changes requested):

internal/controller/environment/kyma/kymaenvironment.go
sdischer-sap: Changed it to always update data

maximilianbraun (commented):
lgtm, thanks @sdischer-sap for continuing

sdischer-sap (approved):

PR #172, opened 2025-05-12
fix Observe error

vickysuch:
when there's no forProvider.DirectoryFeatures provided, the API response will have DEFAULT value hence when compare we should not directly compare the value from forProvider.DirectoryFeatures


sdischer-sap (changes requested):
Nicely done, lets add tests

sdischer-sap (approved):
Thank you, lgtm

PR #175, opened 2025-05-14
fix: remove error event for subaccount deletion

### LINKED ISSUE #155: [BUG] Error event returned during subaccount deletion
**Describe the bug**
When delete a subaccount, there's error event returned. We observed this while writting tests, see here:
https://github.com/SAP/crossplane-provider-btp/pull/146/files#r2060236009

v1.0.3

**To Reproduce**
Delete subaccount resource

**Expected behavior**
No error

**Screenshots**
N/A

**Additional context**
N/A

LeonardoAgustoni:
Resolves #155 

If the state of the Subaccount is already `DELETING` after the API call, then no error event will be created.

### DIFF
--- a/internal/controller/account/subaccount/subaccount.go
+++ b/internal/controller/account/subaccount/subaccount.go
@@ -290,6 +290,11 @@ func deleteBTPSubaccount(
 	}
 
 	deletionState := response.State
+
+	if deletionState == subaccountStateDeleting {
+		return nil
+	}
+
 	return errors.New(fmt.Sprintf("Deletion Pending: Current status: %s", deletionState))
 }
 
--- a/internal/controller/account/subaccount/subaccount_test.go
+++ b/internal/controller/account/subaccount/subaccount_test.go
@@ -871,38 +871,52 @@ func TestConnect(t *testing.T) {
 
 func TestDelete(t *testing.T) {
 	type args struct {
-		cr           resource.Managed
-		mockClient   *MockSubaccountClient
+		cr         resource.Managed
+		mockClient *MockSubaccountClient
 	}
 	type want struct {
 		err error
 	}
 	tests := map[string]struct {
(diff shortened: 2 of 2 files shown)


internal/controller/account/subaccount/subaccount.go
```diff
@@ -289,12 +284,16 @@ func deleteBTPSubaccount(
 		return errors.Wrap(err, "deletion of subaccount failed")
 	}
 
-	deletionState := response.State
-	return errors.New(fmt.Sprintf("Deletion Pending: Current status: %s", deletionState))
+	if response.State != subaccountStateDeleting {
```
sdischer-sap, line 287: I think we do not need to verify that at all. 
As long as the API did not return an error we can consider the operation to be done. 

Continuous reconciliations will verify that the resource is gone for good and remove the k8s once thats done.

internal/controller/account/subaccount/subaccount_test.go
```diff
@@ -871,38 +871,51 @@ func TestConnect(t *testing.T) {
...
+					mockDeleteSubaccountExecute: func(r accountclient.ApiDeleteSubaccountRequest) (*accountclient.SubaccountResponseObject, *http.Response, error) {
+						return &accountclient.SubaccountResponseObject{Guid: "123", State: subaccountStateDeleting}, &http.Response{StatusCode: 200}, nil
+					},
+				},
+			},
+			want: want{
+				err: nil,
+			},
+		},
+		"DeleteFailWithWrongState": {
```
sdischer-sap, line 898: So then we also would not need this test case

sdischer-sap (changes requested):
Nice getting in the right direction, I think we can even simplify a bit more (see comments)

sdischer-sap (approved):
lgtm, thank you

Issue #79, opened 2024-12-18
[BUG] Remove `Resourceusage` error log


sdischer-sap:
**Describe the bug**
We often see those errors in stakeholders clusters:
```
2024-04-04T10:59:29Z    DEBUG    events    Blocking deletion while target still exist    {"type": "Warning", "object": {"kind":"ResourceUsage","name":"2a3314ac-6f31-42ff-983e-470fb7e0d84d.da6b1384-4e92-4407-bede-0f9c11e04247","uid":"7be14b24-88b5-4221-80c1-7b3bd9abc29e","apiVersion":"account.btp.orchestrate.cloud.sap/v1alpha1","resourceVersion":"120726988"}, "reason": "UsageAccounting"}
2024-04-04T10:59:29Z    ERROR    Reconciler error    {"controller": "providerconfig/resourceusage.account.btp.orchestrate.cloud.sap", "controllerGroup": "account.btp.orchestrate.cloud.sap", "controllerKind": "ResourceUsage", "ResourceUsage": {"name":"2a3314ac-6f31-42ff-983e-470fb7e0d84d.da6b1384-4e92-4407-bede-0f9c11e04247"}, "namespace": "", "name": "2a3314ac-6f31-42ff-983e-470fb7e0d84d.da6b1384-4e92-4407-bede-0f9c11e04247", "reconcileID": "9e7ad8ba-10ab-495e-a43a-eb051e94a5f0", "error": "cannot update ResourceUsage status: resourceusages.account.btp.orchestrate.cloud.sap \"2a3314ac-6f31-42ff-983e-470fb7e0d84d.da6b1384-4e92-4407-bede-0f9c11e04247\" not found"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
    sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:329
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
    sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
    sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:227
```
Its not really harmful, but pollutes the logs massively, so we should get rid of it.

[...]

PR #180, opened 2025-05-22
fix: ResourceUsage error during its removal

sdomarecki-sap:
This PR fixes the problem with excessive logging bug described in #79 
Normally switching off logging would do the trick, but the functionality of deleting ResourceUsage finalizer was not working completely.
The code inserted ensure that when ResourceUsage is marked as deleted, its underlying finalizer and k8s resource is also deleted.


sdischer-sap:
This is not treating the issue we were looking into. 
Its simply blindly removing the the resourceUsage and thus missing its purpose. 
Will create a new PR to fix the issue.

PR #205, opened 2025-06-25
feat: add recreateOnSubscriptionFailure field to Subscription resource

gergely-szabo-sap:


internal/controller/account/subscription/subscription_test.go
```diff
@@ -268,6 +271,116 @@ func TestCreate(t *testing.T) {
 	}
 }
 
+func TestRecreateOnFailed(t *testing.T) {
```
sdischer-sap, line 274: I like this idea of testing the whole flow in integration here. Makes it more clear what the indented behaviour really is.

Issue #95, opened 2025-01-10
[FEATURE] Proper name handling for `KymaEnvironment`


sdischer-sap:
**Is your feature request related to a problem? Please describe.**
We found in an internal support issue, that the way IDs are managed for `KymaEnvironemnt` leads to a problem when the instance gets manually updated in the BTP cockpit. In this case the label that is being used to link the k8s resource to the external one gets flashed and thus the resource would no longer be reconciled. 

**Describe the solution you'd like**
I can not really tell why this has been implemented like that initially, but I do not see anything that would prevent us from properly using a GUID as external-name annotation for the connection. It should be the default approach to any crossplane instance. I suggest to implement it like that but of course ensure the backwards compatibility here. 

**Describe alternatives you've considered**
There might be other identifiers that can be used, but we need to have something in the k8s resource for identification and not vice versa.

PR #217, opened 2025-07-07
feat: support proper external-name handling for Kyma resources

Lasserich:
This PR adds proper `external-name` name handling for Kyma-environment resources using its external-name annotation. This is now equivalent to the `environment id` parameter from the BTP CLI. 

This is set once during creation and is used for look-ups in later reconciliations, allowing for an easy import of resources via this identifier.

- https://github.com/SAP/crossplane-provider-btp/issues/95


internal/clients/kymaenvironment/kymaenvironment.go
```diff
@@ -30,7 +31,7 @@ func (c KymaEnvironments) DescribeInstance(
 	ctx context.Context,
 	cr v1alpha1.KymaEnvironment,
 ) (*provisioningclient.BusinessEnvironmentInstanceResponseObject, error) {
-	environment, err := c.btp.GetEnvironmentByNameAndType(ctx, cr.Name, btp.KymaEnvironmentType())
+	environment, err := c.btp.GetEnvironmentById(ctx, meta.GetExternalName(&cr))
```
sdischer-sap, line 34: At least once we will need to do the lookup via the old method to ensure compatibility. 
We can rewrite the external-name to lookup via ID in the following loops though.

internal/controller/environment/kyma/kymaenvironment_test.go
```diff
@@ -36,6 +36,76 @@ import (
...
+		crCompareOpts []cmp.Option
+		cr            resource.Managed
+		err           error
+	}
+
+	var cases = map[string]struct {
+		args args
+		want want
+	}{
+		"KymaAlreadyExists": {
```
sdischer-sap, line 57: This is not really a use case. If the cluster already exists it should be catched in the Observe. 
However handling errors properly of course is a use case. So lets just change naming and error message here to something api error related

Lasserich:
Good catch, this unit test currently only tests that the `hasUpdate` output of `DescribeInstance` actually converts to `LateInitialize` set to true. Probably redundant since its only reason is to be assigned directly. 
Yes I tested it manually by creating a resource without setting its external name and then letting the observe run.

internal/controller/environment/kyma/kymaenvironment_test.go
```diff
@@ -205,15 +205,36 @@ func TestObserve(t *testing.T) {
...
 		},
+		"ExternalNameChanged": {
+			args: args{
+				client: fake.MockClient{MockDescribeCluster: func(ctx context.Context, input *v1alpha1.KymaEnvironment) (*provisioningclient.BusinessEnvironmentInstanceResponseObject, bool, error) {
+					return &provisioningclient.BusinessEnvironmentInstanceResponseObject{
+						State:      internal.Ptr("OK"),
+						Parameters: internal.Ptr("{\"name\":\"kyma\"}"),
+					}, true, nil
+				}},
+				cr: environment(withExternalName("CHANGED-VALUE"), withUID("1234")),
```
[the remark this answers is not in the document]
sdischer-sap: ah yes I remember. Well but then I would rather either don't set the external-name at all (neither in the args nor want) or set it different to show the indented use case (even though its not actually required due to using mocks here).

sdischer-sap (approved):
lgtm

Issue #200, opened 2025-06-20
[TASK] Add crossplane provider version to tf user-agent


sdischer-sap:
**Context**
The **btp terraform provider** supports to set a **custom user-agent**. See related [ticket](https://github.com/SAP/terraform-provider-btp/issues/622).
We discussed it with the API teams that setting a custom user agent for our use case would **allow better usage tracking**. 
The goal is to enhance the user-agent to also contain the crossplane provider version.

**Acceptance Criteria** 
- [x] Provider sends enhanced version when running as a container in k8s
- [x] Stretch: Provider sends enhanced version in dev mode as well

PR #222, opened 2025-07-10
Feat/custom user agent

sdischer-sap:
Relates to #200 

Sets a custom user agent for tf http calls as described in [issue](https://github.com/SAP/terraform-provider-btp/issues/622).

**Format** 
`crossplane/<provider_version>` (<provider_version> defaults to "dev" when running in local)
**Format of whole user agent (as defined by tf provider)**
`terraform/<tf_version> terraform-provider-btp/<tf_provider_version> custom-user-agent/crossplane/<provider_version`


Lasserich (approved):
LGTM nice work! :)

Issue #181, opened 2025-05-22
[FEATURE] Stop preventing Kyma updates depending on the status


sdischer-sap:
**Context**
The provider currently does only attempt updating the kyma environement in case the instance is in an ok state as identified via the state attribute returned from the API. Since this state reflects the last status update there are scenarios in which an **instance stays in the `UPDATING` state forever**, thus preventing any further updates.

**Scope**
Provided that the API always returns the latest state of data there is not really a point in preventing those updates. The precondition to this is that the **API immediately returns the updated data** after the update has been send. If the old data is returned while the changes are still applied **(which sometimes can take a couple of minutes)** this change might not be so easy. So this **needs to be verified as part of this ticket.**

**Acceptance criteria** 
- [ ] provider ignores the state for detecting drifts
- [ ] provider only does one update for change (otherwise we need to rethink this issue)

**Workarounds & Alternatives**
Unstuck kyma instance by applying an empty update in the cockpit (this is not always even possible)

**Additional context**
This issue came up a lot during our attempts to fix continuous updates for stakeholders.

sdischer-sap:
@GenosseOtt good to go to refinement once you approve.

Lasserich:
**I have replicated the behavior using the following Manifest:**
```yaml
---
apiVersion: environment.btp.sap.crossplane.io/v1alpha1
kind: KymaEnvironment
metadata:
  name: my-kyma-instance
spec:
  subaccountRef:
    name: lasse-kyma-subaccount
  cloudManagementRef:
    name: cis-local
  writeConnectionSecretToRef:
    name: kyma-kubeconfig-local
    namespace: default
  forProvider:
    planName: azure
    parameters:
      region: westeurope
      machineType: Standard_D2s_v5
      autoScalerMin: 3
      autoScalerMax: 3
      administrators:
      - ...@sap.com
```
**Changing these two parameters to five:**
```yaml
      autoScalerMin: 5
      autoScalerMax: 5
```
This causes exactly one [Update call](https://github.com/SAP/crossplane-provider-btp/blob/614edece599e00cde80ec5313620c957e899cabc/internal/controller/environment/kyma/kymaenvironment.go#L105)(`needsUpdate = true`) on the resource. In all subsequent calls the correct configuration is being returned from the API (`diff = ""`).

**Conclusion:**
[...]

PR #233, opened 2025-07-22
feat: fix Kyma update depending on the status

Lasserich:
This PR fixes Kyma Environments being in a dangling state because they were not updated when their state was not "OK".
- https://github.com/SAP/crossplane-provider-btp/issues/181


sdischer-sap (approved):
lgtm

PR #241, opened 2025-07-25
feat: support managing of `KymaModule`

Lasserich:
This PR adds the functionality to manage your Kyma Modules straight away in Crossplane without having to connect to the Kyma Cluster yourself. For this you can use the CR like the following:

```yaml
apiVersion: environment.btp.sap.crossplane.io/v1alpha1
kind: KymaModule
metadata:
  name: cloud-manager
spec:
  kymaEnvironmentBindingRef:
    name: my-kyma-binding-two
  forProvider:
    name: cloud-manager 
    channel: fast

```

Note: Currently we support the full set of [default and optional Kyma modules](https://help.sap.com/docs/btp/sap-business-technology-platform/kyma-modules) for `forProvider.name`.

### DIFF
--- a/.github/workflows/cron_long_e2e_test.yaml
+++ b/.github/workflows/cron_long_e2e_test.yaml
@@ -4,6 +4,10 @@ name: Long E2E Test
 
 on:
   workflow_dispatch:
+    inputs:
+      testFilter:
+        description: 'testFilter string that will be passed to make e2e-long testFilter=functionNameOfTest.'
+        required: false
   schedule:
     - cron: '10 1 * * 1-5'  # every weekday at 1:10 AM
 
@@ -53,7 +57,11 @@ jobs:
       - name: Run long e2e test
         timeout-minutes: 180
         run: |
-          make test-e2e-long
+          if [ -n "${{ github.event.inputs.testFilter }}" ]; then
+            make test-e2e-long testFilter=${{ github.event.inputs.testFilter }};
+          else
+            make test-e2e-long;
+          fi
         env:
           BUILD_ID: ${{ env.BUILD_ID }}
           CIS_CENTRAL_BINDING: ${{ secrets.CIS_CENTRAL_BINDING }}
--- a/Makefile
+++ b/Makefile
@@ -242,13 +242,19 @@ test-e2e: $(KIND) $(HELM3) build generate-test-crs
 	@UUT_CONFIG=$(BUILD_REGISTRY)/$(subst crossplane-,crossplane/,$(PROJECT_NAME)):$(VERSION) UUT_CONTROLLER=$(BUILD_REGISTRY)/$(subst crossplane-,crossplane/,$(PROJECT_NAME))-controller:$(VERSION) go test $(PROJECT_REPO)/test/... -tags=e2e -short -count=1 -timeout 30m
 	@$(OK) e2e tests passed
 
-
+#run single test-e2e-long test with <make e2e testFilter=functionNameOfTest>
 test-e2e-long: $(KIND) $(HELM3) build generate-test-crs
 	@$(INFO) running integration tests
 	@echo UUT_CONFIG=$$UUT_CONFIG
 	@echo UUT_CONTROLLER=$$UUT_CONTROLLER
-	go test -v  $(PROJECT_REPO)/test/... -tags=e2e -count=1 -test.v -timeout 240m
+	go test -v  $(PROJECT_REPO)/test/... -tags=e2e -count=1 -test.v -run '$(testFilter)' -timeout 240m 2>&1 | tee test-output.log
 	@$(OK) integration tests passed
+	@echo "===========Test Summary==========="
+	@grep -E "PASS|FAIL" test-output.log
+	@case `tail -n 1 test-output.log` in \
+     		*FAIL*) echo "❌ Error: Test failed"; exit 1 ;; \
+     		*) echo "✅ All tests passed"; $(OK) integration tests passed ;; \
+     esac
 
 #run single e2e test with <make e2e testFilter=functionNameOfTest>
 .PHONY: test-acceptance


apis/environment/v1alpha1/cfenvironment_types.go
Lasserich, line 1: This was a side quest, not related to Modules.

internal/clients/kymamodule/client.go
```diff
@@ -0,0 +1,16 @@
...
+	"context"
+
+	"github.com/sap/crossplane-provider-btp/apis/environment/v1alpha1"
+)
+
+type Client interface {
+	GetModule(ctx context.Context, moduleName string) (*v1alpha1.ModuleStatus, error)
+	EnableModule(ctx context.Context, moduleName string, moduleChannel string, customResourcePolicy string) error
+	DisableModule(ctx context.Context, moduleName string) error
+	// Does not have to be public
```
Lasserich, line 13: Should we leave these methods here then, since they need the `Client` but did not wanna give them an "uglier" method signature. (e.g. I did not like putting client in the params)

internal/clients/kymamodule/module.go
```diff
@@ -0,0 +1,148 @@
...
+// GVRKyma is the GroupVersionResource for Kyma CRs.
+// ref https://github.com/kyma-project/cli/blob/838d9b9e8506489da336bf790e4814fbe1caba0b/internal/kube/kyma/types.go#L154
+var GVRKyma = schema.GroupVersionResource{
+	Group:    "operator.kyma-project.io",
+	Version:  "v1beta2",
+	Resource: "kymas",
+}
+
+// GetDefaultKyma gets the default Kyma CR from the kyma-system namespace and cast it to the Kyma structure.
+// ref https://github.com/kyma-project/cli/blob/838d9b9e8506489da336bf790e4814fbe1caba0b/internal/kube/kyma/kyma.go#L144
```
Lasserich, line 62: Same question here is this how we should reference foreign code? Even though here I even adjusted it a little.

internal/controller/environment/kymamodule/kymamodule.go
```diff
@@ -0,0 +1,168 @@
...
+
+	// Create a dynamic client using the rest.Config
+	dynamicClient, err := dynamic.NewForConfig(restConfig)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create dynamic client")
+	}
+
+	svc := c.newServiceFn(dynamicClient)
+	return &external{
+			client:     kymamodule.NewKymaModuleClient(*svc),
```
Lasserich, line 106: This is currently failing and I have no idea why, apparently some interface can not be converted into another one... Maybe this is related to the `Resource` method that the dynamic client wants to have implemented that returns a panic rn.

internal/controller/environment/kymamodule/zz_setup.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/sap/crossplane-provider-btp/apis/environment/v1alpha1"
+	providerv1alpha1 "github.com/sap/crossplane-provider-btp/apis/v1alpha1"
+	"github.com/sap/crossplane-provider-btp/internal/clients/kymamodule"
+	"github.com/sap/crossplane-provider-btp/internal/controller/providerconfig"
+	"github.com/sap/crossplane-provider-btp/internal/tracking"
+)
+
+// Setup adds a controller that reconciles KymaModule managed resources.
+func Setup(mgr ctrl.Manager, o controller.Options) error {
+	return providerconfig.DefaultKymaModuleSetup(mgr, o, &v1alpha1.KymaModule{}, v1alpha1.KymaModuleKind, v1alpha1.KymaModuleGroupVersionKind, func(kube client.Client, usage resource.Tracker, resourcetracker tracking.ReferenceResolverTracker, newServiceFn func(dynamic dynamic.Interface) *kymamodule.KymaModuleClient) managed.ExternalConnecter {
```
Lasserich, line 20: So actually here I was exchanging the usual `DefaultSetup` with a custom `DefaultKymaModuleSetup` method which takes a different `newServiceFn` in order to connect to the Kyma Instance instead of any BTP stuff. Is this the right place or does it rather interfere with Crossplane being even able to pick up the object?

internal/controller/providerconfig/defaultsetup.go
```diff
@@ -53,6 +55,49 @@ func DefaultSetup(mgr ctrl.Manager, o controller.Options, object client.Object,
 		Complete(ratelimiter.NewReconciler(name, r, o.GlobalRateLimiter))
 }
 
+type KymaModuleConnectorFn func(
+	kube client.Client,
+	usage resource.Tracker,
+	resourcetracker tracking.ReferenceResolverTracker,
+	newServiceFn func(dynamic dynamic.Interface) *kymamodule.KymaModuleClient,
```
Lasserich, line 62: Ya same here, could make it more generic and have one Setup function. Just preferred this way out of pragmatism.

internal/clients/kymamodule/client.go
sdischer-sap: I am not sure I guess get the question. Usually you do not have unexported methods in the interface. If it can be simple methods in the implementation and then you can leave them out. No need to abstract any method on the object.

internal/controller/providerconfig/defaultsetup.go
sdischer-sap: The reason why this is suboptimal is because its opening up another place to maintain common behaviour each reconciler has. Like passing this Debug Logger or setting up exponential backoff behaviour, which comes for free with the other default implementation. It can be separated if there is a reason of course, but so far I am not sure there is one.

sdischer-sap (commented):
Put some remarks. 
I haven't checked the kyma code, so might be missing something, but I would have hoped it would be possible to approach this without borrowing so much code.

internal/controller/environment/kymamodule/kymamodule.go
Lasserich: Fixed together with Valentin.

internal/clients/kymamodule/module.go
```diff
@@ -0,0 +1,148 @@
...
+	DefaultKymaName           = "default"
+	DefaultKymaNamespace      = "kyma-system"
+)
+
+type KymaModuleClient struct {
+	dynamic dynamic.Interface
+}
+
+// Resource implements dynamic.Interface.
+func (c KymaModuleClient) Resource(resource schema.GroupVersionResource) dynamic.NamespaceableResourceInterface {
```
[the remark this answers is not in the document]
Lasserich: Dynamic is basically a low level Kubernetes client which does not need a strict definition of resources in order to fetch them from the KubeAPI server. E.g. we do not need to know the exact KymaCR structure and neither do we need to have interface methods defined for it. E.g. DeepCopy() etc. is not necessary. 

This is useful because we do not want to actually support creating KymaCR in the Crossplane cluster and rather just use their structure. After some discussion we decided to continue by using the regular kube client but with an unstructured.Unstructered object to depict this behavior.

internal/controller/kymamodule/kymamodule_test.go
```diff
@@ -0,0 +1,336 @@
...
+				client: &fake.MockKymaModuleClient{MockObserve: func(moduleCr *v1alpha1.KymaModule) (*v1alpha1.ModuleStatus, error) {
+					return &v1alpha1.ModuleStatus{}, nil
+				}},
+			},
+			want: want{
+				cr:  module(),
+				obs: managed.ExternalObservation{ResourceExists: true, ResourceUpToDate: true},
+				err: nil,
+			},
+			secretData: expiredSecretData(),
```
Lasserich, line 51: @enrico-kaack-comp this is what makes me unhappy, here I am abusing passing expired secrets to not run into another NewKymaClient call which I am unsure how to mock.

enrico-kaack-comp: resolved in a call to use interface to mock the behaviour when testing

internal/utils.go
```diff
@@ -15,6 +20,11 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
+const (
+	errExtractSecretKey     = "no Secret Found"
+	errGetCredentialsSecret = "could not get Secret data"
```
enrico-kaack-comp, line 25: consistent upercase/lowercase spelling

internal/clients/kymamodule/module.go
```diff
@@ -0,0 +1,182 @@
...
+const (
+	errKymaModuleCreateFailed = "Could not create KymaModule"
+	errKymaModuleDeleteFailed = "Could not delete KymaModule"
+	errFailedParse            = "failed to parse kubeconfig"
+	errFailedCreateClient     = "failed to create Kubernetes client"
+	DefaultKymaName           = "default"
+	DefaultKymaNamespace      = "kyma-system"
+)
+
+type Client interface {
```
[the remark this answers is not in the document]
Lasserich: We do so in every resource because when referencing it is always the folder name aka the resource name and then .Client. Guess we should be fine here.

internal/clients/kymamodule/module.go
```diff
@@ -0,0 +1,182 @@
...
+	DeleteModule(ctx context.Context, moduleName string) error
+}
+
+type KymaModuleClient struct {
+	kube client.Client
+}
+
+var _ Client = &KymaModuleClient{}
+
+// GVKKyma is the GroupVersionKind for Kyma CRs.
```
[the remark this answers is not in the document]
Lasserich: Yes not that we create it but rather query for and modify.

internal/controller/kymamodule/secretfetcher.go
```diff
@@ -0,0 +1,78 @@
...
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/sap/crossplane-provider-btp/apis/environment/v1alpha1"
+	"github.com/sap/crossplane-provider-btp/internal"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	kymaExpirationLayout    = "2006-01-02 15:04:05.999999999 -0700 MST" // go default time format
```
enrico-kaack-comp, line 14: if it is go default time format, do we dann even need to define it ourselves or do we have a constant from go that supports it?

internal/controller/kymamodule/secretfetcher.go
```diff
@@ -0,0 +1,78 @@
...
+
+	return kymaCreds, nil
+}
+
+// getKubeconfig checks if the secret contains valid kubeconfig data and is not expired
+func getKubeconfig(secret map[string][]byte) ([]byte, error) {
+
+	expirationBytes := secret[v1alpha1.KymaEnvironmentBindingExpirationKey]
+	if len(expirationBytes) == 0 {
+		// No expiration time found
```
enrico-kaack-comp, line 57: comment not necessary

internal/clients/kymamodule/module_test.go
```diff
@@ -0,0 +1,101 @@
...
+				client: &KymaModuleClient{
+					kube: &test.MockClient{MockGet: test.NewMockGetFn(errors.New("CRASH"))},
+				},
+			},
+			want: want{
+				payload: nil,
+				err:     errors.New("CRASH"),
+			},
+		},
+	}
```
enrico-kaack-comp, line 55: this test feels a bit useless. It basically tests 
```
if err := nil {
return err
}
```
What about a test that tests `	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(obj.Object, mg); err != nil {
` to check that the structured->unstructured conversion works?

internal/clients/kymamodule/module.go
```diff
@@ -0,0 +1,182 @@
...
+// CreateModule adds module to the default Kyma CR in the kyma-system namespace
+// if moduleChannel is empty it uses default channel in the Kyma CR
+// ref https://github.com/kyma-project/cli/blob/838d9b9e8506489da336bf790e4814fbe1caba0b/internal/kube/kyma/kyma.go#L212
+func (c *KymaModuleClient) CreateModule(ctx context.Context, moduleName string, moduleChannel string, customResourcePolicy string) error {
+	kymaCR, err := getDefaultKyma(ctx, c)
+	if err != nil {
+		return err
+	}
+
+	kymaCR = enableModule(kymaCR, moduleName, moduleChannel, customResourcePolicy)
```
enrico-kaack-comp, line 88: where is the test for the logic of enable/disable kyma module?

internal/controller/kymamodule/secretfetcher.go
Lasserich: Hm I am also quite confused by this, several google interviews and chat consultations suggest there is none such thing. I am free to revisit but as of now I am afraid there is none.

internal/utils.go
```diff
@@ -156,3 +166,20 @@ func hasPrefix(buf []byte, prefix []byte) bool {
 	trim := bytes.TrimLeftFunc(buf, unicode.IsSpace)
 	return bytes.HasPrefix(trim, prefix)
 }
+
+func LoadSecret(ctx context.Context, kube client.Client, name string, namespace string) (map[string][]byte, error) {
```
enrico-kaack-comp, line 170: have you checked if we have this functionality already in the codebase?

internal/clients/kymamodule/module_test.go
Lasserich: good point, will replace that. :)

internal/controller/kymamodule/secretfetcher.go
```diff
@@ -0,0 +1,78 @@
...
+
+type SecretFetcherInterface interface {
+	FetchSecret(ctx context.Context, cr *v1alpha1.KymaModule) ([]byte, error)
+}
+
+type SecretFetcher struct {
+	kube client.Client
+}
+
+func (c *SecretFetcher) FetchSecret(ctx context.Context, cr *v1alpha1.KymaModule) ([]byte, error) {
```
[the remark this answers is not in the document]
enrico-kaack-comp: valid strategy

internal/controller/kymamodule/secretfetcher.go
```diff
@@ -0,0 +1,76 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	return kymaCreds, nil
+}
+
+// getKubeconfig checks if the secret contains valid kubeconfig data and is not expired
+func getKubeconfig(secret map[string][]byte) ([]byte, error) {
```
enrico-kaack-comp, line 53: the name should reflect that it only returns a valid kubeconfig --> caller know it doesnt have to check validity anymore

internal/controller/kymamodule/secretfetcher.go
```diff
@@ -0,0 +1,76 @@
...
+
+type SecretFetcherInterface interface {
+	Fetch(ctx context.Context, cr *v1alpha1.KymaModule) ([]byte, error)
+}
+
+type SecretFetcher struct {
+	kube client.Client
+}
+
+func (c *SecretFetcher) Fetch(ctx context.Context, cr *v1alpha1.KymaModule) ([]byte, error) {
```
enrico-kaack-comp, line 33: secret fetcher returns a kubeconfig and not a secret, maybe not super nice but ok i think.

enrico-kaack-comp (approved):

internal/controller/kymamodule/secretfetcher.go
Lasserich: ahh fair point 🥲

Lasserich: then we leave as is? After the discussions about scalability I have the feeling we have to re-visit at some later point anyway.

Lasserich:
Ahh merged in main again... E2es are passing [here](https://github.com/SAP/crossplane-provider-btp/actions/runs/16961369312).

PR #263, opened 2025-08-07
deps: bulk update

maximilianbraun:


apis/v1alpha1/resourceusage_types.go
```diff
@@ -33,8 +33,6 @@ type ResourceUsage struct {
 
 	//Spec ResourceUsageSpec `json:"spec"`
 	Spec ResourceUsageSpec `json:"spec"`
-
-	Status metav1.Status `json:"status,omitempty"`
```
Lasserich, line 37: Does this not have any side effects?

internal/controller/account/cloudmanagement/cloudmanagement.go
```diff
@@ -159,6 +159,12 @@ type external struct {
 	tfClient cmclient.ITfClient
 }
 
+// Disconnect is a no-op for the external client to close its connection.
+// Since we dont need this, we only have it to fullfil the interface.
+func (c *external) Disconnect(ctx context.Context) error {
```
Lasserich, line 164: Why don't we have to Mock this Method in the `cloudmanagement_test.go` when we needed its Mock for example in the `service_manager_tf_client_test.go` [here](https://github.com/SAP/crossplane-provider-btp/pull/263/files#diff-1205537680f2290b6162c854c8c594aae961c886534c2f542380be8485aa40edR721)?

test/e2e/kyma_env_test.go
```diff
@@ -36,7 +36,7 @@ func TestKymaEnvironment(t *testing.T) {
 		Assess(
 			"Await resources to become synced",
 			func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
-				if err := resources.WaitForResourcesToBeSynced(ctx, cfg, manifestDir, wait.WithTimeout(time.Minute*50)); err != nil {
+				if err := resources.WaitForResourcesToBeSynced(ctx, cfg, manifestDir, nil, wait.WithTimeout(time.Minute*50)); err != nil {
```
Lasserich, line 39: This has been changed on package level no? Just to make sure I did not miss any changes to this method.

Lasserich (approved):
Minor remarks but overall lgtm.

internal/controller/account/cloudmanagement/cloudmanagement.go
enrico-kaack-comp: Excellent question I didnt think of, since the compiler and all tests run successful.
But digging into it, here is the explanation:
If I see this correctly in `cloudmanagement_test.go`, we dont mock the ExternalClient Interface (that would require the Disconnect method), but mocking the kube client in the external struct.
This way the normal external struct is used that fullfills the interface but we mock the internal kube client. So we can call the normal e.g. `external.Observe()` to test it and it will call not the kube client but this mock implementation.

In `service_manager_tf_client_test.go` we have the `managed.ExternalClient` in the test struct and when we mock this, we have to add the `Disconnect` since it needs to follow the interface.

Overall I think the tests are build different, as in: they are mocking different parts.

enrico-kaack-comp:
Long e2e also successful: https://github.com/SAP/crossplane-provider-btp/actions/runs/16987656638

test/e2e/kyma_env_test.go
enrico-kaack-comp: exactly, released in latest 1.6 https://github.com/crossplane-contrib/xp-testing/releases/tag/v1.6.0

internal/controller/account/cloudmanagement/cloudmanagement.go
enrico-kaack-comp: I have not yet looked into the tf tests/resources too much, I can not tell you yet unfortunately. It could also be different ways of writing tests, changed over time for mroe recent resources etc.

PR #272, opened 2025-08-14
feat(e2e-long) test filter: trigger run of single tests

enrico-kaack-comp:
Our short tests allow to specify the `go test -run` parameter to define testFilter and then only run this test.
For the e2e-long, this is also helpful, especially when working with external systems that take many minutes.

When we now trigger a manual Long E2E Test, we have an input in the gh action that is passed to the `go test -run` parameter. Also I added the test summary from the short tests to the long tests.

Sample Test Run with this parameter runs only TestDirectory here (works): https://github.com/SAP/crossplane-provider-btp/actions/runs/16961721013

### DIFF
--- a/.github/workflows/cron_long_e2e_test.yaml
+++ b/.github/workflows/cron_long_e2e_test.yaml
@@ -4,6 +4,10 @@ name: Long E2E Test
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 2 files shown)


Lasserich (approved):
lgtm

PR #275, opened 2025-08-15
fix(deps): renovate runs go mod tidy after updating

enrico-kaack-comp:
I noticed several Renovate PRs in which `make reviewable` failed because it runs `go mod tidy` while renovate doesnt run it. 
So i added this option, hopefully making it easier to upgrade dependencies without manual intervention.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -3,10 +3,19 @@
   "extends": [
     "config:recommended"
   ],
+  "postUpdateOptions": [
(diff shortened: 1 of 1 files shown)


Lasserich (approved):
lgtm 🧹

enrico-kaack-comp:
I will merge witout e2e tests as this config only affects renovate and not our coding

PR #280, opened 2025-08-19
fix(#182): bug kymaenvironment change in writeconnectionsecrettoref voids the secret

### LINKED ISSUE #182: [BUG] KymaEnvironment change in writeConnectionSecretToRef voids the secret
**Describe the bug**
When updating Kymaenvironment.spec.writeConnectionSecretToRef.name with `newValue`, a new secret (`newValue`) will be created but it is empty. 

v1.1.2

**To Reproduce**
1. Create a Kymaenvironment which successfully reconciles. Check the secret created from `.spec.writeConnectionSecretToRef.name`. It exists and is not empty. 
2. update `.spec.writeConnectionSecretToRef.name` to `newValue`
3. The secret `newValue` exists but is empty 

**Expected behavior**
.spec.writeConnectionSecretToRef should be updateable and the new secret should be not empty. 

**Screenshots**

**Additional context**

gergely-szabo-sap:
This PR fixes #182.

The idea is to update the connection details secret in each Observe step. To make this operation fast, we cache the obtained connection secret by "memoizing" the `GetConnectionDetails` function. 

This memoization procedure is facilitated by a new, generic `memoize` package.


internal/memoize/memoize.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	m.lock.Lock()
+	defer m.lock.Unlock()
+	if value == nil {
+		// value == nil indicates that we want to delete the key
+		delete(m.m, *k)
+	} else {
+		if len(m.m) >= m.keyLimit {
+			return
+		}
```
enrico-kaack-comp, line 164: In the main comment for the Map struct should be mentioned that if the Map is full, no new values are stored (could also be that it would remove the oldest entry).
Since this func doesnt return an error, I do not expect this function to fail although it silently does here. So amybe add an error return?

internal/memoize/memoize.go
```diff
@@ -0,0 +1,201 @@
...
+		return
+	}
+	k := key.GetMemoKey()
+	if k == nil {
+		// We can't invalidate a key if it has no string representation
+		return
+	}
+	m.lock.Lock()
+	defer m.lock.Unlock()
+	delete(m.m, *k)
```
enrico-kaack-comp, line 184: nit: inconsistent, you dont use defer in similar cases here

internal/memoize/memoize.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+// WithGCFrequence method sets the garbage collection frequence of a
+// Map value.
+func (m *Map[V]) WithGCFrequency(frequency time.Duration) *Map[V] {
+	m.gcFrequency = frequency
+	return m
+}
+
+// WithKeyLilmit method sets the key limit of a map value.
```
enrico-kaack-comp, line 71: Add comment: what happens when i decrease the limit below the current size? The "cache" could not take any new values until some expired/are invalidated.

internal/memoize/memoize.go
```diff
@@ -0,0 +1,201 @@
...
+// objects.
+var DefaultKeyLimit = 5000
+
+// DefaultGCFrequency sets the default garbage collection frequencye
+// for any new Map objects.
+var DefaultGCFrequency = 5 * time.Minute
+
+// MinimumGCFrequency show the minimum value of the garbage collection
+// frequency.
+const MinimumGCFrequency = 1 * time.Minute
```
enrico-kaack-comp, line 201: since this is the minimum frequency enforced by the code, isnt a more reasonable default something like 1s in case we use it fore more frequent caching? Of course can also be changed if we use it for higher frequency caching later, was just wondering.

enrico-kaack-comp (approved):

internal/memoize/memoize.go
gergely-szabo-sap: > Since this func doesnt return an error, I do not expect this function to fail although it silently does here. So amybe add an error return?

I think it's fine if the Set operation is a no-op for these cases. The recommended usage of this type will be documented in the package description.

internal/memoize/memoize.go
```diff
@@ -0,0 +1,201 @@
...
+	k := key.GetMemoKey()
+	if k == nil {
+		// We can't store a ConnectionDetails if the instance has no Labels
+		return
+	}
+	m.lock.Lock()
+	defer m.lock.Unlock()
+	if value == nil {
+		// value == nil indicates that we want to delete the key
+		delete(m.m, *k)
```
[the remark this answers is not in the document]
gergely-szabo-sap: I don't get this comment. Could you elaborate on this?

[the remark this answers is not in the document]
enrico-kaack-comp:
```suggestion
if value == nil {
		// value == nil indicates that we want to delete the key
		delete(m.m, *k)
		return
}
if len(m.m) >= m.keyLimit {
			return
}
//...
```
that feels like the normal go flow since the value == nil is an early exit condition.
But as said, absolutely nit, no need to change that.

PR #304, opened 2025-09-08
chore: fixing helm install in e2e_test workflow file

gergely-szabo-sap:
I also fixed the line endings


enrico-kaack-comp (approved):
Approved and manual E2E test run successful (https://github.com/SAP/crossplane-provider-btp/actions/runs/17544176058/job/49821700317).
Can not run e2e from PR since this one still uses the workflow definition from main (that is not working)

PR #306, opened 2025-09-08
fix: e2e-long helm install

enrico-kaack-comp:
Also apply fix https://github.com/SAP/crossplane-provider-btp/pull/304 also to e2e-long tests.


Lasserich (approved):
lgtm

PR #311, opened 2025-09-10
feat(no-fork): introduce terraform framework plugin external clients

christophrj:
closes #208 with the exception of the `SubAccountApiCredentials` API (see comment on #208).
additionally closes #211 by introducing http debug logs over terraform provider btp.

The following controllers are affected:
* `SubaccountTrustConfiguration`
* `GlobalaccountTrustConfiguration`
* `DirectoryEntitlement`
* `SubaccountServiceBroker`

Tested manually to some extend but relies mostly on successful e2e test execution.

PR #329, opened 2025-09-11
chore(btp-tf-provider): upgrade to v1.16.1

enrico-kaack-comp:
The significant change is the `DirectoryEntitlement` added field `planUniqueIdentifier`. It is optional so I dont expect it to cause troubles. But we should run upgrade tests, either manual or automatic.

With #311 covering the `DirectoryEntitlement` this should be the same test for upgrades as it is there.


Lasserich (approved):
lgtm

PR #339, opened 2025-09-16
fix(upgrade-test): run from a branch as target without release

enrico-kaack-comp:
Previously, the env variable where not correctly passed between the jobs so the test did not run the latest main provider but just the latest released version.
Now we cleanly pass the variable as output from the build job to the upgrade-test job.


Lasserich (approved):
lgtm nice catch 🚀

PR #343, opened 2025-09-17
feat: Import tests for `ServiceManager` und `CloudManagement`

### LINKED ISSUE #215: [FEATURE] Import tests for `ServiceManager` und `CloudManagement`
**Context**
We decided to have import testing as a requirement for a **v1beta1** resource. We therefor should do it for our own v1beta1 resource already now:
- `ServiceManager`
- `CloudManagement'

**Technical Steps** 
Since those resources are upjet based and we do not have a go client to prepare the instance to import, I guess the simplest approach here would be this flow:
1. Create Instance as part of the test with management policies -delete
2. Delete instance
3. Import it back in with setting `external-name`

Reading a manifiest from the e2e resources and patching its external-name on the fly might require a bit of work. Consider contributing potential helper code into the [xp-testing](https://github.com/crossplane-contrib/xp-testing) framework.

**Additional context**
this is a follow up on #115

dattito:
Closes #215 by implementing import testing for `ServiceManager` and `CloudManagement` like described in the issue.

### DIFF
--- a/test/e2e/cloudmanagement_test.go
+++ b/test/e2e/cloudmanagement_test.go
@@ -8,11 +8,13 @@ import (
 	"time"
 
 	"github.com/sap/crossplane-provider-btp/apis"
+	"github.com/sap/crossplane-provider-btp/apis/account/v1alpha1"
 	"github.com/sap/crossplane-provider-btp/apis/account/v1beta1"
 	"github.com/sap/crossplane-provider-btp/internal"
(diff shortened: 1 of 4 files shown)


enrico-kaack-comp (approved):
LGTM.
One question, why is the cloudmanagement a different approach (making API call directly) than service_manager (using the ManagmentPolicies)?

enrico-kaack-comp:
youre right, I misread the one

Issue #341, opened 2025-09-17
[BUG] Subscription: Observe only with wrong external-name panics


enrico-kaack-comp:
**Describe the bug**
When we apply a Subscription MR with ManagmenetPolicy: Observe (might also happen with other Policies) and give it a technical name that is not the correct one (e.g. only `<application-technical-name>` instead of `<application-tehcnical-name>/<plan-name>` the controller crashes:

```
E0917 07:24:36.713403       1 runtime.go:79] Observed a panic: runtime.boundsError{x:1, y:1, signed:true, code:0x0} (runtime error: index out of range [1] with length 1)
goroutine 665 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic({0x22e9ce0, 0xc0012f1c80})
	k8s.io/apimachinery@v0.29.1/pkg/util/runtime/runtime.go:75 +0x85
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
	sigs.k8s.io/controller-runtime@v0.17.0/pkg/internal/controller/controller.go:108 +0xb2
panic({0x22e9ce0?, 0xc0012f1c80?})
	runtime/panic.go:792 +0x132
github.com/sap/crossplane-provider-btp/internal/clients/subscription.splitExternalName(...)
	github.com/sap/crossplane-provider-btp/internal/clients/subscription/subscription.go:198
github.com/sap/crossplane-provider-btp/internal/clients/subscription.(*SubscriptionApiHandler).GetSubscription(0xc0009804e0, {0x2855a30, 0xc000036380}, {0xc001301950?, 0xc001334701?})
	github.com/sap/crossplane-provider-btp/internal/clients/subscription/subscription.go:116 +0x305
github.com/sap/crossplane-provider-btp/internal/controller/account/subscription.(*external).loadSubscription(0xc00083f800, {0x2855a30, 0xc000036380}, 0xc001694200)
	github.com/sap/crossplane-provider-btp/internal/controller/account/subscription/subscription.go:224 +0xba
github.com/sap/crossplane-provider-btp/internal/controller/account/subscription.(*external).Observe(0xc00083f800, {0x2855a30, 0xc000036380}, {0x2879418, 0xc001694200})
	github.com/sap/crossplane-provider-btp/internal/controller/account/subscription/subscription.go:121 +0x8f
[...]

PR #348, opened 2025-09-19
fix (#341): external name validation for subscription

enrico-kaack-comp:
Fix: #341 
The format of the `externalName` was not validated before splitting it at `/` and accessing index `0` and `1`. 
Now we return an error if the format is incorrect.


Lasserich (approved):
lgtm, nice little fix 🚀 
We would ignore having more than one slash now right?

PR #353, opened 2025-09-25
test: verify external-name in ServiceManager Tf client

### LINKED ISSUE #317: (ExternalName) ServiceManager Add Unit Test
`ServiceManager` MR already set external name and uses it for matching, but Unit Tests are missing.
Use [cloudfoundry provider](https://github.com/SAP/crossplane-provider-cloudfoundry/blob/main/internal/controller/org/controller_test.go) for inspiration on what to test. 

## Acceptance Criteria
- [ ] Create sets the external-name to the external identifier (either an id or a composite key that is used for later matching)
- [ ] Observe uses the external-name annotation as primary matching key so the initial observe would match an existing resource even if it was not created in a previous reconcile of the provider
- [ ] Unit test tests the external-name matching behaviour

rdksap:
This PR adds
- check of diff in external-names to TestConnectResources of service_manager_tf_client_test.go. 
- a test case checking the split of ServiceManager external-name when mapping native CR to Tf ones.

Closes #317


christophrj (approved):
lgtm 👍  thanks @rdksap

PR #357, opened 2025-09-29
chore: enforce minimal privilege of GITHUB_TOKEN in workflows

enrico-kaack-comp:
Potential fix for [https://github.com/SAP/crossplane-provider-btp/security/code-scanning/52](https://github.com/SAP/crossplane-provider-btp/security/code-scanning/52)

The best way to fix this problem is to explicitly add a `permissions` block to the workflow or job definition to restrict the GITHUB_TOKEN to the minimal privilege necessary. Since the steps in this workflow only perform checkout, setting up Go, building, and testing—which all require only read access to repository contents—the minimal needed is `contents: read`. This can be done by adding the following block before the `jobs:` key at the root of the workflow file:

```yaml
permissions:
  contents: read
```

---

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/internal/controller/account/subaccount/fakes.go
+++ b/internal/controller/account/subaccount/fakes.go
@@ -29,6 +29,7 @@ type MockSubaccountClient struct {
 	returnSubaccount  *accountclient.SubaccountResponseObject
 	mockDeleteSubaccountExecute  func(r accountclient.ApiDeleteSubaccountRequest) (*accountclient.SubaccountResponseObject, *http.Response, error)
 	returnErr         error
+	getSubaccountErr  error
 }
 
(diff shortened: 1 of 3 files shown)


Lasserich (approved):

enrico-kaack-comp:
Yes, this is for minimal permissions.
From the security advisor:
> If a GitHub Actions job or workflow has no explicit permissions set, then the repository permissions are used. Repositories created under organizations inherit the organization permissions. The organizations or repositories created before February 2023 have the default permissions set to read-write.

From https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions
> If you specify the access for any of these permissions, all of those that are not specified are set to none.

PR #358, opened 2025-10-01
EXPERIMENT: change e2e to matrix testing for parallelisation

enrico-kaack-comp:


enrico-kaack-comp:
unfortunately, this would result in two approvals needed which is not really good. I will try another way.

sdischer-sap (approved):

PR #365, opened 2025-10-13
EXPERIMENT(Devconatiner): Adding devcontainer setup

SatabdiG:
Experimental branch for devcontainer


.devcontainer/setup.sh
```diff
@@ -0,0 +1,229 @@
...
+if ! command -v kind &> /dev/null; then
+  echo "📦 Installing kind..."
+  curl -Lo ./kind https://kind.sigs.k8s.io/dl/latest/kind-linux-amd64
+  chmod +x ./kind
+  sudo mv ./kind /usr/local/bin/kind
+else
+  echo "✓ kind already installed"
+fi
+
+# Install crossplane CLI
```
sdischer-sap: Do we have a need for that? I never used the CLI. Just curious.

.devcontainer/.env.example
```diff
@@ -0,0 +1,27 @@
+# Copy this to .env for local devcontainer development
+# NEVER commit .env to git!
+
+CIS_CREDENTIALS='{
+  "endpoints": {
+    "account_budgets_service_url": "https://account-budgets-service.cfapps.sap.hana.ondemand.com",
```
sdischer-sap: Would it be expected that the user puts in his credentials here?
If possible I would like to somehow enforce to reference the env from somewhere outside the repo. 
Just because we had some incidents already in which we accidently leaked e2e credentials because they were accidently pushed.

.devcontainer/README.md
```diff
@@ -0,0 +1,44 @@
...
+     - `BTP_GLOBAL_ACCOUNT_ID`: Your global account ID
+
+2. **Launch**: Click the "Code" button → "Create codespace on main"
+
+3. **Wait ~3 minutes** for automatic setup
+
+4. **Test it works**:
+   ```bash
+   # Install the BTP provider
+   kubectl crossplane install provider ghcr.io/sap/crossplane-provider-btp:latest
```
sdischer-sap: This would not run the dev version right? Is that what we expect here?

.devcontainer/.env.example
SatabdiG: You are absolutely correct! This file is meant to be a template only. They are meant to copy this to .env which is gitignored.

For local prevention I added the install-hook.sh which should install Pre-commit and a Pre-Push hook to prevent accidental push of credential-looking files (.env, credentials.json, pattern-matched files (*-credentials.yaml), and content patterns (clientsecret, password, etc.). It can be bypassed by  --no-verify so it's not absolute but does help. 

I can also think about some other way of credential passage, maybe vault but that might increase complexity for onboarding a bit.  What do you think?

sdischer-sap: That sounds quite good already. I was thinking if maybe there can be just a path reference in the config which suggest to not store it in the repo directly. Something like:
````
...
# credentials_file = "/Users/...env"
...
````
But if thats really filter by content already and forces you to active bypass that than it should be good enough.

.devcontainer/README.md
sdischer-sap: Thats perfect, thank you! Allows to really address the needs of those 2 different personas

.devcontainer/setup.sh
sdischer-sap: Cool. Totally makes sense. I wasn't aware of that option.

PR #374, opened 2025-10-17
Docs: external name definition and handling

enrico-kaack-comp:
This should document the meaning of external name and its implementation behavior. 

Goal is that we have a clear understanding when implementing it, unify the handling of all our MR resources and giving stakeholders a more expectable and repeatable behavior when using external-name especially in import scenarios.

### DIFF
--- a/.github/workflows/e2e_test.yaml
+++ b/.github/workflows/e2e_test.yaml
@@ -167,3 +167,23 @@ jobs:
           BTP_TECHNICAL_USER: ${{ secrets.BTP_TECHNICAL_USER }}
           CLI_SERVER_URL: ${{ secrets.CLI_SERVER_URL }}
           GLOBAL_ACCOUNT: ${{ secrets.GLOBAL_ACCOUNT }}
+
+  # required for test matrix status reporting. Ignores cleanup-e2e job failure/success.
+  e2e-test-success:
+    runs-on: ubuntu-latest
+    needs: 
+      - prepare-matrix
+      - e2e-test
+    timeout-minutes: 1
+    if: always()
+    steps:
+      - name: Fail for failed prepare-matrix job
+        if: |
+          needs.prepare-matrix.result == 'failure' || 
+          needs.prepare-matrix.result == 'cancelled'
+        run: exit 1
+      - name: Fail for failed matrix jobs
+        if: |
+          needs.e2e-test.result == 'failure' || 
+          needs.e2e-test.result == 'cancelled'
+        run: exit 1


docs/development/external-name-handling.md
```diff
@@ -0,0 +1,104 @@
...
+### Implementation guideline
+
+#### Observe()
+1. Check if external-name is empty.  
+   - If we need backwards compatibility, we create our API query from the spec.forProvider fields.  
+     - If resource exists we perform the normal needsCreation/needsUpdate and set the external name to the correct value.  
+     - If resource does not exist, it needsCreation.  
+   - Else no backwards compatibility needed, return resourceExist: false.
+2. External-name is set, we check its format.  
+   - If format is not correct (not a valid GUID or not a valid compound key): return error.
```
SatabdiG: Do we need to verify the format for the External Name? Unlike k8, an external name isn't bound to rigid DNS requirements. It can be any identifier that the provider chooses, including ARN. Checking for the sanctity of the external name is hard, in my opinion. 
For example AWS uses ARN: AWS uses ARNs: arn:aws:ec2:us-east-1:123456789012:vpc/vpc-0abc123

docs/development/external-name-handling.md
```diff
@@ -0,0 +1,104 @@
...
+This is easy in the case of a unique identifier. If we have a compound key, we deconstruct the compound key to perform the API requests. 
+
+### Special Case: defaulting of external-name
+If external-name is unset, crossplane-runtime will run a default initializer that will set the external name to metadata.name before the first Observe() is called.
+
+This poses two problems:
+- If we set any kind of initializer ourselves later, we don’t get the default initializer automatically and therefore might run into bugs if we don’t check for empty initializer
+- At this moment, the finalizer would break our definition as the metadata.name is only an identifier within the Kubernetes cluster, not external system. 
+
+**TODO: how do we handle this?**
```
SatabdiG: In my opinion,  this is crucial to answer before we consider actual implementation. How we handle this problem should be the baseline for the rest of the document. 

When external-name is unset/empty, the controller cannot perform any CRUD operations 
because it has no identifier to use in API calls to the provider. 

I think we need to answer the following before we can safely implement the pattern: 
1. What should Observe() do when external-name is empty?
   - Return an error and block reconciliation?
   - Assume resource doesn't exist and proceed to Create()?
   - Attempt to use spec.forProvider fields (but this is problematic - described in another comment)?

2. How do we prevent resource leaks?
   Scenario:
   - Create() succeeds, provider returns ID "vpc-abc123"
   - Network failure/restart occurs before the external-name annotation is written
   - Next reconcile: external-name is still empty
   - If Observe() proceeds to Create() again → duplicate resource created
   
   Reference: Crossplane Issue #2350 documents this exact problem
   https://github.com/crossplane/crossplane/issues/2350

Before finalizing this design we need to explicitly specify:
- The behavior when external-name is empty
- Safeguards against the Create() success / annotation write failure scenario
- Whether there are any cases where empty external-name is acceptable vs. should error

docs/development/external-name-handling.md
```diff
@@ -0,0 +1,104 @@
...
+
+This is not super important for us right now since we plan to have an export cli to create the spec for us but might help to have a more common behaviour accross our providers.
+
+Since the matching between MR and external resource is done based on external-name, a wrongly filled spec would result in Update() calls to the resource changing it. To prevent this, the user would not set the managementPolicy to Update. If there is then a missmatch, the Observe() method would determines this, sets the externalObservation.Diff and it would be visible in the debug log. TODO: this should be put as a warning event on the resource maybe, because crossplane-runtime only does a log.Debug with the Diff. As a user adopting existing landscapes it would be good to know if the current spec WOULD trigger an update/diff for confidence.
+
+## Open Questions, Things not considered
```
SatabdiG: General comment About Provider-BTP Implementation:
I had a few general questions: 

**1. Does BTP actually lack unique identifiers?**
Even if BTP's API requires multiple fields to query a resource, the platform 
likely assigns GUIDs internally. 

**2. Can BTP change identifier formats?**
If we hardcode compound key construction logic (e.g., "org-123|space-prod"), 
what happens when:
- BTP adds GUID fields to resources in a future API version?
- BTP changes the required identification parameters?
- BTP introduces new identifier schemes?

Hardcoded compound key logic couples our provider to the current BTP API structure.
If BTP evolves its APIs, our parsing/construction logic breaks.

Alternatively, we could consider using spec.forProvider directly
Instead of encoding fields into external-name, use spec.forProvider fields 
directly in API calls:
- More resilient to API changes
- Clearer mapping to provider API
- No parsing/delimiter collision issues
- external-name only populated if BTP returns a true unique identifier

docs/development/external-name-handling.md
```diff
@@ -0,0 +1,104 @@
+# External Name Handling
+
+## Context
+`crossplane.io/external-name` is an annotation for Managed Resources. Across providers we lack a unified definition and implementation guideline. This ADR proposes a consistent definition, expected behavior, and implementation approach.
+
+## Background
+
+### Definitions
+Common interpretations of the external name:
```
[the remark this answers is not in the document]
enrico-kaack-comp: I agree with your definition, but I could not read it out of any source like the one pager you also referenced.

docs/development/external-name-handling.md
```diff
@@ -0,0 +1,104 @@
...
+- This default conflicts with cases where the external system uses a different unique identifier.
+
+### Importing Existing External Resources (Official Flow) [3]
+1. Create an MR manifest with `spec.managementPolicies: Observe`.
+2. Add the external-name annotation with the external resource identifier. If not globally unique, also supply distinguishing `spec.forProvider` fields.
+3. Apply the MR. Controller populates `status.atProvider`.
+4. Copy the fields from `status.atProvider` to `spec.forProvider` and change `managementPolicies` to `*` to gain full control.
+
+## Our Approach
+
```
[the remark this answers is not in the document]
SatabdiG: Thank you for the great conversation! You raise a really valid UX concern about consistency.

**1. Provider-specific behavior is already normal**
Users already encounter differences across BTP resources (some are immutable, some async, different deletion behaviors). I think "some use external-name, others use spec.forProvider" could fit naturally as another resource characteristic we document.

**2. Import might actually be simpler**
- **ID-based:** "Here's the BTP ID, put it in external-name" ✅
- **No-ID:** "Fill in these spec fields that identify it" ✅

**3. Drift could be tricky**
Compound keys mean two places with the same info:
- spec.forProvider.spaceName = "staging-space"  
- external-name = "org-123|staging-space"

If they drift apart (someone edits the annotation, partial update, etc.), how do we handle it? Spec-based keeps one source of truth.

**Your points make sense:**
> "best identifier we can build"

Fair! Though I wonder if we could skip building one and just use spec.forProvider directly?

> "BTP changes require spec changes either way"

True! Though I think spec-based keeps changes localized, while compound keys also need parsing logic updates.

**Would it help to try both?** Maybe prototype CloudManagement with each approach and see which feels better in practice? Happy to help with that!

docs/development/external-name-handling.md
SatabdiG: I did look Crossplane's official guidance on this. 

1. The offical guidance is to default to metadata.name. Some providers also query by metadata.name. I think it's hard to rely on API response of 404, it can also return 400 or something else. I couldn't find any providers that do this in the wild. (Maybe I haven't looked far enough). Interestingly Upjet providers: Use DisableNameInitializer flag for provider-generated IDs
2. The official docs acknowledge leaks are a known risk when providers generate IDs. Crossplane's approach seems to be detection (via `crossplane.io/external-create-pending` annotation) + stopping reconciliation + requiring manual intervention, rather than automatic prevention.

So your position aligns with Crossplane's documented behavior. :)

docs/development/external-name-handling.md
```diff
@@ -0,0 +1,128 @@
...
+### Definition
+The external-name (string) identifies the external resource in the external system. 
+A provider can adopt two qualities:
+
+(1) The existance of the external-name express that there should be an external resource it identifies. So if the user creates a MR with the external-name set, the user explicitly states that this resource already exists in the external system and should be adopted/matched. This prevents unintended adoption/matching of existing resources. (2) The identifier than contains the information the provider needs to identify the external resource. This can be an unique identifier or a compound key.
```
enrico-kaack-comp: Dev Sync: 
- less strict about knowledge of users
- intent via mangement policies doesnt cover the case of accidently mathcing resource

docs/development/external-name-handling.md
```diff
@@ -0,0 +1,128 @@
...
+
+This is not super important for us right now since we plan to have an export cli to create the spec for us but might help to have a more common behaviour accross our providers.
+
+**TODO: how do we decide on this matter above?**
+
+Since the matching between MR and external resource is done based on external-name, a wrongly filled spec would result in Update() calls to the resource changing it. To prevent this, the user would not set the managementPolicy to Update. If there is then a missmatch, the Observe() method would determines this, sets the externalObservation.Diff and it would be visible in the debug log. 
+**TODO**: this should be put as a warning event on the resource maybe, because crossplane-runtime only does a log.Debug with the Diff. As a user adopting existing landscapes it would be good to know if the current spec WOULD trigger an update/diff for confidence.
```
enrico-kaack-comp: In Observe, we need to put this **Diff** somehwere: event and status.

Also see Circuit Breaker pattern from KymaEnv might be worth to put in a general concept for every resource (hash of spec fields stored in status to compare in Observe to determine if it needs update).
Circuit Breaker as follwo up.

PR #375, opened 2025-10-20
fix: Improve merge logic for parameterSecretRefs and parameters

jn-av:
## 📝 Description
Currently, the merge between `parameterSecretRefs` and `parameters` happens at a high level.  
Since both objects contain a `data` field directly underneath, the values of these fields are simply overwritten rather than merged.

This leads to cases where a single field (for example, `adminPassword`) overwrites the entire `data` object, which is unintended and causes loss of existing values.

---

## 💡 Proposed Change
Implement a **recursive (deep) merge** for these objects to ensure that:
- Only explicitly provided fields are updated.
- Existing data remains intact and is not overwritten unintentionally.

---

## ✅ Benefits
- Prevents accidental data loss when merging.
- Ensures expected behavior when both `parameters` and `parameterSecretRefs` define overlapping keys.
- Improves reliability and maintainability of parameter handling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,12 @@ make test-acceptance
 
 This will spin up a specific kind cluster which runs the provider as docker container in it. The e2e tests will run kubectl commands against that cluster to test the provider's functionality.
 
+If you want to run a single E2E Test locally simply set the `testFilter` variable like this:
+
+```bash
+make test-acceptance testFilter=<functionNameOfTest>
+````
+
 :warning:
 Please be aware that as part of the e2e tests a script will be executed which injects the environment configuration (see below) into the test data. Therefor you will see a lot of changes in the directory `test/e2e/testdata`after running the command. Make sure to not commit those changes into git.
 
--- a/internal/controller/kymamodule/fake/fake.go
+++ b/internal/controller/kymamodule/fake/fake.go
@@ -3,8 +3,11 @@ package fake
 import (
 	"context"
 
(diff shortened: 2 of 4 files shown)


sdischer-sap:
@jn-av looks already promising ;) Let us know when its ready for review.

internal/clients/account/serviceinstance/serviceinstance.go
```diff
@@ -160,9 +160,33 @@ func mergeJsonData(mergedData map[string]interface{}, jsonToMerge []byte) error
 	return nil
 }
 
-// mergeMaps merges a
+// addMap merges toAdd into mergedData recursively. For matching keys that are both maps,
+// it merges their contents; otherwise the value from toAdd overwrites the one in mergedData.
 func addMap(mergedData map[string]interface{}, toAdd map[string]interface{}) {
```
SatabdiG: Nitpick: Maybe, we can update the description of addMap, add an example. It would help future maintainers understand the code immediately.
```go
// addMap merges toAdd into mergedData recursively.
//
// Merge behavior:
//   - When both values are maps: Recursively merges their contents
//   - When types differ or value is not a map: toAdd's value overwrites mergedData's value
//
// Example:
//   base:  {"data": {"user": "admin", "timeout": 30}}
//   toAdd: {"data": {"timeout": 60, "password": "secret"}}
//   result: {"data": {"user": "admin", "timeout": 60, "password": "secret"}}
//                     ↑ preserved      ↑ overwritten  ↑ added
```

Issue #335, opened 2025-09-15
[FEATURE] Give `forProvider.name` to `KymaEnvironment`


Lasserich:
**Context**
Currently `KymaEnvironment` resources are seemingly using the usual Kubernetes `metadata.name` as a "default" parameter to the Kyma API itself. 

This results in the Kubernetes object name being set as the Kyma Instance name on BTP side.

Since we want to decouple Kubernetes from API logic we want to move this behavior (as per best practice) to a `forProvider.name` field on the `KymaEnvironment` resource. 

**Scope**
- [ ] Remove using `cr.name` when creating/updating `KymaEnvironment` 
- [ ] Add an optional `forProvider.name` property and use its value for API requests 
- [x] Ensure backwards compatibility by using `LateInitialize` in the `Observe()` method and setting `forProvider.name` if it is empty to `cr.name`

**We figured out that the Kyma Instance name is unchangeable**
- [x] remove the name differences in UpToDate check

## Note
Please **double check in a GitOps case** if this introduces challenges

**Additional context**
See the bug raised here to understand the behavior.
https://github.com/SAP/crossplane-provider-btp/issues/333

PR #385, opened 2025-10-24
feat: introduce `forProvider.name` in `KymaEnvironment`

Lasserich:
Relates to #335. 

Unit testing is a bit painful because it fundamentally changes a lot of hardcoded values.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -125,7 +125,13 @@ pull-docs:
 	fi
 	@git -C "$(WORK_DIR)/$(TERRAFORM_PROVIDER_SOURCE)" sparse-checkout set "$(TERRAFORM_DOCS_PATH)"
 
-generate.init: $(TERRAFORM_PROVIDER_SCHEMA) pull-docs
+# cleans the .work directory used for upjet/tf provider. Necessary after tf provider version upgrades so we just do this every time to have a clean dir
+.PHONY: clean-work
+clean-work:
+	@$(INFO) cleaning work directory
+	@rm -rf .work
+
+generate.init: clean-work $(TERRAFORM_PROVIDER_SCHEMA) pull-docs
 
 .PHONY: $(TERRAFORM_PROVIDER_SCHEMA) pull-docs terraform.buildvars
 
--- a/btp/cisclient.go
+++ b/btp/cisclient.go
@@ -487,41 +487,13 @@ func (c *Client) GetEnvironmentById(
 func (c *Client) GetCFEnvironmentByNameAndOrg(
 	ctx context.Context, instanceName string, orgName string,
 ) (*provisioningclient.BusinessEnvironmentInstanceResponseObject, error) {
-	var environmentInstance *provisioningclient.BusinessEnvironmentInstanceResponseObject
 	// additional Authorization param needs to be set != nil to avoid client blocking the call due to mandatory condition in specs
 	envInstances, err := c.getCFEnvironments(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	for _, instance := range envInstances {
-		if instance.EnvironmentType != nil && *instance.EnvironmentType != CloudFoundryEnvironmentType().Identifier {
-			continue
-		}
-		if instance.Id != nil && *instance.Id == instanceName {
-			environmentInstance = &instance
-			break
-		}
-
-		var parameters string
-		var parameterList map[string]interface{}
-		if instance.Parameters != nil {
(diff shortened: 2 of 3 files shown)


internal/controller/environment/kyma/kymaenvironment.go
```diff
@@ -80,6 +81,14 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 		return managed.ExternalObservation{}, errors.New(errNotKymaEnvironment)
 	}
 
+	// Ensures backwards compatibility and upgrades existing KymaEnvironments
+	if cr.Spec.ForProvider.Name == nil {
```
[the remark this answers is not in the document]
sdischer-sap: Or we can have a discussion about that in a bigger round any maybe decide for this approach. 
Maybe my concern isn't valid. I just haven't used this approach before and haven't seen that in any of our providers yet. So I think we need to agree on that as a valid pattern.

sdischer-sap (commented):
Nicely done. I have one concern that I put in the comments about the approach of patching spec.

internal/controller/environment/kyma/kymaenvironment.go
```diff
@@ -80,6 +81,14 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
...
+		cr.Spec.ForProvider.Name = &cr.Name
```
SatabdiG: I agree with everything Stephan mentions above. There is potentially another problem in mutating cr.Spec without checking if it exists in BTP. If an old resource exists with no forProvider.name and gets upgraded. We return immediately without checking BTP. Next reconcile will try to Observe and might create duplicate resource.

SatabdiG: Maybe we can use a runtime check, something like the following: 

```go
func getEffectiveName(cr *v1alpha1.KymaEnvironment) string {
    if cr.Spec.ForProvider.Name != nil && *cr.Spec.ForProvider.Name != "" {
        return *cr.Spec.ForProvider.Name
    }
    return cr.Name
}
```
instead of rewriting the Spec which would violate the contract. To fix the issue above I think we would need `ResourceLateInitialized` which is for autogenerated IDs not user facing fields.

internal/controller/environment/kyma/kymaenvironment.go
```diff
@@ -208,7 +221,7 @@ func (c *external) needsCreation(cr *v1alpha1.KymaEnvironment) bool {
 func (c *external) needsUpdateWithDiff(cr *v1alpha1.KymaEnvironment) (bool, string, error) {
 
 	desired, err := internal.UnmarshalRawParameters(cr.Spec.ForProvider.Parameters.Raw)
-	desired = kymaenv.AddKymaDefaultParameters(desired, cr.Name, string(cr.UID))
```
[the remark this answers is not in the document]
Lasserich: Ah yes sure thing, thanks!

internal/controller/environment/kyma/kymaenvironment.go
Lasserich: Hm interesting thoughts! I do not really understand how this would result in a duplicate creation. Since this is meant only for already existing instances. Basically the flow is we copy over the name during `LateInitialize`, if this metadata.name is faulty or already exists then in the next Observe we use the copied name and get an error from the API which is rightfully there no? 
I feel like by having the getEffectiveName we would drag on old quirks - metadata.name should never be used for neither creation nor observes.

SatabdiG: The duplication happens because the early ResourceLateInitialized: true return masks the "Not Found" status of the potential external resource. By skipping the BTP API call, we don't confirm that the external resource truly exists. When the next Observe runs, if the BTP API says "Not Found," Crossplane will treat the in-cluster CR as a new resource to be created externally. 

The code is meant for existing instances, but the current code doesn't enforce that intent. It only checks if the field is missing, which is a state that might be true for both: a) A truly existing, but old, BTP resource. b) An in-cluster CR that was never successfully created in BTP.

By returning early, case (b) gets incorrectly treated like case (a), leading to an eventual creation attempt.

I agree about metadata.name  but that piece of code is meant only as temporary lookup and not persisting something in cr.Spec.ForProvider.Name without baking it into the resource's state

Lasserich:
E2E successful run here:
https://github.com/SAP/crossplane-provider-btp/actions/runs/19095751428

Issue #352, opened 2025-09-24
[FEATURE] Provide BTP Instance sharing support for `ServiceInstance` creation


snellissery:
**Context**
Today Crossplane provided BTP ServiceInstance CRD from BTP provider does not support BTP instancesharing.
This is an existing BTP supported feature and we have use-cases to utilize this feature.
Reference existing  CRD: https://pages.github.tools.sap/cloud-orchestration/browser/Providers/provider-btp/account.btp.sap.crossplane.io/serviceinstance/v1alpha1

**Scope**
Please help us to onboard BTP instance sharing feature for ServiceInstance CRD as provided by BTP in https://help.sap.com/docs/btp/btp-cli-command-reference/btp-share-services-instance 

**Technical Steps**
There is a shared fields available in the underlying subaccount_service_instance in the terraform provider. Should be easy to add [this field](https://registry.terraform.io/providers/SAP/btp/latest/docs/resources/subaccount_service_instance#shared-1) to our CRD and just transfer it over.

### Remarks
- Make sure its actually set (from the status, CLI or maybe Cockpit)
- Check if there are additional attributes we might have missed in initial implemention, if its easy, add them 

### Out of scope
- adding detailed e2e testing for that feature

## Acceptance Criteria
- support [this field](https://registry.terraform.io/providers/SAP/btp/latest/docs/resources/subaccount_service_instance#shared-1)  in `ServiceInstance`

Lasserich:
Hi @snellissery thanks for raising this! How are you feeling about this issue, do you feel comfortable contributing this feature to the provider? :) 

Lets hop on a call together and discuss next steps on this issue!

PR #406, opened 2025-11-18
feat: support shared parameter for ServiceInstance

Lasserich:
small change in the terraform client which did not get this field parsed before.

Refer to [this issue](https://github.com/SAP/crossplane-provider-btp/issues/352).


Lasserich:
Ahh okay that makes sense, then I will take it out of the e2e options!

enrico-kaack-comp (approved):
Fully missed my noticiations sorry. LGTM

Issue #398, opened 2025-11-13
[TASK] External-Name(Subaccount): Ensure ADR compliance


enrico-kaack-comp:
## Description

This issue tracks the implementation work needed to ensure the Subaccount resource fully complies with the [External Name Handling ADR](https://github.com/SAP/crossplane-provider-btp/pull/374). The goal is to verify and implement proper external name handling across all CRUD operations according to the established guidelines.

## Tasks

> Please note, for all tasks the [External Name Handling ADR](https://github.com/SAP/crossplane-provider-btp/pull/374) acts as a source of truth, the details here are for just for reference. If deviating from the ADR, please note why.

### 1. Implement Proper External Name Handling

**Acceptance Criteria:**
- [ ] **Observe() Method Implementation**

- [ ] **Create() Method Implementation**

- [ ] **Update() Method Implementation**

- [ ] **Delete() Method Implementation**

- [ ] **Review Migration Logic**:
  - Evaluate existing migration logic (introduced in v1.4.0) against ADR requirements
  - Update or remove as needed to ensure ADR compliance
  - Update documentation for any breaking changes

### 2. Unit Test Coverage for External Name Handling

**Acceptance Criteria:**
- [ ] **Observe() Tests**:
  - Test empty external-name scenarios (witch backwards compatibility)
  - Test invalid GUID format in external-name
  - Test valid GUID that doesn't exist (404 response) → should return `resourceExists: false`
  - Test valid GUID that exists → normal observation flow
  - Test drift detection and diff reporting

- [ ] **Create() Tests**:
  - Test successful creation sets external-name to returned GUID
  - Test "resource already exists" error handling (external-name should not be set)
  - Test other creation errors (external-name should not be set)

- [ ] **Update() Tests**:
  - Test update with valid existing external-name
  - Test validation of immutable fields (if any)

- [ ] **Delete() Tests**:
  - Test deletion with valid external-name
  - Test 404 response handling (should not be treated as error)
  - Test async deletion scenarios with deletion states
  - Test deletion of externally removed resource

### 3. E2E Import Test Coverage

**Acceptance Criteria:**
- [ ] **Import Flow Test**:
  - Test importing existing Subaccount resources using the external-name annotation
  - Verify imported resources transition to healthy state with correct external-name
  - Test that imported resources can be observed and updated properly
[...]

PR #412, opened 2025-11-21
chore: subaccount follow external-name definition

enrico-kaack-comp:
#398 

- subaccount controller follows ADR logic
- subaccount_test unit tests test the behaviour
- overwrite default initializer
- import test in e2e

ALSO: import test is a reuseable framework. Can be used for other resources later on.

## Review Suggestion
1. review the reuseable ImportTest helper `test/e2e/import_utils_test.go` and its usage for subaccount `test/e2e/subaccount_test.go`
2. changes in the controller logic of subaccount in `internal/controller/account/subaccount/subaccount.go` following the docs in https://github.com/SAP/crossplane-provider-btp/blob/main/docs/development/external-name-handling.md
3. the changes in the unit tests also according to the doc

I dont like the implemntation in `isResourceAlreadyExistsError`, maybe we have a better idea here?

## Follow Up in additional ticket
- upgrade test

### DIFF
--- a/btp/cisclient.go
+++ b/btp/cisclient.go
@@ -565,13 +565,44 @@ func (c *Client) ExtractOrg(cfEnvironment *provisioningclient.BusinessEnvironmen
 		label = *cfEnvironment.Labels
 	}
 
-	return c.NewCloudFoundryOrgByLabel(label)
+	return NewCloudFoundryOrgByLabel(label)
 }
 
-func (c *Client) NewCloudFoundryOrgByLabel(label string) (*CloudFoundryOrg, error) {
-	var cloudFoundryOrg *CloudFoundryOrg
-	err := json.Unmarshal([]byte(label), &cloudFoundryOrg)
-	return cloudFoundryOrg, err
+// NewCloudFoundryOrgByLabel creates a CloudFoundryOrg from a JSON-formatted labels string.
+// In legacy format, the keys have a trailing colon (:), while in the new format they do not.
+// The function handles both formats.
+func NewCloudFoundryOrgByLabel(rawLabels string) (*CloudFoundryOrg, error) {
+	if rawLabels == "" {
+		return nil, errors.New("labels string is empty")
+	}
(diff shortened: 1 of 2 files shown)


internal/controller/account/subaccount/subaccount.go
```diff
@@ -481,6 +507,23 @@ func specifyAPIError(err error) error {
...
 
+// isResourceAlreadyExistsError checks if the API error code is 409 for subaccount already exitsts in the region
+func isResourceAlreadyExistsError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	if genericErr, ok := err.(*accountclient.GenericOpenAPIError); ok {
+		if accountError, ok := genericErr.Model().(accountclient.ApiExceptionResponseObject); ok {
+			return internal.Val(accountError.Error.Code) == 409
```
SatabdiG: Maybe it can panic if we don't add explicit nil checks before dereferencing `accountError.Error.Code`. 
Could we use HTTP response and check `httpResp.StatusCode == 409` as fallback?

internal/controller/account/subaccount/subaccount.go
```diff
@@ -481,6 +507,23 @@ func specifyAPIError(err error) error {
...
+		}
+		if genericErr.Body() != nil {
+			return false
```
SatabdiG: If the genericErr.Body() != nil why are we returning false

internal/controller/account/subaccount/subaccount.go
```diff
@@ -481,6 +507,23 @@ func specifyAPIError(err error) error {
 	return err
 }
 
+// isResourceAlreadyExistsError checks if the API error code is 409 for subaccount already exitsts in the region
+func isResourceAlreadyExistsError(err error) bool {
+	if err == nil {
+		return false
```
SatabdiG: Maybe this function would benefit from more explicit logging? So we have more context on what fails?

PR #414, opened 2025-11-24
Feat(OIDCCertLogin): Allow intermediate certificates

### LINKED ISSUE #428: OIDCCertBasedLogin allow intermediate certificates
A user is using the OIDCCertBasedLogin. It does not support intermediate certificates right now. He provided us with a fix that I pushed into a branch and verified with him that it is working #414 

What is missing is to understand the the logic, test it manually too and polish the logic to make it merge ready.

## Definition of Done
- [ ] Adapt unit tests to cover this behaviour
- [ ] test it manually
- [ ] ensure the feature is not breaking previous behaviour and is still working with root certificate only

enrico-kaack-comp:
This PR adds support for the OIDCCertLogin for the cert to contain intermediate certificates (before, the intermediate certificates were missing).

To not break the previous logic (which ignored intermediate certificates), this introduces a new `spec.ForProvider.containsCertificateChain` bool flag (default false --> old behaviour) to explicitly enable this new behaviour.

Furthermore, this adds unit tests to cover this new parsing of the chain, reusing the openssl teest that only has one certificate by adding an intermediate and root ca.

Closes #428

### DIFF
--- a/internal/clients/account/serviceinstance/serviceinstance.go
+++ b/internal/clients/account/serviceinstance/serviceinstance.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"reflect"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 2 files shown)


sdischer-sap (approved):
This looks very good to me, thanks for the in depth documentation as well.

PR #426, opened 2025-12-03
fix: CloudFoundryEnvironment legacy label format parsing

enrico-kaack-comp:
A Cloud Foundry environment created long in the past have a different format of the Labels field:
```
// old format
labels={"Org ID:":"id", "Org Name":"test-name", "API Endpoint:":"api-endpoint"}
// new format
labels={"Org ID":"id", "Org Name":"test-name", "API Endpoint":"api-endpoint"}
```
The difference is the `:` in the end of the json key. This fix will handle both formats.

**NOTE** the missing `:` for `Org Name` in the old format, as this old format didnt include this.

### DIFF
--- a/btp/cisclient.go
+++ b/btp/cisclient.go
@@ -586,14 +586,13 @@ func NewCloudFoundryOrgByLabel(rawLabels string) (*CloudFoundryOrg, error) {
 	}
 
 	oldOrgId, oldOrgIdExists := labels["Org ID:"]
-	oldOrgName, oldOrgNameExist := labels["Org Name:"]
 	oldApiEndpoint, oldApiEndpointExist := labels["API Endpoint:"]
 
-	if oldOrgIdExists || oldOrgNameExist || oldApiEndpointExist {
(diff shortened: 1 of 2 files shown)


btp/cisclient.go
```diff
@@ -565,13 +565,45 @@ func (c *Client) ExtractOrg(cfEnvironment *provisioningclient.BusinessEnvironmen
...
+			}
+		}
+	}
+
+	//map the old format
+	return &CloudFoundryOrg{
+		Id:          labels["Org ID:"],
+		Name:        labels["Org Name:"],
+		ApiEndpoint: labels["API Endpoint:"],
+	}, nil
```
SatabdiG: Do we also need to check if the keys exist for the old format and return an error? 
Right now if old format keys are missing, we'd return a CloudFoundryOrg with empty strings.

Something like:
```go
!hasOrgName || !hasAPIEndpoint {
    return nil, errors.New("labels missing required fields in both new and old formats")
}
```

PR #435, opened 2025-12-08
docs: added docks for the extrnal name support

mpechkurov:
This PR includes external name documentation for user. 
Also includes script which will generate the docs for the external name based on the template.

### DIFF
--- a/scripts/generate-external-name-docs.go
+++ b/scripts/generate-external-name-docs.go
@@ -1,17 +1,3 @@
-// Copyright 2025 The Crossplane Authors. All rights reserved.
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
-//
-//     http://www.apache.org/licenses/LICENSE-2.0
-//
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
-
 // This tool scans all *_types.go files in the apis directory, extracts External-Name
 // Configuration comments from resource type definitions, and generates comprehensive
 // documentation in docs/user/external-name.md.
@@ -36,6 +22,7 @@ package main
 
 import (
 	"bufio"
+	"flag"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


scripts/generate-external-name-docs.go
```diff
@@ -0,0 +1,291 @@
+// Copyright 2024 The Crossplane Authors. All rights reserved.
+//
```
enrico-kaack-comp: Copyright is in this case SAP SE 2025 :)

docs/development/external-name-handling.md
```diff
@@ -87,6 +87,41 @@ If our delete request is successful, we return EVEN IF the deletion operation is
...
+// External-Name Configuration:
+//   - Follow Standard: yes|no [explanation if no]
+//   - Format: <identifier description>
+//   - How to find:
+//     - UI: <navigation path>
+//     - CLI: <command> (field: <field_name>)
+//
+```
+
+## Key Elements
```
schlingemann: It could make sense to have an example here for the three elements, e.g. the subaccount elements.

That way we ensure the explanations (like "How to find") look similar for all CRDs.

apis/account/v1alpha1/directory_types.go
```diff
@@ -101,6 +101,14 @@ type DirectoryStatus struct {
 // +kubebuilder:object:root=true
 
 // A Directory is a managed resource that allows grouping of subaccounts in the SAP Business Technology Platform
+//
+// External-Name Configuration:
+//   - Follow Standard: yes
```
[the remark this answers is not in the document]
mpechkurov: We can change it.

scripts/generate-external-name-docs.go
```diff
@@ -0,0 +1,293 @@
+// Copyright 2025 The Crossplane Authors. All rights reserved.
```
sdischer-sap: Is this script copied from crossplane? 
We generally need to remove this copyright headers unless its indeed something which has been fully developed by crossplane. There is an OSPO github action runner to check for proper licensing headers.

Makefile
```diff
@@ -309,3 +309,10 @@ publish:
 		docker push $(DOCKER_REGISTRY)/$${image}:$(VERSION); \
 	done
 	@$(OK) "Publishing images $(PUBLISH_IMAGES) to $(DOCKER_REGISTRY)"
+
+# Generate external-name documentation from *_types.go files
+.PHONY: docs.generate-external-name
+docs.generate-external-name:
```
enrico-kaack-comp: can we get this integrated into `make reviewable` and `make check-diff` https://github.com/SAP/crossplane-provider-btp/blob/main/.github/workflows/reviewable_check_diff.yaml#L32 ?

mpechkurov: I can work on this in separate PR.

PR #458, opened 2025-12-19
fix: handle different formats for connection details

CsDenes:
This PR fixes the parsing of legacy labels formats for CloudFoundryEnvironment when creating the connection details.

The parsing of legacy labels when creating a new CloudFoundryOrg was fixed in https://github.com/SAP/crossplane-provider-btp/pull/426.


internal/clients/cfenvironment/client.go
enrico-kaack-comp: add comment that the connection details support handling the old format

internal/clients/cfenvironment/client.go
```diff
@@ -90,33 +90,35 @@ func GetConnectionDetails(instance *provisioningclient.BusinessEnvironmentInstan
...
 
-	if cflabels.OrgName != nil {
-		details[v1alpha1.ResourceOrgName] = []byte(*cflabels.OrgName)
+	if orgName, ok := cflabels["Org Name"]; ok {
+		details[v1alpha1.ResourceOrgName] = []byte(orgName)
 	}
-	if cflabels.OrgId != nil {
-		details[v1alpha1.ResourceOrgId] = []byte(*cflabels.OrgId)
+
+	for _, orgIdKey := range []string{"Org ID:", "Org ID"} {
```
enrico-kaack-comp: nit: this way the old format take precedence over the new format. I suggest we do it the other way round just to be safe. If both combinations are available, use the newer one (without `:`)

internal/clients/cfenvironment/client.go
Noel-Ch:
```suggestion
// GetConnectionDetails creates ConnectionDetails from a JSON-formatted labels string.
// In legacy format, the keys have a trailing colon (:), while in the new format they do not.
// The function handles both formats.
func GetConnectionDetails(instance *provisioningclient.BusinessEnvironmentInstanceResponseObject) (managed.ConnectionDetails, error) {
```

enrico-kaack-comp (approved):

Issue #427, opened 2025-12-03
[EPIC] External-Name Compliance


enrico-kaack-comp:
# Context

In the second round of external-name discussion, we saw the neccessity to formaly define our external-name handling in the [ADR](https://github.com/SAP/crossplane-provider-btp/blob/main/docs/development/external-name-handling.md).

Now we have to apply this decision to all our managed resources.

They can be divided in different groups depending on the implementaiton complexity:
1. Already having external-name with the right format, only needing some missing edge-case unit tests and import test (and upgrade-test once it is ready)
2. Having external-name but requiring external-name format to be defined and maybe changed.
3. Dont have external-name support at all and it needs to be implemented.

**Note on import tests**: for the import tests there is a reuseable struct that should make it easy to implement.

The following is a list with the resources categorized:

## Only ensure compliance, add edge-case unit tests, import tests and upgrade test

| Name | Type | Format | Compliance | Import Test | Upgrade Test |
|------|------|--------|------------|-------------|--------------|
| Subaccount | Native | UUID (subaccount GUID) - [Source](https://github.com/SAP/crossplane-provider-btp/tree/main/internal/controller/account/subaccount/subaccount.go#L392-L429) | ✅  #398  | ✅  #398 | ✅ #574 |
| Directory | Native | UUID (directory GUID) - [Source](https://github.com/SAP/crossplane-provider-btp/tree/main/internal/clients/directory/directory.go#L128) | ✅ #439  | ✅ #439 | ✅ #575 |
| ServiceBinding | Hybrid | Service binding ID - [Source](https://github.com/SAP/crossplane-provider-btp/tree/main/internal/controller/account/servicebinding/servicebinding.go#L209) | #440  | #440  | #440  |
| ServiceInstance | Hybrid | Service instance ID - [Source](https://github.com/SAP/crossplane-provider-btp/tree/main/internal/controller/account/serviceinstance/serviceinstance.go#L164) | ✅ #497  |  ✅ #497 |  ✅ #497 |
| KymaEnvironment | Native | UUID (environment instance ID) - [Source](https://github.com/SAP/crossplane-provider-btp/tree/main/internal/controller/environment/kyma/kymaenvironment.go#L156) | ✅   #530  | ✅   #530  | ✅   #530 |
[...]

PR #461, opened 2025-12-22
feat: `KymaModule` external-name compliance

### LINKED ISSUE #320: (ExternalName) `KymaModule` Add full external name support
## Problem
`KymaModule` MR has no support for external name now. 
We need to set it during Create and use it for matching during observe.

First evaluate if KymaModule can have external name support since it maps the MR to one cr in the Kyma Cluster. It might not need external-name support for the export/import functionality.

NOTE: follow the external-name ADR here https://github.com/SAP/crossplane-provider-btp/blob/main/docs/development/external-name-handling.md

## Context
During Create external-name is set to `forProvider.name`. Kyma CR contains a list of modules in this cluster, one `KymaModule` is only mapped into this list. There is no unique key besides the module name. Every module can only exist once with one channel specified. Choosing a `forProvider.name` is a parameter and unique key. 

## Acceptance Criteria
- [x] Create sets the external-name to the external identifier (either an id or a composite key that is used for later matching)
- [x] Observe uses the external-name annotation as primary matching key so the initial observe would match an existing resource even if it was not created in a previous reconcile of the provider
- [x] Unit test tests the external-name matching behaviour

Lasserich:
This PR ensures external-name compliance for the `KymaModule` resource. 

- fixes https://github.com/SAP/crossplane-provider-btp/issues/320

### DIFF
--- a/internal/clients/kymamodule/module.go
+++ b/internal/clients/kymamodule/module.go
@@ -6,6 +6,7 @@ import (
 
 	client "sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/crossplane/crossplane-runtime/pkg/meta"
 	"github.com/pkg/errors"
 	"github.com/sap/crossplane-provider-btp/apis/environment/v1alpha1"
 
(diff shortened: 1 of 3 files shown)


enrico-kaack-comp (changes requested):
Thank you for starting this.
Over the last month we worked on the ADR for external name compliance so we have a few more points to check to make kyma module ADR compliant.
See EPIC #427  and example issues for e.g. subaccount in #398 that defines more cases of how the external name should be handled. There is no Task for Kyma Module specific yet (the #320 you mentioned I commented with closed but failed to actually put it in close, sorry for that).

enrico-kaack-comp:
handled in #602 with newer requirements with regard to upgrade/import tests

Issue #416, opened 2025-11-27
[FEATURE] Add `origin`/`idp` parameter support for Terraform provider configuration to enable custom IdP authentication


Sim-Jar:
### Problem / Current Limitation

When creating Terraform-managed resources (e.g., `SubaccountTrustConfiguration`, `GlobalAccountTrustConfiguration`, `SubaccountApiCredential`, etc.) with custom Identity Providers (IdPs) or Corporate IdPs, authentication fails because the `origin` parameter is not passed to the BTP Terraform provider.

**Current behavior:**
- The provider only passes `username`, `password`, `globalaccount`, and `cli_server_url` to the Terraform provider
- Authentication always uses the default SAP ID Service (`sap.ids`)
- Users with credentials in custom IdPs cannot authenticate

**Error observed:**
```
Login failed
Error: cannot get Service Account credentials
```

### Use Case

Organizations using **custom Identity Authentication Service (IAS) tenants** or **Corporate SAML/OIDC IdPs** need to specify the `origin` parameter during BTP CLI authentication. The BTP CLI and Terraform provider both support this via:

**BTP CLI:**
```bash
btp login --user <username> --password <password> --origin <custom.idp.com>
```

**Terraform Provider:**
```hcl
provider "btp" {
  username    = "user@company.com"
  password    = "password"
  idp         = "custom.accounts.ondemand.com"
  globalaccount = "subdomain"
}
```

However, the Crossplane provider does not currently pass this parameter to the Terraform provider.

### Affected Resources

All Terraform-managed resources are affected:
- `SubaccountTrustConfiguration`
- `GlobalAccountTrustConfiguration`
- `SubaccountApiCredential`
- `SubaccountServiceBroker`
- `DirectoryEntitlement`

### Proposed Solution

Extract the `origin` field from CloudFoundry/Service Account credentials and pass it to the Terraform provider as the `idp` configuration parameter.

#### Implementation Changes Needed

**File:** `internal/clients/tfclient/setup.go`

**Function:** `TerraformSetupBuilder` (lines ~98-103)

**Current code:**
```go
ps.Configuration = map[string]any{
    "username":       userCredential.Username,
    "password":       userCredential.Password,
    "globalaccount":  pc.Spec.GlobalAccount,
    "cli_server_url": pc.Spec.CliServerUrl,
}
```

**Proposed change:**
```go
// Set default origin if not provided
origin := userCredential.Origin
if origin == "" {
    origin = "sap.ids"
}

[...]

PR #466, opened 2026-01-08
feat: add originidp parameter support for terraform provider configuration

jn-av:
**What this PR does:**

This pull request adds the ability to configure a custom Identity Provider (IdP) by specifying an origin in the provider's credential secret. This enables users with custom IdPs to successfully authenticate against the BTP.

This enhancement unblocks the management of several resources for these users, such as `SubaccountTrustConfiguration`, `GlobalAccountTrustConfiguration`, and `SubaccountApiCredential`.

Addresses the feature request in issue #416.

**How it works:**

The origin for the IdP can now be configured within the provider's Secret. The Terraform provider will use this origin during the authentication process. If the origin field is omitted from the secret, the provider will fall back to its default behavior, using the standard SAP Identity Provider. This approach ensures backward compatibility, as existing configurations will continue to function without any changes.

**Examples of Usage:**

Secret with a custom origin:
```yaml
apiVersion: v1
kind: Secret
metadata:
  namespace: default
  name: sa-provider-secret
type: Opaque
stringData:
  credentials: |
    {
      "email": "<mail>",
      "username": "<user-name>",
      "password": "<password>",
      "idp": "my-custom-idp"
    }
```

Secret without a custom origin (default behavior):
```yaml
apiVersion: v1
kind: Secret
metadata:
  namespace: default
  name: sa-provider-secret
type: Opaque
stringData:
  credentials: |
    {
      "email": "<mail>",
      "username": "<user-name>",
      "password": "<password>"
    }
```


internal/clients/tfclient/setup_test.go
```diff
@@ -0,0 +1,263 @@
...
+	testSecretNS      = "test-namespace"
+	testProviderName  = "test-provider"
+	testUsername      = "test-user@example.com"
+	testPassword      = "test-password"
+	testGlobalAccount = "test-global-account"
+	testCliServerURL  = "https://cli.example.com"
+	testCustomIDP     = "custom-idp.example.com"
+)
+
+func TestTerraformSetupBuilder_ConditionalIDP(t *testing.T) {
```
sdischer-sap: Nice, just wanted to say I like that approach ;) 
With a new feature add a test for that and only care about this newly added behaviour. Adding missing tests for existing features can be added in the future. 👍

PR #467, opened 2026-01-11
Add support for slices when merging maps

jpraychev:
Whenever maps from inline and secret parameters are merged, slices operations are ignored and the final json parameters are not complete. For example, the following snippet tries to merge the two maps but the results ignores the slice from secret parameters.

```
	inlineParams := map[string]any{
		"data": map[string]any{
			"user":    "admin",
			"timeout": 30,
			"array":   []any{1, 2, 3},
		},
	}
	secretParams := map[string]any{
		"data": map[string]any{
			"timeout":  60,
			"password": "secret",
			"array":    []any{4, 5},
		},
	}

result: map[data:map[array:[1 2 3] password:secret timeout:30 user:admin]]
```

The feature adds support for slices by appending elements from one slice to the other if slices are under the same `key`, otherwise it preserves the previous merging logic.


jn-av (approved):
LGTM! Nice find, I didn't think of that case when I first wrote this. Also, thanks for updating `interface{}` to `any`. It looks way cleaner and more modern now ✅

PR #513, opened 2026-02-18
fix: pass IDP origin to CF API login

### LINKED ISSUE #506: [BUG] CF API login does not pass Identity Provider origin
# CloudFoundryEnvironment: CF API login does not pass Identity Provider origin

v1.6.1

## Problem

The `CloudFoundryEnvironment` controller authenticates against the Cloud Foundry UAA using only username and password, without passing the Identity Provider origin. This causes authentication to fail with `invalid_grant` when the Service Account uses a custom Identity Provider.

The CF client is created in `internal/clients/cfenvironment/cfenvironment.go`:

```go
cfv3config, err := config.New(url, config.UserPassword(username, password))
```

The `Idp` field from the Service Account secret is available in `btp.Credential.UserCredential.Idp` but is never passed to the CF client configuration. Without the origin login hint, the CF UAA cannot resolve the user against the correct Identity Provider.

This affects both code paths that create the CF client:
- `createClient()` (line 108) — used during **Observe** (`DescribeInstance` → `getManagers`)
- `createClientWithType()` (line 119) — used during **Create** (`CreateInstance`)

**To Reproduce**
1. Configure a ProviderConfig with a Service Account that uses a custom Identity Provider (with `idp` field set in the SA secret)
2. Create a `CloudFoundryEnvironment` CR
3. The controller attempts to log into the CF API without the IdP origin
4. CF UAA returns `invalid_grant` ("User authentication failed")
5. Both Observe and Create fail — the resource never becomes `READY`

**Expected behavior**
The controller should pass the Identity Provider origin from the Service Account secret to the CF client configuration, so that the CF UAA can resolve the user against the correct IdP.

## Impact

This is a blocking issue: `CloudFoundryEnvironment` resources can never reach `READY` state when a custom Identity Provider is used. Unlike the related issue with hardcoded `sap.ids` for Org Managers (#505 ), this affects the controller's ability to interact with the CF API at all — no observation, creation, or deletion is possible.

**Additional context**
The Service Account secret already supports an `idp` field, and the Terraform setup builder (`internal/clients/tfclient/setup.go:106-108`) correctly passes it through for TF-based resources. The CF client setup does not follow the same pattern.

Sim-Jar:
## Summary

- `newOrganizationClient()` now accepts an `origin` parameter and conditionally passes it to `config.Origin()` in the go-cfclient configuration
- `createClient()` and `createClientWithType()` forward `UserCredential.Idp` from provider credentials
- Fixes CF API login returning `invalid_grant` when a custom identity provider is configured

Closes #506

## Changes

- **`internal/clients/cfenvironment/cfenvironment.go`**: Add `origin string` parameter to `newOrganizationClient()`, use `config.Option` slice with conditional `config.Origin()` append. Pass `c.btp.Credential.UserCredential.Idp` in both `createClient()` and `createClientWithType()`.
- **`internal/clients/cfenvironment/cfenvironment_test.go`**: Add `TestNewOrganizationClient_Validation` table-driven test covering missing org name, missing orgGuid, empty origin, and non-empty origin cases.

## Test plan

- [ ] Existing unit tests pass (`go test ./internal/clients/cfenvironment/...`)
- [ ] New `TestNewOrganizationClient_Validation` test passes
- [ ] Manual verification: CF login with custom IDP no longer returns `invalid_grant`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/clients/cfenvironment/cfenvironment.go
+++ b/internal/clients/cfenvironment/cfenvironment.go
@@ -107,7 +107,7 @@ func (c CloudFoundryOrganization) createClient(environment *provisioningclient.B
 
 	cloudFoundryClient, err := newOrganizationClient(
 		org.Name, org.ApiEndpoint, org.Id, c.btp.Credential.UserCredential.Username,
-		c.btp.Credential.UserCredential.Password,
+		c.btp.Credential.UserCredential.Password, c.btp.Credential.UserCredential.Idp,
 	)
 	return cloudFoundryClient, err
 }
@@ -118,7 +118,7 @@ func (c CloudFoundryOrganization) createClientWithType(org *btp.CloudFoundryOrg)
 ) {
 	cloudFoundryClient, err := newOrganizationClient(
 		org.Name, org.ApiEndpoint, org.Id, c.btp.Credential.UserCredential.Username,
-		c.btp.Credential.UserCredential.Password,
+		c.btp.Credential.UserCredential.Password, c.btp.Credential.UserCredential.Idp,
 	)
 	return cloudFoundryClient, err
 }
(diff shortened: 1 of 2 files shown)


sdischer-sap (approved):
Nice, this looks good to me. Unfortunately there appear to be conflicts. Could you resolve them? Good to be merged from my side

n-losse:
Cannot fix the merge conflicts in this PR since the fork repo is archived. Reopening as https://github.com/SAP/crossplane-provider-btp/pull/831

PR #527, opened 2026-02-25
fix: add caching to Entitlement

### LINKED ISSUE #256: [Spike] Optimization of Entitlement: many network requests
**Describe the bug**
When applying entitlements, customers report **`429 Too Many Requests`** api responses starting at **20-50 entitlements**.
The entitlements are synced and ready, but this is a general scaleability issue in our provider.
I think entitlements can easily grow in numbers for bigger user projects so we should make sure to handle them efficiently.

v1.2.1

## Cheap workaround
- Adjust Poll interval - use different `Default`

**To Reproduce**
1. Apply Entitlement CR (around 20)
2. Run controller locally with debug flag
3. See logs for entitlements (e.g. `HTTP Request "method": "GET", "path": "/entitlements/v1/assignments"`), that many entitlements should then probably lead to 429 Too Many Requests.

**Expected behavior**
Our provider **should be able to handle many entitlements at once** without running into rate limits since customers might have a huge number of entitlements.

**Additional context**
For every entitlement, for every sync (default 1min) for every service name we make one call to the btl cli to get all entitlements in the subaccount:
```go
	response, _, err := c.btp.EntitlementsServiceClient.
		GetDirectoryAssignments(ctx).
		SubaccountGUID(cr.Spec.ForProvider.SubaccountGuid).
		AssignedServiceName(cr.Spec.ForProvider.ServiceName).
		Execute()
	if err != nil {
		return nil, err
	}
```
https://github.com/SAP/crossplane-provider-btp/blob/main/internal/clients/entitlement/entitlement.go#L36

If we have multiple entitlements with the same service names we would make this call repeatedly.
This call can be changed to:
- return all entitlements for all subaccount and directories in the global account
- return all entitlements for all service names
We could also make it without the specification of the service name and would get the entitlements for all service names.
Then we would only need to do one network call for all entitlements and reuse this for this sync round for all entitlment crs we have. Same would go for the subaccount specification here.
The network request might get larger but we reduce the overall number of requests (maybe it is possible to specify multiple service names/subaccounts in one request even, then we could have the same network request size as we have now but without multiple requests.

**Steps**
- [ ] reproduce the situation that 20 entitlement crs produce 429. By this we can analyse what entitlement pattern lead to this (maybe all in one subaccount or multiple, maybe all one service plan or multiple, maybe there is no pattern)
- [ ] based on this decide on how to improve this by changing the btp cli call and reuse the answer for others, maybe a cache

## Acceptance critera
- [ ] put followup work into new issues
- [ ] bring to dev sync - add to community guide

dattito:
Fixes https://github.com/SAP/crossplane-provider-btp/issues/256.

Because a lot of Entitlement CRs in a cluster lead to a burst of API requests against the BTP server, this PR introduces a chaching layer that prevents to many `GET` calls to the BTP server. The cache invalidation timeout is 10 seconds.

Some modifications, like Create, Update or Delete, will automatically invalidate the cache, because we can be sure that the external state has changed.

### DIFF
--- a/internal/clients/entitlement/client.go
+++ b/internal/clients/entitlement/client.go
@@ -19,6 +19,8 @@ const (
 
 type Client interface {
 	DescribeInstance(ctx context.Context, cr *apisv1alpha1.Entitlement) (*Instance, error)
+	DescribeInstanceCached(ctx context.Context, cr *apisv1alpha1.Entitlement) (*Instance, error)
+	ClearDescribeInstanceCache(cr *apisv1alpha1.Entitlement)
 	CreateInstance(ctx context.Context, cr *apisv1alpha1.Entitlement) error
 	DeleteInstance(ctx context.Context, cr *apisv1alpha1.Entitlement) error
 	UpdateInstance(ctx context.Context, cr *apisv1alpha1.Entitlement) error
--- a/internal/clients/entitlement/client.go
+++ b/internal/clients/entitlement/client.go
@@ -19,8 +19,6 @@ const (
 
 type Client interface {
 	DescribeInstance(ctx context.Context, cr *apisv1alpha1.Entitlement) (*Instance, error)
-	DescribeInstanceCached(ctx context.Context, cr *apisv1alpha1.Entitlement) (*Instance, error)
-	ClearDescribeInstanceCache(cr *apisv1alpha1.Entitlement)
 	CreateInstance(ctx context.Context, cr *apisv1alpha1.Entitlement) error
 	DeleteInstance(ctx context.Context, cr *apisv1alpha1.Entitlement) error
 	UpdateInstance(ctx context.Context, cr *apisv1alpha1.Entitlement) error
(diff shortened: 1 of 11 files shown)


enrico-kaack-comp (changes requested):

internal/clients/entitlement/client.go
```diff
@@ -19,6 +19,8 @@ const (
 
 type Client interface {
```
maximilianbraun: I would suggest to hide the cache and just have another client which is cached and make this as transparent as you can. 

The CachingClient constructor should just take a client which does the actual calls and decide when to retrieve from cache or when to call the service. Basically [decorator pattern](https://refactoring.guru/design-patterns/decorator).

Not a blocker, rather a design choice. I would suggest that for final implementation.

maximilianbraun (commented):
additionally to my comment: 

The final implementation (we should validate how well the approach works in real life, beforehand) should expose metrics (https://book.kubebuilder.io/reference/metrics.html), ie. how the cache hit / miss rate is. This would anyway be interesting for evaluation in real life and to fine tune parameters. The Library is exposing different metrics already. Maybe logging them every 30s might already help for validation. 

I wonder on the other side how its about syncronisation. The cache might only be effective if consecuritve runs wait for each other, the entitlement get call is quite expensive (terms of size) and might take some time to transport. 

I think the loader can help with that. https://maypok86.github.io/otter/user-guide/v2/examples/#loading

sdischer-sap:
Closing in favor of #744

PR #563, opened 2026-03-12
fix: KymaEnvironementBinding orphaned bindings on kube.Status.Update() error

enrico-kaack-comp:
In theKymaEnvironmentBinding Logic that handles the rotation, we have a bug that lead to orphaned bindings in the external system.

In Create(), we create a new Binding, than write the binding id into the status field. If the status write fails, we return an error and loose the information of the created binding --> we dont track it anymore, see here. This might be beause of a conflicting resource.
```go
func ... Create() {
	clientBinding, err := c.client.CreateInstance(ctx, cr.Spec.KymaEnvironmentId, ttl)
	if err != nil {
		return managed.ExternalCreation{}, errors.Wrap(err, errCreate)
	}
	// Create new binding from client binding
	newBinding := v1alpha1.Binding{
// ...
	}
	// Add new binding to status
	cr.Status.AtProvider.Bindings = append(cr.Status.AtProvider.Bindings, newBinding)

	return managed.ExternalCreation{
		ConnectionDetails: connectionDetails,
	}, errors.Wrap(c.kube.Status().Update(ctx, cr), errStatusUpdate)
}
```

## Improvement

this pr introduces a retry with exponential backoff to retry the kubernetes status update on conflict errors. For that, it makes a get for the resource to have the current resourceVersion for the update. If this fails in create, it will try to delete the external Binding again to prevent orphaned resources.
If this still fails we have the orphaned resource but this will be unlikely.

## Long Test Run
(since kyma is skipped in PR tests due to long runtime): TestKymaEnvironment ✅ https://github.com/SAP/crossplane-provider-btp/actions/runs/24136957969/job/70427571445

Long: https://github.com/SAP/crossplane-provider-btp/actions/runs/24142182042

## Reference

Quote from @jtucci PR #618  adressing this:
> Problem
After upgrading from Crossplane 1.19.1 to 1.20.5, KymaEnviromentBinding resources repeatedly created bindings until hitting the 10-binding API limit, leaving all resources in failed state.

> When Create() succesfully provisions a binding via the api but fails to write the binding id back to CR status, the binding becomes orphaned. It's invisible to the controller but still consuming a slot. On the next reconcile the controller sees empty status, sets ResourceExists=false, and fires Create() again. This repeats untill the quota is exhausted.

> Root cause:
> In crossplane v1.19, SetConditions(ReconcileSuccess()) on an already-successful resource was a no-op. Condition.Equal() didn't compare ObservedGeneration, so the condition wasn't replaced and no real watch event was emitted. In v1.20 the new ObservedGenerationPropagationManager stamps a fresh ObservedGeneration on every condition, and Equal() now includes that field, so every reconcile produces a genuine status write and a corresponding watch event.

> The extra watch traffic causes the informer cache to lag the API server's resourceVersion more frequently. Between Observe and Create, the reconciler performs three metadata/status writes (finalizer, ExternalCreatePending anotation, observe-phase status) each bumping the resourceVersion. When Create() then attempts it's own Status().Update() using the stale in-memory object, the API server returns a 409 Conflict and the binding ID is never persisted.

>The original Create() did not retry or handle this conflict:

> cr.Status.AtProvider.Bindings = append(cr.Status.AtProvider.Bindings, newBinding)
return managed.ExternalCreation{...}, errors.Wrap(c.kube.Status().Update(ctx, cr), errStatusUpdate)
When the status write failed, the binding ID was never persisted to the CR. The next
Observe() saw no bindings in status, returned ResourceExists: false, and Create() was
called agian... creating another orphaned binding on the provisioning API. This repeated until
the 10-binding limit was hit.

> Fix:
> updateStatusWithRetry: re-fetches the CR on every attempt (including the first) to get the latest resourceVersion, retries up to 5 times on 409 conflict errors with exponential backoff , respects context cancellation, and returns immediately on non-conflict errors
If all retries are exhausted, rolls back the provisioning API binding via DeleteInstances to prevent it from becoming an orphaned slot
Adds unit tests covering: success on first attempt, conflict resolved on retry, retries exhausted with rollback, retries exhausted with rollback failure, re-fetch failure, and context cancellation during backoff


jtucci:
@enrico-kaack-comp @sdischer-sap Any movement on this? We're seeing significant impact in production. Every one of our users Kyma clusters has hit the 10-binding limit, completely blocking new deployments into those clusters. This includes certificate renewals that external customer-facing applications rely on, meaning live services are at risk if certificates expire and cannot be reissued. As a workaround we're having to instruct our users to manually remove these bindings which is disruptive and causing friction. 

The retry PR is a step in the right direction but I'm not sure it will fully close the gap. If all retries fail the binding is still orphaned, and there's no recovery path for bindings that are already lost right? Have you considered having `Create` explicitly calling `DeleteInstances` to roll back the BTP-side binding before returning an error, making the operation truly atomic ?

Happy to help, let me know if there is anything I can do to move this forward :)

internal/controller/kymaenvironmentbinding/kymaenvironmentbinding.go
```diff
@@ -204,9 +228,18 @@ func (c *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
 		"kubeconfig": []byte(clientBinding.Credentials.Kubeconfig),
 	}
 
+	// Try to update status with retry
+	statusErr := c.updateStatusWithRetry(ctx, cr, 5)
+	if statusErr != nil {
+		// Status update failed even after retries - store the binding ID
```
jtucci:
```suggestion
	if statusErr != nil {
      // All retries exhausted so roll back the BTP binding to prevent orphaning                                                                                                                                                                     
      _ = c.client.DeleteInstances(ctx, []v1alpha1.Binding{{Id: newBinding.Id}}, cr.Spec.KymaEnvironmentId) 
```

enrico-kaack-comp:
@jtucci thank you for raising prio. I will try to get this one fixed quickly.

wagneran:
+1, has potential to cause a total outage.

enrico-kaack-comp:
Integrated the changes from @jtucci in #618

internal/controller/kymaenvironmentbinding/kymaenvironmentbinding.go
```diff
@@ -65,6 +74,40 @@ func (c *external) Disconnect(ctx context.Context) error {
...
+			// Re-fetch to resolve resource version conflicts before retrying
+			if err := c.kube.Get(ctx, types.NamespacedName{Name: cr.Name, Namespace: cr.Namespace}, cr); err != nil {
+				return errors.Wrap(err, "re-fetch before status retry failed")
+			}
+		}
+
+		mutate(cr)
+
+		lastErr := c.kube.Status().Update(ctx, cr)
+		if lastErr == nil {
```
gergely-szabo-sap: Do you mean
```
lastErr = c.kube.Status().Update(ctx, cr)
```
?

internal/controller/kymaenvironmentbinding/kymaenvironmentbinding.go
```diff
@@ -204,9 +257,26 @@ func (c *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
...
+	if statusErr != nil {
+		c.record.Event(cr, event.Warning(reasonStatusUpdate, errors.Wrap(statusErr, "while updating status during creation - rolling back binding (binding will be deleted again)")))
+		// Status update failed after all retries. Roll back the binding on the provisioning API
+		// to prevent it from becoming orphaned and consuming quota.
+		rollbackErr := c.client.DeleteInstances(ctx, []v1alpha1.Binding{newBinding}, cr.Spec.KymaEnvironmentId)
+		if rollbackErr != nil {
+			return managed.ExternalCreation{}, errors.Wrap(rollbackErr, "failed to roll back binding after status update failure: "+errCreate)
+		}
+		return managed.ExternalCreation{
+			ConnectionDetails: connectionDetails,
```
gergely-szabo-sap: Is `ConnectionDetails` valid here?

gergely-szabo-sap (changes requested):

internal/controller/kymaenvironmentbinding/kymaenvironmentbinding.go
enrico-kaack-comp: important catch, really good one thank you :)

internal/controller/kymaenvironmentbinding/kymaenvironmentbinding.go
enrico-kaack-comp: not valid at this point correct. Since crossplane doesnt update the connection details on error, it would not have an impact but it is more correct return nil here.

gergely-szabo-sap (approved):
lgtm

wagneran:
Hi @enrico-kaack-comp - why was @jtucci original PR #618 closed and this merged with no co-author? Some of the patterns used in #618 exhibit a design pattern that look robust (especially around the test harness) and should be considered.

enrico-kaack-comp:
@wagneran it was merged with coauthor. I just checked, the merge commit has 
```
Co-authored-by: jtucci
```
in it. I forgot to put the email in it, my bad, sorry. Not sure if I can fix this after.

Why the original PR was closed and not this one? 
Because one Pr needs to be closed, I focused on getting the fix merged as it seems pressing and timezone difference made communication time-consuming. I decided for this branch as I thought it is easier to work on non-forks. 

I will recheck the #618 for the test patterns.

PR #572, opened 2026-03-13
chore: External Name support for CertBasedOIDCLogin and KubeConfigGenerator (not applicable)

enrico-kaack-comp:
CertBasedOIDCLogin and KubeConfigGenerator dont generate external resources so they dont have an external name.
They also dont need one as you can simply create a new one instead of importing with external name.

This PR describes this in the docs.


sdischer-sap (approved):
Perfect, lgtm

PR #600, opened 2026-03-27
fix: entitlement external resource got deleted when it should be updated

### LINKED ISSUE #597: [BUG]: Entitlement quota zeroed out when deleting one of multiple Entitlement CRs for the same service plan
**Describe the bug**
When multiple `Entitlement` CRs exist for the same subaccount/service/plan and one of them is deleted, the entitlement quota in BTP is transiently set to `0` instead of the correct remaining value (sum of the surviving CRs). The surviving CRs will eventually self-heal by reconciling and restoring the correct quota, but there is a window during which the quota is incorrectly zero, which may cause service disruptions and resources getting stuck. 

v1.x.x

**To Reproduce**
1. Create two or more `Entitlement` CRs targeting the same subaccount, service name, and service plan, each with a numeric `amount` (e.g. CR-A with `amount: 5`, CR-B with `amount: 3`).
2. Wait for both CRs to reach a synced/ready state — BTP should reflect the combined quota (e.g. `8`).
3. Delete one of the CRs (e.g. CR-A).
4. Observe the BTP entitlement quota for that service plan immediately after the deletion is processed.

**Expected behavior**
The BTP quota should be updated to the sum of the remaining CRs (e.g. `3`) without passing through zero. The surviving CR should continue to manage its portion of the quota uninterrupted.

**Actual behavior**
The BTP quota is briefly set to `0` during deletion. The surviving CR eventually reconciles and restores the correct value, but the transient zero can cause service disruptions in the interim.

**Additional context**
This affects numeric-quota entitlements only (i.e. those using `.spec.forProvider.amount`). Enable-based (non-numeric) entitlements are not affected because the controller already returns early without calling the BTP API when sibling CRs exist.

**Root Cause**
The root cause is in internal/clients/entitlement/entitlement.go. When `DeleteInstance` is called, `GenerateObservation` has already computed the correct remaining quota via `MergeRelatedEntitlements` (which sums amounts from surviving CRs and stores the count in `EntitlementsCount`). However, `DeleteInstance` then unconditionally overwrites         `Required.Amount` with 0 before calling `UpdateInstance`, discarding the correctly merged value:
```go                                                                                                                                                                                
  // internal/clients/entitlement/entitlement.go
  if isNumericQuota {
      amount := 0
      cr.Status.AtProvider.Required.Amount = &amount  // always zeros, even when siblings exist
  }                                                                                                                                                                             
```
  The fix is to only zero the amount when no sibling CRs remain.                                                                                                                

```go
  Fix:
  if isNumericQuota {
      // Only zero out if no other entitlements remain. If survivors exist,
      // Required.Amount already holds the correct merged value from GenerateObservation.
      if relatedEntitlements == nil || *relatedEntitlements == 0 {
          amount := 0
          cr.Status.AtProvider.Required.Amount = &amount
      }
  }

```                  
A fix for this is already included in PR #527.

dattito:
Fixes #597.

Cherry-picked the commit from #527.


enrico-kaack-comp (approved):
thank you for picking this out of the bigger PR.

Did you manually test this behaviour? Do you want to add unit tests for this?

PR #606, opened 2026-04-01
fix: `KymaEnvironment` connection secret empty when in state drift

### LINKED ISSUE #483: [BUG] KymaEnvironment - empty connectionDetails / connection secret on drift
**Describe the bug**

`KymaEnvironment` is configured with `managementPolicies: ["Observe"]`.
The connection details are empty if there is *any* drift between resource and observed state

v1.5.0

**To Reproduce**

1. create a `KymaEnvironment` resource for an existing environment, eg.

    ```yaml
    apiVersion: environment.btp.sap.crossplane.io/v1alpha1
    kind: KymaEnvironment
    metadata:
      name: "example"
      annotations:
        crossplane.io/external-name: "ENVIRONMENT-UUID-HERE"
    spec:
      deletionPolicy: Orphan
      managementPolicies:
        - Observe
      providerConfigRef:
        name: example
      subaccountRef:
        name: example
      cloudManagementRef:
        name: example-cis
      writeConnectionSecretToRef:
        name: "default"
        namespace: example-kyma
      forProvider:
        name: "example"
        planName: azure
        parameters:
          # current environment parameters
    ```

2. wait until ready/synced
3. modify any of the `parameters` or open the environment in the BTP Cockpit and manually update the environment (e.g. add  new `adminstrator`)
4. delete connection secret `example-kyma` and wait until it's recreated
5. 💥 recreated connection secret has no keys / no data
6. 💥 in case of using the `KymaEnvironment` in a Crossplane composition, the `connectionDetails` of the observed resource
   are empty too.

**Expected behavior**

`connectionDetails` and connection secret always populated, regardless of state drift

**Screenshots**

N/A

**Additional context**

[SAP-internal Slack discussion](https://sap-cx.slack.com/archives/C03S7EFAZJ9/p1769444144506129)

enrico-kaack-comp:
Close #483 

### Problem

If the KymeEnvironment falls in the needUpdate case, the connection data/secret is not set (early return). If in the meantime the secret gets deleted, it is recreated empty.

Question: Before merging, is this actually intended behaviour and we dont need this fix?

### Changes

**`internal/controller/environment/kyma/kymaenvironment.go`**
- Moved `GetConnectionDetails` to execute **before** the drift check so connection details are always available
- When drift is detected (`needsUpdate=true`) and connection details were fetched successfully, they are now included in the returned `ExternalObservation` — fixing the bug
- When drift is detected but connection details could not be read (e.g. environment still initializing, no labels yet), a Kubernetes warning event is emitted via `c.record` so the situation is visible in `kubectl describe`, and the observation still returns `ResourceUpToDate: false` to trigger the normal reconcile loop without surfacing the read error
- When no drift and connection details fail, the original error-return behavior is preserved
- Added `record event.Recorder` to `connector` and `external` structs to enable event emission

**`internal/controller/environment/kyma/zz_setup.go`**
- Construct an `event.Recorder` from `mgr.GetEventRecorderFor(name)` and pass it into `connector`

**`internal/controller/environment/kyma/zz_connect.go`**
- Forward `c.record` from `connector` to `external` in `Connect()`

**`internal/controller/environment/kyma/kymaenvironment_test.go`**
- Added regression test `DriftWithConnectionDetails` that asserts connection details are populated even when `ResourceUpToDate: false`

### Behaviour summary

| State | Before | After |
|---|---|---|
| No drift, get connection data succeed | ✅ Secret populated | ✅ Secret populated |
| **Drift, get connection data succeed** | ❌ Secret empty | ✅ Secret populated |
| Drift, get connection data failed | ❌ Secret empty, no indication | ✅ Secret empty, warning event emitted |
| No drift,get connection data failed | ❌ Error returned | ❌ Error returned (unchanged) |

Long running E2E test that covers Kymaenvironment https://github.com/SAP/crossplane-provider-btp/actions/runs/23858584110


sdischer-sap (approved):
lgtm

PR #618, opened 2026-04-07
fix: prevent orphaned KymaEnvironmentBinding bindings 

jtucci:
# Problem 

After upgrading from Crossplane 1.19.1 to 1.20.5, KymaEnviromentBinding resources repeatedly created bindings until hitting the 10-binding API limit, leaving all resources in failed state.

When Create() succesfully provisions a binding via the api but fails to write the binding id back to CR status, the binding becomes orphaned. It's invisible to the controller but still consuming a slot. On the next reconcile the controller sees empty status, sets ResourceExists=false, and fires Create() again. This repeats untill the quota is exhausted.

## Root cause: 

In crossplane v1.19, `SetConditions(ReconcileSuccess()) `on an already-successful resource was a no-op. `Condition.Equal()` didn't compare ObservedGeneration, so the condition wasn't replaced and no real watch event was emitted. In v1.20 the new ObservedGenerationPropagationManager stamps a fresh ObservedGeneration on every condition, and `Equal() `now includes that field, so every reconcile produces a genuine status write and a corresponding watch event.

The extra watch traffic causes the informer cache to lag the API server's resourceVersion more frequently. Between Observe and Create, the reconciler performs three metadata/status writes (finalizer, ExternalCreatePending anotation, observe-phase status) each bumping the resourceVersion. When `Create()` then attempts it's own `Status().Update()` using the stale in-memory object, the API server returns a 409 Conflict and the binding ID is never persisted.

The original `Create()` did not retry or handle this conflict:

```go
cr.Status.AtProvider.Bindings = append(cr.Status.AtProvider.Bindings, newBinding)
return managed.ExternalCreation{...}, errors.Wrap(c.kube.Status().Update(ctx, cr), errStatusUpdate)
```

When the status write failed, the binding ID was never persisted to the CR. The next
`Observe()` saw no bindings in status, returned `ResourceExists: false`, and `Create()` was
called agian... creating another orphaned binding on the provisioning API. This repeated until
the 10-binding limit was hit.

# Fix:
- updateStatusWithRetry: re-fetches the CR on every attempt (including the first) to get the latest resourceVersion, retries up to 5 times on 409 conflict errors with exponential backoff , respects context cancellation, and returns immediately on non-conflict errors
- If all retries are exhausted, rolls back the provisioning API binding via DeleteInstances to prevent it from becoming an orphaned slot
- Adds unit tests covering: success on first attempt, conflict resolved on retry, retries exhausted with rollback, retries exhausted with rollback failure, re-fetch failure, and context cancellation during backoff

### DIFF
--- a/internal/controller/kymaenvironmentbinding/kymaenvironmentbinding.go
+++ b/internal/controller/kymaenvironmentbinding/kymaenvironmentbinding.go
@@ -8,7 +8,9 @@ import (
 
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/pkg/errors"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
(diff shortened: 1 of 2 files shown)


enrico-kaack-comp:
Tracked in #563

PR #622, opened 2026-04-13
docs: update image

jn-av:
In our documentation, the guide still shows the message queue in the image. This has been updated to HANA to reflect the current guide and the direction we want to move forward.


jn-av:
Will be added by @lennardrother

PR #623, opened 2026-04-14
fix: service binding rotation status update

enrico-kaack-comp:
Related follow up to #563 

Basically in our rotation concept, we update the Status of the resource in Observe and Create. But we do not handle a 409 conflict that might occur when the resource was modified in the meantime by e.g. crossplane runtime itself.

This PR refactors the retry logic in #563 and than changes the keyrotator logic of service binding rotation to make the `RetireBinding` function idempotent for retrying.

Long E2E test (since it also refactores KymaEnvBinding): ✅  https://github.com/SAP/crossplane-provider-btp/actions/runs/24825574948


sdischer-sap (approved):
LGTM

enrico-kaack-comp:
Long E2E failing for Directory Entitlement, probably flaky, retry it

enrico-kaack-comp:
Long E2E finally worked in https://github.com/SAP/crossplane-provider-btp/actions/runs/24825574948

PR #624, opened 2026-04-14
docs: Use SAP HANA Cloud as example for services section

lennardrother:
- Use SAP HANA Cloud and SAP HANA Cloud Administration Tools as examples for service instance and subscription instead of SAP Message Queuing service and SAP Business Application Studio
- Remove duplicated content from consume-service.mdx and link to HANA provider docs instead
- Fix typos, formatting and broken links


jn-av (approved):
Thanks a lot! looks great!

PR #376, opened 2025-10-21
chore: Update git submodule `build` url

paulchen5:
The https://github.com/upbound/build repo has been archived and the author recommends to switch to https://github.com/crossplane/build. 

I used the provided commands to update the submodule url and sync it's contents.

```bash
git submodule set-url -- build https://github.com/crossplane/build
git submodule sync
cd build
git checkout main
git pull
```

However it seemed to work for you even without the update

Issue #336, opened 2025-09-15
[TASK] migrate build submodule to crossplane build for go 1.25 upgrade


christophrj:
### Context

Our build directory points to [upbound build](https://github.com/upbound/build/tree/3b99463225581259ce39c7d7a45290be12515abb) that has been archived and replaced with [crossplane build](https://github.com/crossplane/build).

This prevents us from e.g. upgrading to golangci-lint v2.x since upbound build is using the [--out-format flag](https://github.com/upbound/build/blob/3b99463225581259ce39c7d7a45290be12515abb/makelib/golang.mk#L118) that has been removed with [v2](https://golangci-lint.run/docs/product/migration-guide/#--out-format).

Relates to: #82 

### Acceptance Criteria
- [ ] our build directory points to [crossplane build](https://github.com/crossplane/build)

sdischer-sap:
has been done some time ago

PR #626, opened 2026-04-15
chore: migrate to single-image build and crossplane/build submodule

### LINKED ISSUE #82: [Task] Single-Image Builds
**Context**
Crossplane at some point change the way **how images are being built.** The initial approach of having 2 images (one containing just the CRDs and a link to the second one) has been changed to to a **single build process**.

Makes sense to adapt that to our builds as well. Like usual we can do it once for the btp-provider and apply it later to the other ones.

**Acceptance Criteria**
- [ ] releases in registry contain just one image
- [ ] MCP can install those releases properly

jtucci:
Closes #82, #336, #604, supersedes #376

My team has deadlines around Crossplane v2 migration planning and we rely heavily on this provider, so wanted to help get things unblocked.

When we stood up a v2 test cluster the provider pods wouldn't start. Took some digging but the root cause is that Crossplane v2 removed `spec.controller.image` from the package API (crossplane/crossplane#6487, shipped in v2.0.0). Without that field, Crossplane falls back to using the package image itself as the runtime but the package image was `FROM scratch` with only `package.yaml` in it, so there was nothing to execute and the container failed immediately.

This PR is not about implementing v2 features like namespace-scoped resources, its just getting the provider to a point where it actually installs and runs on v2.

Followed the same approach as crossplane-provider-cloudfoundry in SAP/crossplane-provider-cloudfoundry#101.

**Changes**

- `.gitmodules` / `build`: switched from archived `upbound/build` to `crossplane/build`
- `package/crossplane.yaml`: bumped to `meta.pkg.crossplane.io/v1`, removed `spec.controller.image`. Runtime is now embedded in the `.xpkg` by `crossplane xpkg build`
- `cluster/images/`: replaced `crossplane-provider-btp` + `crossplane-provider-btp-controller` with a single `provider-btp/` image using `imagelight.mk`
- `Makefile`: migrated from `image.mk` to `imagelight.mk`, replaced custom `publish` target with `xpkg.build` + `xpkg.release.publish`, added `local-deploy` target for local e2e
- Release workflows: replaced `make build` + `make publish` with per-platform `make xpkg.build` + `make xpkg.release.publish`, swapped `DOCKER_REGISTRY`/`BUILD_REGISTRY` for `XPKG_REG_ORGS`
- Test suite: removed all controller image references (`UUT_CONTROLLER`, `ControllerImage`, `fromControllerRepository` etc.) from e2e and upgrade tests; adopted `E2E_REUSE_CLUSTER` pattern so the provider is pre-deployed via `local-deploy` rather than having `xp-testing` install it. `xp-testing`'s `loadCrossplanePackageToCluster` expects a standalone package OCI image containing `package.yaml` which doesnt exist in single-image builds. Happy to discuss if `xp-testing` should be updated to handle this natively instead

Also fixes #604. The `io.crossplane.xpkg` label was missing because the package was built with plain `docker build`, its now set automatically by `crossplane xpkg build`.

**Testing**

Tested locally with `make test-acceptance` against kind + Crossplane 1.20.1. Provider reaches Healthy. Short e2e tests pass. Some tests timed out waiting on BTP resources but I havent verified if those are pre-existing so flagging it in case.

Also deployed on a Gardener cluster running Crossplane v2 with active provider-btp managed resources. Provider installed and existing resources kept reconciling fine.

I noticed the same issue likely affects other SAP providers that havent made this migration yet. Happy to contribute similar PRs if useful, just let me know :)


enrico-kaack-comp:
Thank you for your contribution.
@schlingemann since you have been doing the single image build in other providers, can you please take care of this contribution?

sdischer-sap:
Looks like there are some issues with applying the builds in the e2e tests. @schlingemann do you mind checking it?

schlingemann:
Yes, I'm taking a look!

schlingemann:
@jtucci I made some changes, as the build was previously failing both on my local ARM mac and in the action runners. It now works for me locally, and it seems like it's also working in the action. Could you verify that my changes are also compatible with your local setup / platform?

I also made sure that the xpkg build runs during local-setup, so targets like `test-acceptance` can be used right away. That's how it worked previously too, I believe.

Please take note of the CLA btw :)

jtucci:
@schlingemann yup!, just re-ran locally and everything is still working for me. 

The CLA issue is because my commit has two different GitHub identities on it. authored as jon-tucci which is tied to my sap email, but committed as jtucci (my main account which already signed the CLA). I was going to amend the commit to use a single identity and force pus to the fork, you'll need to rebase your two commits on top. Sorry for the hassle :(  let me know if that's ok before I do it!

schlingemann:
Sure, go ahead with the force push 😄

schlingemann (approved):

sdischer-sap (approved):
Nice work, thank you!

PR #630, opened 2026-04-16
fix: kymabinding rotation leaves orphaned expired bindings in external system

### LINKED ISSUE #619: KymaBinding Rotation expired bindings are not explicitly deleted at the BTP service level (leaked resources)
## Summary

When a `KymaEnvironmentBinding` rotates or expires, the controller silently drops the old binding from its status without calling `DeleteEnvironmentInstanceBinding` on the BTP provisioning API. It seems the btp is not really cleaning this up automatically. Even if, we should keep it clean.

## Current Behavior

In `validateBindings`, TTL-expired bindings are filtered out of `validBindings` and removed from status on the next reconcile:

```go
// Second pass: keep non-expired bindings (active or inactive)
for _, b := range bindings {
    if !ttlIsExpired(&b, now) {
        validBindings = append(validBindings, b)
    }
}
```

No `DeleteEnvironmentInstanceBinding` API call is made before the binding ID is dropped from status. Once removed from status, the binding ID is lost and can never be cleaned up by the controller.

## Risk

If the BTP service does not reliably enforce the TTL, bindings will accumulate and never be cleaned up. BTP has a limit on non-expired bindings and seems to not enforce a limit on expired bindings but still we leak external resources.

## Expected Behavior

Before removing an expired or rotation-deactivated binding from status, the controller should call `DeleteEnvironmentInstanceBinding` for that binding ID, similar to what `Delete` already does for CR teardown.

## Suggested Fix

In `validateBindings` (or its caller `Observe`), collect bindings that are about to be dropped and issue delete calls against the BTP API before removing them from status.

enrico-kaack-comp:
Fix #619

## Problem descirption from #619 
When a KymaEnvironmentBinding rotates or expires, the controller silently drops the old binding from its status without calling DeleteEnvironmentInstanceBinding on the BTP provisioning API. It seems the btp is not really cleaning this up automatically. Even if, we should keep it clean.

## Solution
Collect expired bindings, call Delete on those. If the delete failed, log an event with the binding id for manual cleanup but do not treat it as an error

## Long E2E Test
✅  https://github.com/SAP/crossplane-provider-btp/actions/runs/24676233390


SatabdiG (approved):
lgtm

PR #631, opened 2026-04-21
fix: throw error when SubaccountApiCredential secret is missing clien…

### LINKED ISSUE #444: [TASK] `SubaccountApiCredential` throw an error when secret contains only three keys
For a `SubaccountApiCredential` to work, it has to create a secret containing **four** keys. 

The key `client_secret` is only being returned during an initial CREATE against the BTP CLI. Subsequent GET requests fail to retrieve this key. 

This is an issue when you have an orphaned/already existing ApiCredential on BTP side, causing the  `SubaccountApiCredential` to look healthy but actually it is not, because the `client_secret` could not be returned. 

Therefore when creating a secret using the `SubaccountApiCredential` its controller should check, whether the secret to be created contains four keys and if not, raise an error message like 
`"can not read client_secret from source, please delete external resource and re-create Crossplane resource"`. 

ref
- https://github.com/SAP/crossplane-provider-btp/issues/434

1122Louis:
Fixes #444

When a SubaccountApiCredential is orphaned (exists in BTP but not managed by Crossplane), the BTP API returns only 3 keys on GET,- omitting `client_secret`. This caused the resource to appear `Ready=True` with an incomplete connection secret.

This PR adds a check in `Initialize()` that verifies the connection secret contains `attribute.client_secret`. If missing, it returns an error so Crossplane marks the resource as not ready. Certificate-based credentials (where `client_secret` is never present) are skipped.

Also fixes a pre-existing nil pointer dereference when `SubaccountRef` is nil but `SubaccountID` is set.

### DIFF
--- a/config/btp_subaccount_api_credential/config.go
+++ b/config/btp_subaccount_api_credential/config.go
@@ -1,107 +1,124 @@
 package btp_subaccount_api_credential
 
 import (
-	"context"
+"context"
 
-	"github.com/pkg/errors"
+"github.com/pkg/errors"
 
-	"github.com/crossplane/crossplane-runtime/pkg/meta"
-	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
-	"github.com/crossplane/crossplane-runtime/pkg/resource"
-	"github.com/crossplane/upjet/pkg/config"
-	"sigs.k8s.io/controller-runtime/pkg/client"
+"github.com/crossplane/crossplane-runtime/pkg/meta"
+"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
+"github.com/crossplane/crossplane-runtime/pkg/resource"
+"github.com/crossplane/upjet/pkg/config"
+corev1 "k8s.io/api/core/v1"
+"sigs.k8s.io/controller-runtime/pkg/client"
 
-	accountsv1alpha1 "github.com/sap/crossplane-provider-btp/apis/account/v1alpha1"
-
-	securityv1alpha1 "github.com/sap/crossplane-provider-btp/apis/security/v1alpha1"
-	providerv1alpha1 "github.com/sap/crossplane-provider-btp/apis/v1alpha1"
-	"github.com/sap/crossplane-provider-btp/internal/tracking"
+accountsv1alpha1 "github.com/sap/crossplane-provider-btp/apis/account/v1alpha1"
+securityv1alpha1 "github.com/sap/crossplane-provider-btp/apis/security/v1alpha1"
+providerv1alpha1 "github.com/sap/crossplane-provider-btp/apis/v1alpha1"
+"github.com/sap/crossplane-provider-btp/internal/tracking"
 )
 
 const (
-	errTrackRUsage   = "cannot track ResourceUsage"
-	errTypeAssertion = "managed resource is not of type SubaccountApiCredential"
+errTrackRUsage         = "cannot track ResourceUsage"
+errTypeAssertion       = "managed resource is not of type SubaccountApiCredential"
+errMissingClientSecret = "can not read client_secret from source, please delete external resource and re-create Crossplane resource"
 )
 
 // Configure configures individual resources by adding custom ResourceConfigurators.
 func Configure(p *config.Provider) {
-	p.AddResourceConfigurator("btp_subaccount_api_credential", func(r *config.Resource) {
-		r.ShortGroup = "security"
-		r.Kind = "SubaccountApiCredential"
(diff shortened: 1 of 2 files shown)


sdischer-sap (changes requested):
Please format the file, because right now its barely readable. 
From what I read in the description and in the test looks like you are on the right track though.

config/btp_subaccount_api_credential/config.go
```diff
@@ -103,5 +104,21 @@ func (d *DeletionProtectionInitializer) Initialize(ctx context.Context, mg resou
 			return errors.New(providerv1alpha1.ErrResourceInUse)
 		}
 	}
+
+	if cr.Spec.ForProvider.CertificatePassed == nil {
```
sdischer-sap, line 112: Don't really get why `CertificatePassed` should we relevant for this issue? Do you have some reasoning why only in cassed this is empty this check should apply?

config/btp_subaccount_api_credential/config.go
```diff
@@ -103,5 +104,21 @@ func (d *DeletionProtectionInitializer) Initialize(ctx context.Context, mg resou
...
+		secretRef := cr.GetWriteConnectionSecretToReference()
+		if secretRef != nil && secretRef.Name != "" {
+			secret := &corev1.Secret{}
+			err := d.Kube.Get(ctx, client.ObjectKey{
+				Name:      secretRef.Name,
+				Namespace: secretRef.Namespace,
+			}, secret)
+			if err == nil {
+				if _, ok := secret.Data["attribute.client_secret"]; !ok {
```
sdischer-sap: Did you test this end to end? 
I am not sure if there might not be some other edge cases in which a secret is generally available, but empty. If there are those cases this would prevent a creation entirely. 
I would be a bit more persimistic with narrowing the error scenario here and check that those other 3 attributes are present and only the secret is missing to make sure it really is exactly this use case.

sdischer-sap (changes requested):
Great that you found this approach. 
I would change some details though to make it more robust (see remarks).

config/btp_subaccount_api_credential/config.go
1122Louis: Thanks for your comment.... According to the Terraform BTP provider docs (https://registry.terraform.io/providers/SAP/btp/latest/docs/resources/subaccount_api_credential), client_secret is only generated "if the certificate is omitted". So when certificate_passed is set, no client_secret is ever created and the check would always fail for certificate-based credentials even when they're healthy. That's why I added the guard to skip the check in that case.

sdischer-sap: Ah nice catch then. Then nevermind my comment, didn't know that. 
Although maybe you can add a code comment saying that according to docs this is how it works, so somebody else might not take this false assumption as me and potentially get tempted to delete that.

config/btp_subaccount_api_credential/config.go
1122Louis: Thank you very much, I went ahead and updated it! The check now only triggers when client_id, token_url, and api_url are all there but client_secret isn't , so it should only catch the orphaned credential case. Added a unit test too to make sure a half-filled secret doesn't cause issues.

I also tried to run e2e test locally by setting up a kind cluster with the provider installed and real BTP credentials, but ran into a 400 Bad Request from the BTP API when creating the SubaccountApiCredential, which seems to be an entitlement issue in my test subaccount, unrelated to this change. Since my fix only affects the Initialize() logic that runs when reading an existing secret, the unit tests cover the relevant scenario directly.

Let me know if you have any other feedback

config/btp_subaccount_api_credential/config.go
1122Louis: Done, added a link to the Terraform BTP provider docs in the comment

config/btp_subaccount_api_credential/config.go
sdischer-sap: Nice thank you. 
I added a comment here to add verification of this error behaviour as part of the import testing that is part of that ticket: https://github.com/SAP/crossplane-provider-btp/issues/553#issuecomment-4302315857. 

So no implementation work left here. However running the e2e test locally is something you will need at some point, so I would suggest to dig a bit deeper why those are failing at some point.

sdischer-sap (approved):
Nice work!
LGTM now

config/btp_subaccount_api_credential/config.go
1122Louis: Thank you for the review! I'll look into running the E2E tests locally to understand the failures better.

PR #634, opened 2026-04-23
fix: flatten connection details in ServiceBinding Create path

### LINKED ISSUE #633: ServiceBinding secret briefly contains attribute.credentials instead of flattened keys
After a ServiceBinding is created, the connection secret temporarily contains a raw `attribute.credentials` key with the full JSON blob, instead of the expected flattened individual keys (`clientid`, `clientsecret`, `sm_url`, etc.).

This happens because `flattenSecretData` is called in the `Observe` path but not in the `Create` path. Since the controller uses synchronous TF apply (`useAsync=false`), `Create` returns connection details immediately — unflattened. The secret is corrected on the next reconcile when `Observe` runs, but any consumer reading the secret in between sees the wrong format.

Noel-Ch:
Fixes #633

## Summary
- Apply `flattenSecretData` to connection details returned from `Create`, matching the existing behavior in `Observe`
- Prevents `attribute.credentials` from briefly appearing in the connection secret after creation

### DIFF
--- a/internal/controller/account/servicebinding/servicebinding.go
+++ b/internal/controller/account/servicebinding/servicebinding.go
@@ -217,6 +217,11 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
 		return managed.ExternalCreation{}, errors.Wrap(err, errCreateBinding)
 	}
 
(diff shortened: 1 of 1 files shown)


sdischer-sap (approved):
lgtm, thank you!

PR #640, opened 2026-04-28
chore: KymaEnvironmentBinding external name

enrico-kaack-comp:
add documentation for KymaEnvironmentBinding that they dont support external names.


sdischer-sap (approved):
lgtm

PR #643, opened 2026-04-29
fix: upgrade test after single image build 

enrico-kaack-comp:
after #626 , the upgrade test setup fails. This fix it

### DIFF
--- a/.github/workflows/cleanup.go
+++ b/.github/workflows/cleanup.go
@@ -14,30 +14,33 @@ import (
 	"time"
 )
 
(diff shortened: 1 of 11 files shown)


sdischer-sap:
@enrico-kaack-comp interesting problem. 

I guess we now would have an issue with testing from a release with 2 images to one with one right? 
I mean the fix you provided makes it work for older release, but starting from the first with that build it would probably fail again?

schlingemann (approved):

PR #646, opened 2026-05-03
Feat/553 external name compliance subaccount api credential

### LINKED ISSUE #553: [TASK] `External-Name`(SubaccountApiCredential): Ensure ADR compliance
**Part of Epic: #427** 
## Context

## Technical Tasks Checklist

**Acceptance Criteria:**

- [ ] **Migration Logic**:

**Acceptance Criteria:**
- [ ] **Observe() Tests**:

- [ ] **Create() Tests**:

- [ ] **Update() Tests**:

- [ ] **Delete() Tests**:

**Acceptance Criteria:**
- [ ] **Import Flow Test**:

1122Louis:
Closes #553

What I found
Before making any changes I wanted to understand what identifier the BTP Terraform provider actually uses for this resource. I ran the provider locally against BTP and checked the Terraform logs turns out it uses name as the resource id (id_key: name), not a GUID. So this resource is a known ADR exception, same as RoleCollection.

The existing config (SetIdentifierArgumentFn + GetExternalNameFn) was already doing the right thing. No CRUD changes or migration needed.

What I changed
Added the External-Name Configuration comment to zz_subaccountapicredential_types.go marking it as an ADR exception (Terraform constraint)
Ran make docs.generate-external-name to update external-name.md
Added an e2e test that verifies orphaned import fails with a clear client_secret error as Stephan suggested in #631
Added an upgrade test verifying the external-name (credential name) survives a provider upgrade unchanged

## Release Note

```release-note
feat(security): external-name ADR compliance for SubaccountApiCredential - import not supported due to BTP only returning client_secret at creation time

### CODE REVIEW

**Comment by enrico-kaack-comp** [comment]
So do i understand correctly: SubaccountAPICredential is only readable when it is created and not later? Than this needs to be documented for this resource as it does not allow to import the resource on a different system.

**Comment by 1122Louis** [comment]
> So do i understand correctly: SubaccountAPICredential is only readable when it is created and not later? Than this needs to be documented for this resource as it does not allow to import the resource on a different system.

Yes exactly. Updated the doc comment to explicitly state that import is not supported BTP only returns client_secret at creation time, so importing an existing credential will always result in a missing client_secret in the connection secret. The e2e test (TestSubaccountApiCredentialOrphanImportFails) also verifies this behavior.

File: docs/end-user-guides/import-landscape/external-name.md [hunk 1]
```diff
@@ -98,6 +98,18 @@ Instead of importing, create a new KymaEnvironmentBinding resource.
...
+
+- Follows Standard: no (ADR exception: the SAP BTP Terraform provider uses the credential name as its
+resource identifier. Import is not supported — BTP only returns client_secret at creation time and
+never on subsequent reads. Importing an existing credential will always result in a missing
+client_secret in the connection secret, making it unusable.)
+- Format: The name of the API credential (e.g. "my-api-credential"), set via spec.forProvider.name
+- How to find:
+
+  - UI: BTP Cockpit → Subaccount → Security → OAuth Clients → [Select Credential] → Name
+  - CLI: btp list security/api-credential --subaccount-id `<subaccount-id>` (field: name)
```
enrico-kaack-comp: since there is no import we dont need to document how to set the external name as this should not be set by the user

enrico-kaack-comp (approved):
please fix the comment, otherwise looks good, thank you :)

enrico-kaack-comp:
Tip: the run-make-reviewable pipeline can be fixed by running `make generate` locally and add the result. It will basically update external-name.md based on the api defintion comment

PR #651, opened 2026-05-06
fix: cache btpclient between reconciles to minimize login calls

enrico-kaack-comp:
Right now, on every reconcile of any managed resource that uses the btp api client (Entitlement, Subaccount, Directory, GlobalAccount, ServiceManager, KymaEnvironment, KymaEnvironmentBinding, CloudFoundryEnvironment), the `Connect` function creates a new `Client`. This Client performs an oauth flow automatically if necessary. Since we recreate the client all the time, it is doing a login on every reconcile.

The clients oauth flow does perform the oauth flow only if no token exist or if the token is expired/about to epire. So subsequent requests that reuse the same client dont trigger a new login request.

To do so, this PR introduces a Cache for the client. The Key is the ProviderConfigName. On match, it checks the sha256 hash of the secrets (service account and cis binding) against an in-memory vaue. Only than it is a cache hit, otherwise a miss.
On hit, it returns the client that already has the active oauth token inside (as pointer).

Furthermore, the cache tracks lastUsed of every entry and evicts old entries. This is done when an cache entry is read.

One design decision due to the ReadOnly mutex on read:
> 	// Note: On cold start, multiple goroutines may reach this point concurrently
	// for the same pcName, each creating their own client. This is acceptable —
	// last write wins, and the extra clients get GC'd. Holding the lock during
	// createFn() would block all controllers on a single client creation (which
	// involves network I/O on first use). In steady state, only one cached entry exists per pcName.

This is a thundering herd problem, aka all resources creating and logging in at the same time.

Local testing works: I only see one login call even for several minutes of use, compared to a new call per resource on the old version.

TODO: debug through the client library to fully verify the automatic oauth token refresh.

Reviewer Notes: please review carefully and also think of edge cases since we cache credentials we need to make sure they are only returned in the right moments

### DIFF
--- a/internal/controller/environment/cloudfoundry/cfenvironment.go
+++ b/internal/controller/environment/cloudfoundry/cfenvironment.go
@@ -17,6 +17,7 @@ import (
 	"github.com/sap/crossplane-provider-btp/apis/environment/v1alpha1"
 	providerv1alpha1 "github.com/sap/crossplane-provider-btp/apis/v1alpha1"
 	env "github.com/sap/crossplane-provider-btp/internal/clients/cfenvironment"
+	"github.com/sap/crossplane-provider-btp/internal/controller/providerconfig"
 	"github.com/sap/crossplane-provider-btp/internal/tracking"
 
 	"github.com/sap/crossplane-provider-btp/btp"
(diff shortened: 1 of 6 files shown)


sdischer-sap:
Close in favor of #744

Issue #645, opened 2026-04-29
[FEATURE] Add poll jitter to prevent thundering herd after restarts


gergely-szabo-sap:
**Context**

The BTP provider does not implement poll jitter. Without jitter, resources created at similar times requeue simultaneously, causing "thundering herd" spikes — especially after provider restarts when all resources are re-enqueued at once. At scale (500+ resources), this can briefly saturate both the internal rate limiter and external BTP APIs.

**Scope**

- Add `WithPollJitterHook` to the controller options in `cmd/provider/main.go`
- Expose `--poll-jitter-percentage` CLI flag (following provider-kubernetes pattern, default 10%)
- Apply jitter to all controllers (native + hybrid)

**Out of scope**

_Hint_: Smaller feature wishes tend to be implemented and shipped quicker!

- Adaptive poll intervals (longer-term, requires `WithPollIntervalHook`)
- Per-controller jitter settings

**Technical Steps**

1. Add `--poll-jitter-percentage` flag to `cmd/provider/main.go` (default 10, same as provider-kubernetes)
2. Compute jitter: `jitter := time.Duration(float64(*pollInterval) * float64(*pollJitterPercentage) / 100.0)`
3. Apply via `managed.WithPollJitterHook(jitter)` in controller options
4. Reference implementation: [provider-kubernetes `main.go`](https://github.com/crossplane-contrib/provider-kubernetes/blob/main/cmd/provider/main.go) exposes `--poll-jitter-percentage`; [provider-gcp](https://github.com/upbound/provider-gcp/blob/main/cmd/provider/compute/zz_main.go) hardcodes 5%

**Workarounds & Alternatives**

- Manually stagger resource creation times (impractical at scale)

**Additional context**

- Crossplane provides `WithPollJitterHook` in [`crossplane-runtime/pkg/reconciler/managed/reconciler.go`](https://github.com/crossplane/crossplane-runtime/blob/main/pkg/reconciler/managed/reconciler.go)
- Recommended jitter by landscape: Small (<100) — none, Medium (100–500) — ±10s, Large (500–2000) — ±30s, XL (2000+) — ±1m

**Documentation**
- add this to the technical [docs ](https://github.com/SAP/crossplane-provider-btp/blob/main/docs/end-user-guides/setup/configure-controller-flags.mdx)

**Follow-up** 
@mpechkurov  
- follow up if we need to transfer to other provirders
- will create outcomes ticket after implementation

[...]

PR #654, opened 2026-05-07
feature: expose more cli flags

enrico-kaack-comp:
Expose and wire up several CLI flags to be user configureable for fine tuning behaviour of the controller. Also includes documentation.

I tried to describe the parameters for the user side and from its technical side.

Note for reviewer:
Please mark everything that is not clearly explained to be refined.

closes also #645

### DIFF
--- a/cmd/provider/main.go
+++ b/cmd/provider/main.go
@@ -180,40 +180,60 @@ func main() {
 	kingpin.FatalIfError(err, "Cannot create controller manager")
 	kingpin.FatalIfError(apis.AddToScheme(mgr.GetScheme()), "Cannot add Template APIs to scheme")
 
-	setupTerraformControllers(mgr, log, maxReconcileRate, *pollInterval, *pollJitter, *reconcileTimeout, backoffBase, backoffMax, enableManagementPolicies, enableExternalSecretStores, namespace, terraformVersion, providerSource, providerVersion)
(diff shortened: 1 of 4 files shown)


sdischer-sap:
@gergely-szabo-sap would be great if you can review this one, since you worked on the related issue

sdischer-sap (approved):
LGTM, I will resolve those conflicts

gergely-szabo-sap (changes requested):
The new controller flag docs read as if all flags apply uniformly to every controller, but the implementation currently differs between native and Upjet-generated controllers.

In particular:

- `--poll-jitter` appears to apply to both native and Upjet controllers.
- `--max-concurrent-reconciles` only affects native controllers; generated Upjet controllers still use `MaxConcurrentReconciles: 1`.
- `--reconcile-timeout` only affects native controllers wired through `providerconfig.DefaultSetup`; generated Upjet controllers still hardcode `managed.WithTimeout(3 * time.Minute)`.

I don’t think this native-vs-Upjet distinction is documented anywhere clearly, so users may expect these flags to tune all managed resources. Could we update `configure-controller-flags.mdx` to explicitly document which flags apply to native controllers only vs both native and Upjet-generated controllers?

Otherwise the docs may be misleading for most Terraform-backed resources.

sdischer-sap:
@gergely-szabo-sap great catch. Missed that. Ok will leave it open for now then.

Issue #656, opened 2026-05-08
[FEATURE] Support `secretKey` option for non-flattened service binding secrets


daniel-kurzynski:
**Context**

SAP components like Connectivity Proxy and Transparent Proxy Manager expect service binding credentials as a single JSON blob under one key (e.g. `credentials`), rather than flattened into individual top-level keys. The SAP BTP Service Operator supports this via `spec.secretKey`. Currently, the crossplane provider always flattens credential JSON into individual keys, making it incompatible with these components.

**Scope**

Add a `secretKey` field to `ServiceBindingSpec`. When set:

- All credential properties are bundled into a single JSON key with the specified name instead of being flattened into individual top-level keys.
- When combined with `secretFormat: sap-kubernetes`, the `.metadata` descriptor marks this key with `"container": true` per the SAP Kubernetes Service Binding specification, and top-level metadata keys (`type`, `label`, `plan`, `instance_name`, `instance_guid`, `tags`) are still exposed alongside the credential key.
- When used without `secretFormat`, credentials are simply bundled under the key with no metadata enrichment.

**Out of scope**

- Changes to the existing flattening behavior when `secretKey` is not set.
- Automatic rotation behavior changes (rotation works transparently since the pipeline runs on every Observe).

**Technical Steps**

1. Add `SecretKey *string` field to `ServiceBindingSpec` in `apis/account/v1alpha1/servicebinding_types.go`
2. Implement `bundleCredentials` / `assembleCredentialJSON` in `secretformat.go` — when `secretKey` is set, skip flattening and store all credentials under the single key
3. Adapt `enrichConnectionDetails` to emit a single `credentialProperties` entry with `container: true` when `secretKey` is active
4. Wire up via `processConnectionDetails` helper used in both `Observe()` and `Create()`
5. Regenerate CRD manifest
6. Add unit tests and example YAML

**Workarounds & Alternatives**

Today, users must post-process the flattened secret (e.g. via an init container or external script) to re-assemble the JSON blob before mounting it for components that expect the bundled format.

**Additional context**

The SAP BTP Service Operator reference: `spec.secretKey` bundles credentials under one key and marks it with `container: true` in the `.metadata` descriptor.

sdischer-sap:
Merged PR

PR #657, opened 2026-05-08
feat: add secretKey option for non-flattened service binding secrets

daniel-kurzynski:
When secretKey is set (e.g., "credentials"), all credential properties are bundled into a single JSON key instead of being flattened. Combined with secretFormat "sap-kubernetes", the .metadata descriptor marks the key with "container: true" per the SAP specification.

This enables direct consumption by SAP components like Connectivity Proxy and Transparent Proxy that expect a single credentials JSON blob.


daniel-kurzynski:
fixes https://github.com/SAP/crossplane-provider-btp/issues/656

sdischer-sap (approved):
Good stuff, thanks for you contribution. 
LGTM!

PR #658, opened 2026-05-11
docs: unify CloudManagement name to cis-local

lennardrother:
Use "cis-local" as the name of the `CloudManagement` resource everywhere so the different end user guides fit together.


jn-av (approved):
good catch thanks!

PR #670, opened 2026-05-20
fix: propagate forProvider.name to external-name for SubaccountApiCredential

### LINKED ISSUE #664: [FEATURE] SubaccountApiCredential: forProvider.name is ignored and credentials break permanently after reimport
**Describe the bug**

`spec.forProvider.name` on `SubaccountApiCredential` is silently ignored. The credential is always created with the hardcoded default name regardless of what the user specifies in the manifest.

This causes three issues:

1. `forProvider.name` has no effect. The credential is always named `managed-subbaccount-api-credential` regardless of what the user sets.

2. After a composition reimport (K8s resources deleted and reapplied while BTP resources still exist), the `SubaccountApiCredential` permanently breaks. The resource adopts the existing BTP credential but cannot recover the `client_secret` (BTP only returns it once at creation time). The resource gets stuck in a `ReconcileError` state and never recovers. All downstream resources that depend on the connection secret (e.g. `RoleCollection`, `RoleCollectionAssignment`) also break.

3. Name collisions. Because every `SubaccountApiCredential` in a subaccount gets the same hardcoded default name, two compositions targeting the same subaccount will collide. The second creation adopts the first composition's credential, causing cross-contamination of secrets.

v1.9.0

**To Reproduce**

Scenario 1: `forProvider.name` ignored

1. Apply a `SubaccountApiCredential` with `spec.forProvider.name: "my-custom-name"`
2. Wait for the resource to become Ready
3. Check the credential name in BTP Cockpit (Security > API Credentials)
4. The credential is named `managed-subbaccount-api-credential`, not `my-custom-name`

Scenario 2: Permanent stuck state after reimport

1. Deploy a composition that includes a `SubaccountApiCredential`
2. Wait for it to become Ready and verify the connection secret contains `attribute.client_secret`
3. Delete the K8s resources (simulating a reimport) without deleting the BTP credential
4. Reapply the same manifests
5. The `SubaccountApiCredential` finds the existing BTP credential (Observe succeeds) but the connection secret is written without `client_secret`
6. On the next reconcile loop, the initializer detects the missing `client_secret` and returns an error
7. The resource is permanently stuck in `ReconcileError` with message: `cannot read client_secret from source, please delete external resource and re-create Crossplane resource`

**Expected behavior**

1. `spec.forProvider.name` should be used as the credential name (or prefix) when creating the credential in BTP.
2. If the provider detects that credentials are invalid (connection secret missing `client_secret`), it should be able to recover by creating a new credential rather than requiring manual intervention.
3. Multiple `SubaccountApiCredential` resources in the same subaccount should not collide.

**Additional context**

`SetIdentifierArgumentFn` in `config/btp_subaccount_api_credential/config.go` receives the external-name annotation value, not `forProvider.name`. On first creation, the annotation is empty (Upjet's `NameAsExternalName` initializer is disabled via `IdentifierFromProvider`), so the function always falls back to the hardcoded default:

```go
r.ExternalName.SetIdentifierArgumentFn = func(base map[string]any, name string) {
    if name == "" {
        base["name"] = "managed-subbaccount-api-credential"  // always hits this path
    } else {
        base["name"] = name
    }
}
```

After creation, `LateInitialize` copies the actual BTP name back into `forProvider.name`, making it appear like it worked. But the user's original value was never used.

How Upjet manages this resource:
- The external-name annotation is the lookup key for Observe (maps to credential `name` in BTP)
- `DisableNameInitializer: true` (from `IdentifierFromProvider`) means the annotation starts empty, not set to `metadata.name`
- The BTP Terraform provider does not implement `ImportState` for this resource. Pre-setting the annotation triggers Upjet's import path which always fails.
- `name` in the TF schema is `optional, computed` with no `ForceNew`. Terraform treats name changes as in-place updates (which BTP may reject).
- Sensitive fields (`client_secret`, `client_id`, `token_url`, `api_url`) are stored only in the connection secret, not in `status.atProvider`
- `client_secret` is a write-once value that BTP only returns at creation time. Once lost, the only recovery path is creating a new credential.

**Proposed fix options**

Both options share a constraint: once `client_secret` is lost, the old credential is unrecoverable. Recovery always means creating a new credential, leaving the old one orphaned in BTP (Upjet only calls Destroy when the K8s resource itself is deleted, not when the name changes). Orphan cleanup would need to be handled out-of-band.

*Option A: Caller detects and triggers recovery*

The provider adds a `NameSyncInitializer` that propagates `forProvider.name` to the external-name annotation when they differ. The existing `DeletionProtectionInitializer` continues to return an error when `client_secret` is missing. The caller (composition or user) is responsible for detecting the error condition and patching `forProvider.name` to a new value, which triggers the provider to create a fresh credential.

Implementation:
1. Provider adds a `NameSyncInitializer` that propagates `forProvider.name` to the annotation when they differ
2. `DeletionProtectionInitializer` continues to return `errMissingClientSecret` (as today)
3. Caller detects the `ReconcileError` condition
4. Caller patches `forProvider.name` to a new value (e.g. appends a timestamp)
5. Provider syncs new name to annotation, Observe finds nothing, Create runs, fresh credential issued

Pros:
- Provider stays simple, no autonomous behavior
- Caller has full control over when and how recovery happens
- No credentials created without explicit caller decision

Cons:
- Every caller that uses `SubaccountApiCredential` needs its own recovery logic
- Leaks implementation details (write-once secrets, name-as-identity) to callers
- Slower recovery since the caller must detect the error, then act on it
- Scales poorly. N callers means N copies of the same recovery logic.

*Option B: Provider self-heals autonomously (What I would prefer)*

The provider treats `forProvider.name` as a prefix and always appends a random 5-char suffix when creating credentials. When the initializer detects missing `client_secret` in the connection secret, it generates a new name with a fresh suffix, updates the annotation, and lets the normal create flow produce a new credential with a valid secret.

Implementation:
1. Provider adds a `NameSyncInitializer` that:
   - On every creation: appends a random 5-char suffix to `forProvider.name` (or to the default prefix `"managed-sac"` if not set)
   - On detecting missing `client_secret`: generates a new name with a fresh suffix, sets annotation, returns nil (no error)
2. `DeletionProtectionInitializer`'s missing-secret check is replaced by the self-healing logic in the new initializer
3. Upjet's normal flow continues: Observe finds nothing under the new name, Create runs, fresh `client_secret` returned

Naming behavior:
```
forProvider.name: "my-cred"   -> BTP name: "my-cred-x7k2m"
forProvider.name: not set     -> BTP name: "managed-sac-a3f9b"
```

Pros:
- Zero caller changes needed, fixes all existing and future users of this resource
- Eliminates the name collision problem (unique suffix per creation)
- Fast recovery within 1-2 reconcile loops
- Follows the same pattern as `ServiceBinding` which already uses `GenerateRandomName` for rotation

Cons:
- Provider takes autonomous action the caller didn't explicitly request
- BTP credential names become less predictable (have random suffixes)

jtucci:
Fixes #664

## What I found

`spec.forProvider.name` was silently ignored on `SubaccountApiCredential`. The `SetIdentifierArgumentFn` receives the external-name annotation value, not `forProvider.name`. On first creation the annotation is always empty (because `IdentifierFromProvider` disables Upjet's `NameAsExternalName` initializer), so the function always fell back to the hardcoded default `managed-subbaccount-api-credential` regardless of what the user set.

## What I changed

Added a `NameAsExternalName` initializer that runs before `DeletionProtectionInitializer` and propagates `spec.forProvider.name` to the external-name annotation whenever they differ. When the name changes (annotation != forProvider.name), the stale connection secret is deleted first so `DeletionProtectionInitializer` does not block the reconcile with incomplete credential data. The secret is recreated by Upjet after the new credential is provisioned.

## Limitations (Option A)

This implements Option A from the issue. Recovery from the reimport stuck-state (Scenario 2) requires the caller to change `forProvider.name` to a new value, which triggers the initializer to create a fresh credential. The old BTP credential is orphaned and must be cleaned up out-of-band. Option B (autonomous self-healing) remains a future consideration.

### DIFF
--- a/config/btp_subaccount_api_credential/config.go
+++ b/config/btp_subaccount_api_credential/config.go
@@ -22,6 +22,7 @@ const (
 	errTrackRUsage         = "cannot track ResourceUsage"
 	errTypeAssertion       = "managed resource is not of type SubaccountApiCredential"
 	errMissingClientSecret = "cannot read client_secret from source, please delete external resource and re-create Crossplane resource"
+	errUpdateExternalName  = "cannot update external-name annotation"
 )
 
 // Configure configures individual resources by adding custom ResourceConfigurators.
@@ -60,11 +61,65 @@ func Configure(p *config.Provider) {
 
 		// Add pre-delete hook using InitializerFns for finalizer management
 		r.InitializerFns = append(r.InitializerFns, func(kube client.Client) managed.Initializer {
+			return &NameAsExternalName{Kube: kube}
+		})
+		r.InitializerFns = append(r.InitializerFns, func(kube client.Client) managed.Initializer {
 			return &DeletionProtectionInitializer{Kube: kube}
 		})
 	})
 
 	p.ConfigureResources()
+}
+
+// NameAsExternalName propagates forProvider.name to the external-name annotation.
+// When the name changes, the connection secret is deleted so that
+// DeletionProtectionInitializer does not block with stale data.
+type NameAsExternalName struct {
+	Kube client.Client
+}
+
+func (n *NameAsExternalName) Initialize(ctx context.Context, mg resource.Managed) error {
+	cr, ok := mg.(*securityv1alpha1.SubaccountApiCredential)
+	if !ok {
+		return errors.New(errTypeAssertion)
+	}
+
+	if meta.WasDeleted(mg) {
+		return nil
+	}
+
+	if cr.Spec.ForProvider.Name == nil || *cr.Spec.ForProvider.Name == "" {
+		return nil
+	}
+
+	desiredName := *cr.Spec.ForProvider.Name
+	if meta.GetExternalName(cr) == desiredName {
+		return nil
+	}
+
+	if err := n.deleteConnectionSecret(ctx, cr); err != nil {
+		return err
+	}
+
+	meta.SetExternalName(cr, desiredName)
+	return errors.Wrap(n.Kube.Update(ctx, cr), errUpdateExternalName)
+}
+
+func (n *NameAsExternalName) deleteConnectionSecret(ctx context.Context, cr *securityv1alpha1.SubaccountApiCredential) error {
+	secretRef := cr.GetWriteConnectionSecretToReference()
+	if secretRef == nil || secretRef.Name == "" {
+		return nil
+	}
+
+	secret := &corev1.Secret{}
+	if err := n.Kube.Get(ctx, client.ObjectKey{
+		Name:      secretRef.Name,
+		Namespace: secretRef.Namespace,
+	}, secret); err != nil {
+		return resource.IgnoreNotFound(err)
+	}
+
+	return resource.IgnoreNotFound(n.Kube.Delete(ctx, secret))
 }
 
 // DeletionProtectionInitializer implements the managed.Initializer interface
--- a/config/btp_subaccount_api_credential/config_test.go
+++ b/config/btp_subaccount_api_credential/config_test.go
@@ -5,14 +5,186 @@ import (
 	"testing"
 
 	xpv1 "github.com/crossplane/crossplane-runtime/v2/apis/common/v1"
+	"github.com/crossplane/crossplane-runtime/v2/pkg/meta"
 	"github.com/crossplane/crossplane-runtime/v2/pkg/test"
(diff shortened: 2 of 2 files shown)


sdischer-sap (approved):
lgtm.

PR #671, opened 2026-05-21
feat: add external-name ADR compliance for SubaccountTrustConfiguration

### LINKED ISSUE #547: [TASK] `External-Name`(SubaccountTrustConfiguration): Ensure ADR compliance
**Part of Epic: #427** 
## Context

## Technical Tasks Checklist

**Acceptance Criteria:**

- [ ] **Migration Logic**:

**Acceptance Criteria:**
- [ ] **Observe() Tests**:

- [ ] **Create() Tests**:

- [ ] **Update() Tests**:

- [ ] **Delete() Tests**:

**Acceptance Criteria:**
- [ ] **Import Flow Test**:

jn-av:
Implements ADR-compliant external-name handling for `SubaccountTrustConfiguration` as part of Epic #427.

The compound key `<subaccount-id>/<origin>` (e.g. `abc-123/sap.custom`) is used as the external-name identifier. The `origin` portion is `Computed`-only in the Terraform schema — it is never part of `spec.forProvider` and only returned by the API after Create/Import.

This creates a challenge for Upjet's refresh path (used when management policies include `Create`/`Update`/`*`): the TF state has no `origin` on first reconcile, causing `Read(origin="")` → API 404 → spurious Create attempt.

The `OriginInitializer` addresses this by seeding `status.atProvider.origin` from the `origin` portion of the compound external-name annotation before the first reconcile. Upjet's `EnsureTFState` writes observation fields into `terraform.tfstate` but not into `main.tf.json`, so `origin` lands in the state without triggering a "read-only attribute" plan error.

`GetExternalNameFn` reconstructs the compound key `subaccount_id/origin` from the TF state after every Create/Observe, preventing Upjet from overwriting the annotation with just the `origin` value. `GetIDFn` passes the external-name through unchanged since it is already in the correct TF import ID format.

## Changes

- `config/subaccount_trust_configuration/config.go` — `GetExternalNameFn` (compound key reconstruction) + `GetIDFn` (passthrough) + `OriginInitializer`
- `internal/controller/security/subaccounttrustconfiguration/zz_controller.go` — wire `InitializerFns` loop (regenerated via `make generate`)
- `apis/security/v1alpha1/subaccounttrustconfiguration_types.go` — external-name documentation comment
- `docs/end-user-guides/import-landscape/external-name.md` — updated via `make generate`
- `test/e2e/` — import flow e2e test
- `test/upgrade/` — upgrade test for external-name behavior

## Import example

External-name = compound key `<subaccount-id>/<origin>`. `subaccountRef` is NOT required for the import flow — it is populated automatically after Create/Import:

```yaml
metadata:
  annotations:
    crossplane.io/external-name: abc-123-def-456/sap.custom
  name: meta-trustconfiguration-observe
spec:
  forProvider: {}
  providerConfigRef:
    name: account-provider-config
```

## Decision: compound key `<subaccount-id>/<origin>`

After discussion, we chose the compound key format `<subaccount-id>/<origin>` (e.g. `abc-123/sap.custom`) as the external-name. This aligns with the ADR principle that the external-name should be fully self-contained — no `subaccountRef` required for the import/adoption flow.

`subaccountRef`/`subaccountSelector`/`subaccountId` in `spec.forProvider` remain supported and are populated automatically after Create/Import by the normal reconciliation loop.

**Trade-offs:**

| | `origin` only | compound key ✅ |
|---|---|---|
| External-name | `sap.custom` | `abc-123/sap.custom` |
| `subaccountRef` required for import | yes | no |
| Consistent with ADR compound key format | — | ✓ |
| User friendliness | simpler | self-contained |

Closes #547


enrico-kaack-comp:
> The compound key would be more aligned with the ADR.

I would take the decision based on this. The idea that the external-name contains everything that is required to address the external resource is the consistency throughout all resources.

enrico-kaack-comp (dismissed):
nice work, well documented thank you. As commented above, I would sugggest to stay closer to the ADR and use the compound key. So you probably need to adapt a few things here

sdischer-sap (commented):
@jn-av Does this satisfy this Observe-Only Import now as well? 

We were a big back and forth here. I was a bit persistant with changing the external-name in case of upjetted resources, but if that works and is upgrade tested I guess thats fine. There are two points that concern me a little though: 
- now we have different compound key formats, in all other cases we have `/` as a delimiter
- in `SubaccountApiCredentials` we didn't do those adaptations, I guess in order to have a unified approach we would need it there as well

jn-av:
@sdischer-sap You're absolutely right about the separator! My bad, I had the Terraform format in mind. I'll change it to `/`.

jn-av:
> @jn-av Does this satisfy the Observe-Only Import now as well?

Thanks for the reminder! Just tested it, it works.

> In `SubaccountApiCredentials` we didn't do those adaptations. I guess, in order to have a unified approach, we would need it there as well.

That's right, I just checked. I'd also be in favor of using a compound key there. In our ADR, we described it as the expected format, so wherever possible, we should enforce it:

https://github.com/SAP/crossplane-provider-btp/blob/7df0c8a3000500c58937c960ec520de3d5507e8d/docs/contribution-notes/external-name-handling.md?plain=1#L38-L43

I've created a follow-up issue for this, let's move our discussion there to decide whether we want to do it or not: https://github.com/SAP/crossplane-provider-btp/issues/674 @sdischer-sap

enrico-kaack-comp (approved):
Thanks at @sdischer-sap for pointing the delimter out, totally missed that. From my side it looks good.

sdischer-sap (approved):
Nice, lgtm now. 
Thank you for the extra effort!

PR #673, opened 2026-05-27
fix: add recreateOnCreationFailure for KymaEnvironment

### LINKED ISSUE #672: Bug: KymaEnvironment stuck in CREATION_FAILED state with no auto-recovery
**Describe the bug**

When a KymaEnvironment resource enters `CREATION_FAILED` state (e.g., due to a missing entitlement at creation time), the resource never recovers, even after the underlying issue is resolved. The `Observe()` method does not explicitly handle `CREATION_FAILED` and falls into the default branch, which marks the resource `Unavailable()` but returns `ResourceExists: true, ResourceUpToDate: true`. The reconciler never retries creation.

After doing some digging I found that the Subscription resource already handles an equivalent failure pattern via `shouldRecreateOnFailure()` with an opt-in spec field (`recreateOnSubscriptionFailure`). I was thinking we could use that same pattern here for KymaEnvironment 

v1.9.0

**To Reproduce**

1. Create a KymaEnvironment resource before its required entitlement is provisioned
2. BTP accepts the API call but provisioning fails
3. Environment transitions to `CREATION_FAILED` state
4. Provision the entitlement (the prerequisite is now satisfied)
5. Observe that the KymaEnvironment remains stuck in `CREATION_FAILED` indefinitely
6. Only manual deletion and recreation recovers the resource

**Expected behavior**

The provider should offer a recovery path for `CREATION_FAILED` state, similar to how Subscription handles `SUBSCRIBE_FAILED`. Specifically:

- `Observe()` should explicitly branch on `CREATION_FAILED`
- An opt-in spec field (e.g., `recreateOnCreationFailure`) should control whether the provider automatically deletes the failed environment so the next reconcile re-creates it
- Default behavior (`false`) preserves backwards compatibility

**Additional context**

- The `CREATION_FAILED` state is already documented in the API enum (`[CREATING UPDATING DELETING OK CREATION_FAILED DELETION_FAILED UPDATE_FAILED]`) but has no corresponding Go constant or handling logic
- The Subscription controller's `shouldRecreateOnFailure()` pattern might be a good reference for the fix
- CloudFoundryEnvironment likely has the same gap (its `Observe()` only checks for `OK`) but that can be a separate issue

**Open Question: Retry behavior when underlying issue persists**

The recreation loop will repeat if the root cause (e.g., missing entitlement) isn't resolved:

```
Create → CREATION_FAILED → Delete → Create → CREATION_FAILED → Delete → ...
```

Crossplane's managed reconciler has a built-in poll interval (default ~1 min) but no exponential backoff on the happy path since we return nil errors. This could produce steady API churn against BTP.

**Some potential options:**

1. **No limit (match Subscription)** — Subscription's `recreateOnSubscriptionFailure` has the same unbounded loop behavior today. Keeps the implementation simple and consistent across resources.

2. **Max retry counter** — Track recreation attempts (similar to the existing update circuit breaker). After N failed recreations, stop retrying and surface an error condition. User would need to manually reset or fix the issue.

3. **Return error to trigger backoff** — Return an error from `Observe()` after the Delete completes so the managed reconciler applies its exponential backoff (caps at ~1hr). Simpler than a counter, naturally rate-limits without a hard stop.

jtucci:
Fixes #672

When a KymaEnvironment enters `CREATION_FAILED` state, the reconciler gets stuck. This adds an opt-in `recreateOnCreationFailure` spec field that triggers automatic deletion so the next reconcile re-creates the environment. Follows the same pattern as Subscription's `recreateOnSubscriptionFailure`.

## What I changed

- Added `CREATION_FAILED`, `DELETION_FAILED`, `UPDATE_FAILED` state constants
- Added `recreateOnCreationFailure` field to `KymaEnvironmentSpec` (defaults to `false`)
- Added `shouldRecreateOnFailure()` check in `Observe()` that triggers `Delete()` when enabled
- Added `MockTracker` to the kyma fake package for test support
- Regenerated CRD

### DIFF
--- a/internal/controller/environment/kyma/fake/fake.go
+++ b/internal/controller/environment/kyma/fake/fake.go
@@ -4,18 +4,12 @@ import (
 	"context"
 	"net/http"
 
-	"github.com/crossplane/crossplane-runtime/v2/pkg/resource"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
 	"github.com/sap/crossplane-provider-btp/apis/environment/v1alpha1"
-	apisv1alpha1 "github.com/sap/crossplane-provider-btp/apis/v1alpha1"
 	environments "github.com/sap/crossplane-provider-btp/internal/clients/kymaenvironment"
 	provisioningclient "github.com/sap/crossplane-provider-btp/internal/openapi_clients/btp-provisioning-service-api-go/pkg"
-	"github.com/sap/crossplane-provider-btp/internal/tracking"
 )
 
 var _ environments.Client = &MockClient{}
-var _ tracking.ReferenceResolverTracker = &MockTracker{}
 
 type MockClient struct {
 	MockDescribeCluster func(ctx context.Context, input *v1alpha1.KymaEnvironment) (*provisioningclient.BusinessEnvironmentInstanceResponseObject, error)
@@ -39,31 +33,5 @@ func (c MockClient) DeleteInstance(ctx context.Context, cr v1alpha1.KymaEnvironm
 	if c.MockDeleteCluster != nil {
 		return c.MockDeleteCluster(ctx, &cr)
 	}
-	return nil, nil
-}
-
-type MockTracker struct {
-	MockDeleteShouldBeBlocked func(mg resource.Managed) bool
-}
-
-func (m *MockTracker) Track(ctx context.Context, mg resource.Managed) error {
-	return nil
-}
-
-func (m *MockTracker) SetConditions(ctx context.Context, mg resource.Managed) {
-}
-
-func (m *MockTracker) DeleteShouldBeBlocked(mg resource.Managed) bool {
-	if m.MockDeleteShouldBeBlocked != nil {
-		return m.MockDeleteShouldBeBlocked(mg)
-	}
-	return false
-}
-
-func (m *MockTracker) ResolveSource(ctx context.Context, ru apisv1alpha1.ResourceUsage) (*metav1.PartialObjectMetadata, error) {
-	return nil, nil
-}
-
-func (m *MockTracker) ResolveTarget(ctx context.Context, ru apisv1alpha1.ResourceUsage) (*metav1.PartialObjectMetadata, error) {
 	return nil, nil
 }
--- a/internal/controller/environment/kyma/fake/fake.go
+++ b/internal/controller/environment/kyma/fake/fake.go
@@ -4,13 +4,18 @@ import (
 	"context"
 	"net/http"
 
-	environments "github.com/sap/crossplane-provider-btp/internal/clients/kymaenvironment"
-	provisioningclient "github.com/sap/crossplane-provider-btp/internal/openapi_clients/btp-provisioning-service-api-go/pkg"
+	"github.com/crossplane/crossplane-runtime/v2/pkg/resource"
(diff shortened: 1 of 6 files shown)


internal/controller/environment/kyma/fake/fake.go
```diff
@@ -36,3 +41,29 @@ func (c MockClient) DeleteInstance(ctx context.Context, cr v1alpha1.KymaEnvironm
 	}
 	return nil, nil
 }
+
+type MockTracker struct {
```
sdischer-sap: We have gathered a lot of duplicated mocks for this Tracker stuff over time. 
Maybe I miss something, but I think this should be what you need:
https://github.com/SAP/crossplane-provider-btp/blob/main/internal/testutils/tracker_mock.go#L53

Would be great to reuse it at least for this case instead of introducing another one. We need to align the other duplicated ones at some later point as well.

internal/controller/environment/kyma/kymaenvironment.go
```diff
@@ -128,6 +150,18 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 	lastModified := cr.Status.AtProvider.ModifiedDate
 	cr.Status.AtProvider = kymaenv.GenerateObservation(instance)
 
+	if c.shouldRecreateOnFailure(cr, cr.Status.AtProvider.State) {
+		var err error
+		if !environmentBeingDeleted(cr) {
+			_, err = c.Delete(ctx, mg)
+		}
+		return managed.ExternalObservation{
+			ResourceExists:    true,
```
sdischer-sap, line 159: I see we have the same in subscription, but why do we return true here? 
Shouldn't we immediately try to Recreate?

sdischer-sap (changes requested):

internal/controller/environment/kyma/fake/fake.go
jtucci: Will do!

internal/controller/environment/kyma/kymaenvironment.go
jtucci: @sdischer-sap  When I implemented this I tried to match the same pattern Subscription uses in shouldRecreateOnFailure. My assumption on why it was done that way was that returning true here gives the resource time to actually get deleted / avoid a race before we try to recreate. if we returned ResourceExists: false, the reconciler would immediately call Create() while the old failed instance is still being deleted on BTP's side.
  
But I might be wrong about that, do you think it's safe to return false here and let the reconciler handle it, or would that risk a duplicate/conflict on the BTP API?

internal/controller/environment/kyma/fake/fake.go
jtucci: done

internal/controller/environment/kyma/kymaenvironment.go
sdischer-sap: Yeah thats a good point on API delays. Kyma is known to be the slowest in terms of processing (creation can likely take 20mins). I guess we need to set some status condition to Unavailable() in that cause, which should be done already above in this overall state mapping. 

So yes in that case It probably makes sense. `ResourceExists` doesn't necessarily mean healthy, just not immediatly trying creation. Thats good to go for me then.

sdischer-sap (approved):
Thanks as usual for the great work. LGTM now

PR #678, opened 2026-05-28
docs: add info box pointing to releases page for provider version

sk31337:
## Summary

Adds a new `:::note` admonition to the install guide (`docs/end-user-guides/setup/install-provider-btp.mdx`) telling readers where to find a valid value for `<version>` in the self-hosted `Provider` manifest.

The new box sits next to the existing "Not using a managed control plane?" note, in the same style:

> **Which version should I use?**
> Replace `<version>` with your desired version from [the releases page](https://github.com/SAP/crossplane-provider-btp/releases).

## Why

The YAML snippet uses `<version>` as a placeholder, but the docs didn't link to where supported versions are published. New users had to guess or search for the releases page.

## Test plan

- [ ] Docs site renders the new admonition correctly in the "On a self-hosted control plane" tab
- [ ] Link to the releases page resolves


jn-av (approved):
Thanks for the addition, looks good to me!

sk31337:
@jn-av please add [label](https://github.com/SAP/crossplane-provider-btp/actions/runs/26574712444/job/78290773333?pr=678).

PR #679, opened 2026-05-28
fix: e2e-long tests for KymaModule

enrico-kaack-comp:
Fix a bug introduced in #640 (not released) that broke KymaModule. It removed the External Name from the KymaEnvironmentBinding.

The KymaModule has a selector with an extractor of the KymaEnvironmentBinding from its ID (that was written in the external name). Since we removed it, the KymaModule reference to the KymaEnvironmentBinding could not been resolved so it was seen as not ready (although the secret was there). Consequently it was never ready.

## The fix
 The KymaEnvironmentBindingId was never used and served no purpose --> removed the resolver and extractor. The secret reference is extracted into other fields.

Long E2E since the short skip Kyma (which also explains how it got into main): https://github.com/SAP/crossplane-provider-btp/actions/runs/26865910256


sdischer-sap (dismissed):

sdischer-sap (dismissed):
lgtm

maximilianbraun:
Long started https://github.com/SAP/crossplane-provider-btp/actions/runs/28283177680

xdj2e9:
Hey @enrico-kaack-comp,

is the timeout bump still relevant? Almost all recent runs seem to be green in https://github.com/SAP/crossplane-provider-btp/actions/workflows/cron_long_e2e_test.yaml.

After your confirmation, I would be happy to merge 🙂

enrico-kaack-comp:
If the current timeout lead to stable e2e, this PR is not need. Thank you for following up :)

Issue #207, opened 2025-06-26
[EPIC] Upjet no-fork migration


sdischer-sap:
## What is no-fork?
Upjet has introduced a so called no-fork controller implementation. Essentially it is a alternative implementation of the usual `external` controller implementation. The difference is that no-fork is working based on an instance of the underlying provider directly and initializes the flow based on go method calls, rather then invoking a terraform process in the OS of the container. 

## Why we want to use it?
**The biggest advantage of that approach is a huge reduction in terms of resource footprint.** 
Generally this solution **feels more lightweight, stable and controllable**, since with it there are no more interactions leaving the go code itself.

Also: Terraform License change of the executable in versions higher than 1.5.7 into Business Source License (https://www.hashicorp.com/en/license-faq). The provider sdks etc. from Hashicorp remain Mozilla Public License (https://www.hashicorp.com/license-faq#impact-on-terraform-providers), so we would not use the terraform binary anymore.

## Update after initial approach
We introduced upjet in most of the major resources and thought it was fine. After merging we realized that there were actually some issues causing update drifts (see https://github.com/crossplane/upjet/issues/547). We reverted our changes in main and stoped this efforts for some time now. 
Meanwhile the upjet folks provided a [fix](https://github.com/crossplane/upjet/pull/555/changes/680c15d41934f3156ad50e7d6054ab129235e6d7), which is only available in v2 and not in any official release yet (current last version of upjet: v2.2). Doing some small POCs showed that this issue seems to be fixed in general. There are some small adaptations that seem to be required though.

## POC
You can relate to this **WIP POC PR**: https://github.com/SAP/crossplane-provider-btp/pull/680
It should mostly adapt no-fork already in all resources and apply some implementation fixes. Mixing pieces are:
- chores like fixing make reviewable and fixing some test pluming
- fixing the broken ServiceBinding rotation 
- more extensive (upgrade) testing.

While looking into this a lot of dependencies unraveled, which is why in this PR a lot of things have been merged together (like updating terraform provider, go version, etc.). This isn't ideal and should be introduced in smaller iterations if possible. Not sure if its possible though.

## Implementation plan
**Please note that this epic contains completed issues from our previous attempt of implementation.** Those can be ignored and are only meant for reference.

**Here is a rough outline of the requirements and Why we need them:**
- pin latest upjet 
  - to get access to no-fork fix
- update to latest terraform provider 
  - newly added terraform identities are required for no-fork to work properly
[...]

PR #680, opened 2026-05-28
Feat/no fork xp2

sdischer-sap:


sdischer-sap:
This is just meant for reference as a POC for proper implementation in #207, so I will close this

PR #681, opened 2026-05-29
chore: migrate secrets to environment-scoped access

jn-av:
## Summary

- Add `environment:` to workflow jobs that use BTP secrets directly but had no environment set (`cron_long_e2e_test`, `upgrade-test`, `custom-upgrade-test`)
- Remove unused `get-project-column-ids.yml` workflow (never ran, used orphaned `GH_READ_TOKEN`)

## Context

Secrets are being migrated from repo-scope to environment-scope. Jobs must declare an environment to access environment-scoped secrets.

| Workflow | Environment | Changed? | Reason |
|---|---|---|---|
| `pr-e2e.yaml` | `pr-e2e-approval` | No | Triggered by external PRs — manual approval prevents secret exfiltration |
| `upgrade-test.yaml` | `pr-e2e-approval` | Yes | Manually triggerable by anyone with write access |
| `custom-upgrade-test.yaml` | `pr-e2e-approval` | Yes | Manually triggerable by anyone with write access |
| `cron_long_e2e_test.yaml` | `pr-e2e-no-approval` | Yes | Scheduled cron job — approval gate would block automatic execution |
| `release.yaml` | `pr-e2e-no-approval` | No | Release workflow is already gated by manual dispatch |
| `e2e_test.yaml` | *(from caller)* | No | Reusable workflow — inherits environment from calling workflow |


.github/workflows/custom-upgrade-test.yaml
```diff
@@ -54,6 +54,7 @@ jobs:
   custom-upgrade-test:
     runs-on: ubuntu-latest
     needs: prepare-matrix
+    environment: pr-e2e-approval
```
enrico-kaack-comp, line 57: previously, we didnt had an environment here because it was only workflow_dispatch. so now we would manually trigger and than have to approve again.

Thats fine i think, playing it safer in case we later modify the `on:` triggers to not accidentally forget it.
Same for upgrade_test workflow that is `workflow_dispatch` only

.github/workflows/custom-upgrade-test.yaml
```diff
@@ -114,6 +115,7 @@ jobs:
   cleanup-custom-upgrade-test:
     runs-on: ubuntu-latest
     needs: [prepare-matrix, custom-upgrade-test]
+    environment: pr-e2e-approval
```
enrico-kaack-comp: check the e2e file where this is from:
```
    needs: [prepare-matrix, e2e-test]
    if: always() && needs.prepare-matrix.result == 'success'  # only run if prepare-matrix was successful (requires manual approval) and then always run after e2e-test
```
if the prepare-matrix is rejected, the cleanup function is not called but skipped, see example https://github.com/SAP/crossplane-provider-btp/actions/runs/26576404923

enrico-kaack-comp (changes requested):

sdischer-sap (approved):
lgtm

enrico-kaack-comp (approved):
LGTM

PR #687, opened 2026-06-03
feat: add dataCenter and servicePlanID fields to ServiceInstance

### LINKED ISSUE #446: [FEATURE]  Restore Service Plan ID Field for Service Instances
**Context**

When provisioning SAP BTP service instances using Crossplane Provider BTP, users sometimes encounter scenarios where service offering and plan names are ambiguous (e.g., same service offering available in multiple data centers or regions with identical plan names). In such cases, specifying the exact service plan ID (UUID) is necessary to avoid provisioning errors.

The BTP Service Operator provides a `servicePlanID` field that allows users to specify the plan ID directly when service offering and plan name resolution is ambiguous.

According to internal knowledge, this functionality was once implemented in Crossplane Provider BTP but is currently not accessible through the CRD API.

**Scope**

1. **Suggestion on how to Restore API Access**: Make the service plan ID field accessible in the ServiceInstance CRD
   - Add `.spec.forProvider.servicePlanID` field (or similar)
   - Allow users to specify plan ID as alternative to plan name
   - Support plan ID as optional field (plan name remains the default approach)

2. **Documentation**: Document when and how to use service plan ID
   - Explain scenarios where plan ID is needed
   - Provide examples of plan ID usage
   - Document how to retrieve plan IDs from BTP

3. **Suggestion on how to validate**: Add appropriate validation logic
   - Allow either `servicePlanName` OR `servicePlanID` (not both)
   - (if possible) Validate against BTP backend
   - Validate plan ID format (UUID)
   - Provide clear error messages when both are specified

**Out of scope**

**Technical Steps**

**Possible (untested) Workarounds & Alternatives**

1. Create entitlements with [.spec.forProvider.servicePlanUniqueIdentifier](https://doc.crds.dev/github.com/SAP/crossplane-provider-btp/account.btp.sap.crossplane.io/Entitlement/v1alpha1@v1.5.0#spec-forProvider-servicePlanUniqueIdentifier) to ensure correct data center/region selection
2. Use BTP Service Operator as alternative for plan ID scenarios

**Additional context**

- **BTP Service Operator Reference**: [servicePlanID field](https://github.com/SAP/sap-btp-service-operator?tab=readme-ov-file#spec)
- **Crossplane Provider BTP API**: [ServiceInstance Spec](https://doc.crds.dev/github.com/SAP/crossplane-provider-btp/account.btp.sap.crossplane.io/ServiceInstance/v1alpha1@v1.5.0#spec-forProvider)
- **BTP Service Plan IDs**: Can be retrieved via [BTP CLI](https://help.sap.com/docs/btp/btp-cli-command-reference/btp-list-services-plan): `btp list services/plans --subaccount <id>`

**Example Use Case**:
```yaml
kind: ServiceInstance
metadata:
  name: my-xsuaa-service
spec:
  forProvider:
    offeringName: xsuaa
    servicePlanID: "a1b2c3d4-e5f6-7890-abcd-....."  # Direct plan ID
    # planName: "application"  # Alternative: use plan name
    subaccountRef:
      name: my-subaccount
```

**Priority**: Medium - While plan names work in most cases, plan ID support is essential for certain edge cases and multi-region deployments which have not been requested yet by `ManagedControlPlane` users.

### LINKED ISSUE #686: [FEATURE]  Add dataCenter field to ServiceInstance resource for correct service plan resolution
**Context**
When provisioning service instances (e.g., HANA Cloud) via crossplane-provider-btp, the provider resolves the servicePlanID by querying the Service Manager API using only offeringName and planName. However, service offerings like hana-cloud exist across multiple data centers, each with a different servicePlanID for the same plan name (e.g., hana).

Without a dataCenter filter, the provider resolves to a plan ID from a different data center than the one entitled in the subaccount. This causes the ServiceInstance to fail silently or get stuck — the resolved plan ID (e.g., abc123) doesn't match the entitled plan ID (e.g., 6220e8a0-c377-4ec0-b5ba-ea9695f78edc), and the BTP Service Manager rejects the provisioning request because the subaccount only has an entitlement for the plan in its specific data center.

**Scope**
Add an optional dataCenter field to spec.forProvider of the ServiceInstance CRD:

When dataCenter is specified, include it in the Service Manager query when resolving the plan ID
This ensures the provider resolves the plan ID that matches the entitlement in the subaccount's specific data center.
The field should be optional and backward-compatible , when omitted, behavior remains unchanged (no data center filtering, same as today).

**Out of scope**
If needed, describe what parts of a feature can be ignored for now.

_Hint_: Smaller feature wishes tend to be implemented and shipped quicker!

**Technical Steps** 
(Optional) - outline steps for contributors to give implementation guidance and allow transparent technical discussions.

**Workarounds & Alternatives**
Describe, if possible, how you are solving this problem today.

**Additional context**

jtucci:
Adds two optional fields to ServiceInstance for resolving plan ambiguity in multi-datacenter offerings like HANA Cloud:

- `dataCenter` — filters Service Manager offerings by `catalog_name + data_center` when resolving offeringName/planName to a plan ID. Matches the pattern used by BTP Service Operator.
- `servicePlanID` — bypasses name-based resolution entirely. Use when you already have the plan UUID or when the desired offering has an empty data_center value.

Background: Previously, the plan resolver queried offerings only by catalog_name. For services like hana cloud that exist in multiple datacenters, each datacenter has its own offering entry and the default one has an empty data_center value. Without filtering, the API returns all of them and the code picks whichever comes first, which may not be the default. This breaks if a new datacenter gets added to a service offering since existing instances could start resolving to the wrong plan ID.

Both fields are optional. The query includes data_center eq '<value>' in the fieldQuery. When dataCenter is set, it filters for that specific datacenter. When dataCenter is not set, it defaults to data_center eq '' which selects the default (empty) datacenter entry. For multi-datacenter services like hana cloud, the default datacenter has an empty data_center value, so this correctly selects it instead of picking randomly. For services that dont have a data_center field at all (xsuaa, destination, credstore, etc.), data_center eq '' still matches correctly. 

Also updates the feature comparison table (shared, secretKey, dataCenter, servicePlanID rows).

Closes #446
Closes #686


sdischer-sap:
Hey @jtucci, 

thanks again for your contribution. Looks great. I am holding off with the approval, so we can get https://github.com/SAP/crossplane-provider-btp/pull/699 in, which is also touching the ServiceInstance. 
Will approve once we have that merged and check if there might be some conflicts.

internal/clients/servicemanager/api_plan_resolver.go
```diff
@@ -119,14 +119,18 @@ func NewServiceManagerClient(ctx context.Context, creds *BindingCredentials) (*S
 	}, nil
 }
 
-func (sm *ServiceManagerClient) PlanIDByName(ctx context.Context, offeringName, planName string) (string, error) {
-	offeringQuery := fmt.Sprintf("catalog_name eq '%s'", offeringName)
+func (sm *ServiceManagerClient) PlanIDByName(ctx context.Context, offeringName, planName, dataCenter string) (string, error) {
+	offeringQuery := fmt.Sprintf("catalog_name eq '%s' and data_center eq '%s'", offeringName, dataCenter)
+
```
n-losse, line 124: what happens when the dataCenter is not specified in the CRD? looks like the filter query is then "eq emptString" - would that still work? 

also, do all offerings have that field? what is the behavior if they do not? I could not find it in the generic schema definition here https://api.sap.com/api/APIServiceManager/schema - am i looking in the wrong place?

ankitamohanty7: yes datacenter = "" will work, it will get the default one, unlike if datacenter is null, it will get all and just select random plan out of all
the plan in default datacenter has datacenter as "", so that should be default behavoiur uf datacenter is omitted from parameters, it should select the default datacenter

pls check operator code for parity https://github.com/SAP/sap-btp-service-operator/blob/355b1c33ac898af089a83041b300ff5b45d56b25/client/sm/client.go#L528

ankitamohanty7: data_center is not present in all offerings, and the one where it is not present query - data_center eq '' or having no data_center, will take care of them
But we also would have option to pass data_center for service offerings that have data_center field
The problem arises when a new data_center gets added to a service offering , the existing query fetches everything and gives random results and tries to land in a datacenter for which we have no entitlement selected

n-losse: I see it's identical in the operator code, but since it's nowhere in the service manager docs, it looks like relying on undocumented behavior (both for the existence of the field and the empty string filter). One could argue that operator had this for 3 years already, so its stable - but we're still lacking any guarantees, right? 

Could we not add the data_center to the query if it is empty string? That seems a bit safer.

ankitamohanty7: No, we need data_center as blank so that default is always picked when we are not specifying the data_center, that's what the bug is all about. 

The difference is
when we ask for records with data_center eq "" , it means the offering has a data_center field which is blank
Without a query field data_center , it means return all the records. When query returns all records, the code picks the first random one from the list which might have a data_center that we didn't ask for
image-5

jtucci: @n-losse To piggy-back off @ankitamohanty7 response, I tested this against the raw SM API (`GET /v1/service_offerings?fieldQuery=...`)  for services both with and without the `data_center` field. 


**Multi-datacenter service (hana-cloud), no `data_center` filter (The bug)**
```json
// GET /v1/service_offerings?fieldQuery=catalog_name eq 'hana-cloud'
// Returns botjh entries code picks items[0] which is non-deterministic
{
  "num_items": 2,
  "items": [
    {"id": "bbb4ce19-...", "catalog_name": "hana-cloud", "data_center": "sap_na-us-2"}
    {"id": "d0ec53c9-...", "catalog_name": "hana-cloud", "data_center": null},
  ]
}
```


**Multi-datacenter service, `data_center eq ''` (The fix which selects default data center)**

```json
// GET /v1/service_offerings?fieldQuery=catalog_name eq 'hana-cloud' and data_center eq ''
{
  "num_items": 1,
  "items": [
    {"id": "d0ec53c9-...", "catalog_name": "hana-cloud", "data_center": null}
  ]
}
```

**Multi-datacenter service, specific datacenter**

```json
// GET /v1/service_offerings?fieldQuery=catalog_name eq 'hana-cloud' and data_center eq 'sap_na-us-2'
{
  "num_items": 1,
  "items": [
    {"id": "bbb4ce19-...", "catalog_name": "hana-cloud", "data_center": "sap_na-us-2"}
  ]
}
```

**Service without a  `data_center` field (xsuaa), `data_center eq ''` Still safe** 
I actually ran this against 25+ other instances that dont have data center fields and they all worked. 
```json
// GET /v1/service_offerings?fieldQuery=catalog_name eq 'xsuaa' and data_center eq ''
// data_center is null in the response but the filter still matches
{
  "num_items": 1,
  "items": [
    {"id": "9563a892-...", "catalog_name": "xsuaa", "data_center": null}
  ]
}
```

As @ankitamohanty7 mentioned, without the `data_center eq ''` filter, you simply cant reliably provision resources where multiple datacenters exist and you want the default empty one, the API returns all of them and the code picks whichever comes first, which is non-deterministic.

The only alternative would be moving the filtering logic into the provider... fetch all, then filter client-side for `data_center == null`, but that still relies on the same undocumented field existing in the response body and it just moves the dependency from query-time to response-parsing-time, while adding more code. Also since data center isnt in the official SM OpenAPI spec or our generated Go client model wed have to either patch the spec or do raw json parsing to filter client side. The query approach avoids that since fieldQuery is just a pass-through string.

Since the btp service operator has relied on this exact query for 3+ years, the tests above show that services that dont have data center still work, and client side filtering doesnt really reduce the reliance on undocumented behavior I still think this is the right approach. I guess the one benefit of client-side would be if they ever removed support for `data_center` as a query field or added validation that throws an error when its empty.. but is that realistic given the operator has depended on it this long? Is there another scenario you're thinking of that could cause issues that Im missing?

n-losse: Thanks for the detailed answers! 

That the default DC uses an entry with empty string was the missing piece for my understanding. While this all being undocumented it still a bit scary, I now get it and agree. 

That also means that "existing behaviour is unchanged" in the way the feature is phrased is not 100% correct though, right? Maybe we can tweak that description still.

jtucci: @n-losse  Good call, updated the description to be more accurate about the behavior change. Let me know if it reads better now :)

n-losse: Perfect, thanks!

n-losse (approved):
looks good to me, but please still wait for @sdischer-sap as he mentioned

PR #697, opened 2026-06-08
Fix/fix crank missing

SatabdiG:
Pre-installs the Crossplane CLI from releases.crossplane.io instead of GitHub releases, fixing a 404 caused by the binary being renamed from crossplane_linux_amd64 to crank in v2.x.


sdischer-sap (dismissed):
lgtm

sdischer-sap (approved):

PR #698, opened 2026-06-09
ci(e2e): build provider once and reuse across the test matrix

Kukkerem:
## Build the provider once, reuse across the e2e matrix

Splits the e2e workflow into a single `build` job that compiles the image, xpkg, and CLI tools, and a matrix of `e2e-test` jobs that consume those as artifacts instead of each rebuilding.

### Problem
Every matrix shard ran `make e2e`, which invoked `local-build` to rebuild the image and xpkg from scratch for each test function. Each shard also re-fetched `kind`/`kubectl`, exposing CDN rate limits that could fail any shard.

### Fix
- New `build` job (`needs: prepare-matrix`) builds the image + xpkg once with `BUILD_REGISTRY=e2e PLATFORMS=linux_amd64`, fetches `kind`/`kubectl` with a 5-attempt backoff, and uploads image (`docker save`), xpkg, and tools as artifacts (`retention-days: 7`, `if-no-files-found: error`).
- `e2e-test` shards download the artifacts, restore the executable bit `upload-artifact` strips, `docker load` the image, and deploy via a prebuilt path instead of building.
- `Makefile`: add `local-deploy-prebuilt`, which `docker load`s the prebuilt xpkg and tags it as
`$(UUT_XPKG)`; route `test-acceptance` through `ACCEPTANCE_DEPLOY ?= local-build` so local dev rebuilds as before and CI overrides it to `local-deploy-prebuilt`.
- Run e2e via `make test-acceptance` instead of `make e2e`.
- Swap the apt/buildkite Helm install for `azure/setup-helm`.
- Rename `e2e-test-success` to `e2e-tests` and fail it closed on a failed/cancelled `build` job.

### DIFF
--- a/.github/workflows/e2e_test.yaml
+++ b/.github/workflows/e2e_test.yaml
@@ -32,6 +32,8 @@ env:
   # hardcoded BTP CLI version
   btp_cli_version: 2.90.2
   GO_VERSION: '1.25'
+  E2E_BUILD_REGISTRY: e2e
+  E2E_PROVIDER_IMAGE: e2e/provider-btp-amd64
 
 jobs:
   prepare-matrix:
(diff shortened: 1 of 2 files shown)


SatabdiG (approved):
Worked for CF, 🤞 for BTP

PR #699, opened 2026-06-09
fix: expose operationTimeout on ServiceInstance to prevent Terraform timeout loop

### LINKED ISSUE #661: [BUG] ServiceInstance stuck in permanent Creating→Conflict loop when BTP provisioning exceeds 10-minute Terraform timeout
## Summary

When a \`ServiceInstance\` is created for a large service plan (e.g. Kafka Advanced with 9
brokers / 2000G disk), BTP Service Manager takes longer than 10 minutes to provision. A
hardcoded 10-minute Terraform polling timeout fires before BTP finishes, leaving the Terraform
workspace state (\`tfstate\`) empty. The controller then enters a permanent
\`Drift → Update → 409 Conflict\` loop. The instance exists and is healthy in BTP but Crossplane
never adopts it and never reaches \`Ready=True\`.

## Symptoms

On first failure (~10 minutes after creation), \`LastAsyncOperation\` shows:
\`\`\`
Status: False
Reason: ApplyFailure
Message: apply failed: API Error Creating Resource Service Instance (Subaccount):
timeout while waiting for state to become 'succeeded'
(last state: 'in progress', timeout: 10m0s)
try increasing the timeout for this particular operation
\`\`\`

Within ~60 seconds this is overwritten by the subsequent 409 and becomes permanent:
\`\`\`
Status: False
Reason: ApplyFailure
Message: apply failed: API Error Creating Resource Service Instance (Subaccount): Conflict
\`\`\`

Additional steady-state symptoms:
- \`Ready=False, Reason=Creating\` — persists indefinitely
- \`status.atProvider.id\` is empty
- \`crossplane.io/external-name\` stays as the Kubernetes-generated name, never updated to BTP UUID
- Controller logs show a ~60s loop:
  \`\`\`
  Still creating... [10m1s elapsed]
  Creation errored after 10m3s
  Drift detected (delete)
  Successfully requested creation of external resource
  Creating...
  Creation errored after 1s   ← BTP returns 409 Conflict
  \`\`\`

### Layer 1 — hardcoded 10-minute timeout in \`terraform-provider-btp\`

\`terraform-provider-btp\` (bundled binary in the provider image, currently pinned to v1.15.1 in
\`Makefile:11\`) has:

\`\`\`go
// internal/tfutils/tfutils.go:16
const DefaultTimeout = 10 * time.Minute
\`\`\`

This is the default for \`WaitForStateContext\` which polls BTP every 6 seconds for up to 10
minutes (\`btp/provider/resource_subaccount_service_instance.go:458,485\`). Large Kafka clusters
 take longer than 10 minutes to provision in BTP. When the timeout
fires, \`terraform apply\` exits with the error above and **\`tfstate\` stays empty
(\`resources: []\`)**. BTP may or may not have successfully created the instance on its side in later time but Crossplane has
no UUID.

### Layer 2 — async apply callback records timeout as \`ApplyFailure\`

\`crossplane-provider-btp\` runs \`terraform apply\` asynchronously (\`UseAsync=true\`,
\`internal/clients/account/serviceinstance/serviceinstance.go:29\`). When the background
goroutine exits due to the Terraform timeout, the callback
(\`internal/clients/tfclient/tf_async_callback.go:31\`) sets:
- \`LastAsyncOperation=False, Reason=ApplyFailure\` — message contains the timeout text above
- \`AsyncOperation=True, Reason=Finished\`

\`tfstate\` remains empty at this point — no UUID was written.

### Layer 3 — empty tfstate causes Drift → Conflict loop

On the next \`Observe()\` (~60s later):
1. \`terraform apply -refresh-only\` runs with the K8s-generated name as \`id\` (since \`tfstate\`
   is empty, \`crossplane.io/external-name\` was never updated to the BTP UUID)
2. BTP does not recognise the K8s name as a valid instance ID →
   Terraform: \`Drift detected (delete)\` → \`Observe()\` returns \`ResourceUpToDate:false\`
3. \`checkAsyncOperationFailure\` (\`serviceinstance.go:303\`) matches \`Reason=ApplyFailure\` →
   also returns \`ResourceUpToDate:false\`
4. Reconciler calls \`Update()\` → new \`terraform apply\` → BTP returns **HTTP 409 Conflict**
   (the instance with that name was provisioned successfully by BTP during the timed-out apply)
5. \`Conflict\` error overwrites the original timeout message in \`LastAsyncOperation\`
6. Repeat every ~60s indefinitely

## Observed Timeline (live, for advanced cluster)

| Time (UTC) | Event |
|---|---|
| \`17:18:41\` | \`terraform apply\` starts async — \`Creating...\` |
| \`17:18:42–17:28:44\` | \`Still creating... [10s]\` … \`[10m1s]\` — BTP provisioning in progress |
| \`17:28:44\` | \`Creation errored after 10m3s\` — timeout fires, \`tfstate\` stays empty |
| \`17:28:44\` | \`LastAsyncOperation=False, Reason=ApplyFailure\` — timeout message set |
| \`17:29:47\` | Refresh: \`Drift detected (delete)\` → \`Successfully requested creation\` |
| \`17:29:49\` | New apply → \`Creation errored after 1s\` — BTP: Conflict |
| \`17:29:49\` | \`LastAsyncOperation\` message overwritten with \`Conflict\` — loop begins |

Contrast: \`kafka-cluster with less resouces provisioned in **8 minutes**
— within the timeout window — and reached \`Ready=True\` successfully.

## Operational Workaround

1. Find the BTP UUID: BTP Cockpit → Subaccount → Services → Instances → Instance ID
   (or \`btp list services/instance --subaccount <subaccount-guid>\`)
2. Patch the stuck CR:
   \`\`\`bash
   kubectl annotate serviceinstance <name> \
     crossplane.io/external-name=<btp-uuid> --overwrite
   \`\`\`
3. On the next reconcile the provider refreshes by UUID, confirms the instance exists in BTP,
   and adopts it. \`status.atProvider.id\` will be populated and \`Ready\` will transition to \`True\`.

udayakp:
## Summary

Fixes #661

Large BTP service plans (HANA Cloud, Kafka) can take 30–60+ minutes to provision. Terraform's default 10-minute timeout fires before BTP finishes, leaving the `ServiceInstance` in a `Drift → Update → 409 Conflict` loop with no automatic recovery path.

### Root cause

`config/subaccount_service_instance/config.go` had `"timeouts"` in `OmittedFields`, which suppressed the upjet-generated `TimeoutsParameters` block entirely. Even though the CRD struct existed, no `timeouts` block was ever written to `main.tf.json`.

### Changes

- **`config/subaccount_service_instance/config.go`** — remove `"timeouts"` from `OmittedFields` so upjet writes the `timeouts` block to `main.tf.json`
- **`apis/account/v1alpha1/serviceinstance_types.go`** — add optional `operationTimeout *string` field to `ServiceInstanceParameters`
- **`internal/clients/account/serviceinstance/serviceinstance.go`** — wire `operationTimeout` → `Timeouts.{Create, Update, Delete}` in `TfResource()`
- **`internal/clients/account/serviceinstance/serviceinstance_test.go`** — unit test for the new field wiring
- **`package/crds/account.btp.sap.crossplane.io_serviceinstances.yaml`** — regenerated CRD with `operationTimeout` field

### Usage

```yaml
kind: ServiceInstance
spec:
  forProvider:
    operationTimeout: "60m"   # optional; applied to create, update, and delete
    ...
```

### Testing

- Unit tests pass (`make test.run`)
- End-to-end verified on a vcluster: HANA Cloud large plan (60m) and Kafka standard 6-broker plan (30m) both provisioned successfully without hitting the timeout loop

### Notes

As noted in the issue discussion by @sdischer-sap, the `timeouts` field was previously omitted due to an underlying upjet generation limitation. That limitation has since been resolved in upjet, making this fix straightforward to apply.


udayakp:
I had missed the generated Files in my push initially- have added the same and verified the make commands.

sdischer-sap (dismissed):
Nice, thanks for the great contribution. 
We have some pipeline issues, so probably merging might still take a bit. 
But once those issues are resolved will merge your PR.

config/subaccount_service_instance/config.go
```diff
@@ -12,7 +12,7 @@ func Configure(p *config.Provider) {
 		r.ShortGroup = "account"
 		r.Kind = "SubaccountServiceInstance"
 
-		r.ExternalName.OmittedFields = []string{"timeouts"}
```
ankitamohanty7, line 15: what is the relevance of removing this line, this is related to external_name , why are we adding tiemouts back to externalname

config/subaccount_service_instance/config.go
```diff
@@ -12,7 +12,7 @@ func Configure(p *config.Provider) {
...
+		r.ExternalName.OmittedFields = []string{}
```
ankitamohanty7, line 15: Can we configure defaults for serviceinstance, in case user doesn't specify them, current default if 10 mins is low and hangs for bulk creation of small instance or one large instance
Default at the configure file like 
r.operationtimeouts.create = 60 * minute

config/subaccount_service_instance/config.go
udayakp: `OmittedFields` on the `ExternalName` config struct has a specific intended purpose: listing Terraform fields **represented via `crossplane.io/external-name`** so the generator omits them from the CRD schema to avoid duplication. See [upjet source](https://github.com/crossplane/upjet/blob/v2.2.0/pkg/config/resource.go#L151-L157) and [pipeline/crd.go:63](https://github.com/crossplane/upjet/blob/v2.2.0/pkg/pipeline/crd.go#L63) where `deleteOmittedFields` is called during CRD generation.

`"timeouts"` was in `OmittedFields` from the original implementation — the commit that introduced it has no explanation. As noted in #661, it may have been due to an older upjet version not handling the `timeouts` type correctly. The v2 upgrade (PR #653) only changed the import path and did not remove it, so the suppression carried forward.

Removing it from `OmittedFields` with upjet v2 correctly generates `TimeoutsParameters{Create, Update, Delete *string}`, which the mapper wires to the new `operationTimeout` field:

```go
sInstance.Spec.ForProvider.Timeouts = &v1alpha1.TimeoutsParameters{
    Create: t,
    Update: t,
    Delete: t,
}
```

In `terraform-provider-btp`, `timeouts.Create(ctx, tfutils.DefaultTimeout)` uses the value from the `timeouts` block if set, falling back to `DefaultTimeout = 10 * time.Minute`. This PR delivers the configurable field approach suggested in #661, with raising `DefaultTimeout` in `terraform-provider-btp` (`internal/tfutils/tfutils.go`) as a fallback if needed for all consumers. A global large default was not taken as the primary fix since it increases per-instance goroutine and workspace lock duration at scale — `operationTimeout` keeps the large timeout opt-in, scoped only to plans that need it.

config/subaccount_service_instance/config.go
udayakp: Good suggestion. The config file isn't the right place — upjet's resource config struct doesn't have an `OperationTimeouts` field, so `r.operationtimeouts.create = 60 * minute` isn't an available API there.

The actual fallback (`10 * time.Minute`) is `DefaultTimeout` in `terraform-provider-btp/internal/tfutils/tfutils.go`, used by all resource types via `timeoutsLocal.Create(ctx, tfutils.DefaultTimeout)`. Raising it there would apply uniformly to all Terraform and Crossplane users of that provider.

The `operationTimeout` field in this PR covers the per-CR case — users who need a larger window for specific slow plans without changing the global default.

As noted in #661, the suggested path is: configurable field first (this PR), then raise the default upstream as a follow-up if needed. Note that this is a short-term path — the no-fork migration (#207/#680) may replace the tfclient layer and the timeout mechanism with it.

ankitamohanty7: https://github.com/crossplane/upjet/blob/v2.2.0/pkg/config/resource.go#L490

udayakp: You were right — `r.OperationTimeouts` exists on the upjet `Resource` config struct at [`pkg/config/resource.go:490`](https://github.com/crossplane/upjet/blob/v2.2.0/pkg/config/resource.go#L490), and our earlier response incorrectly said it wasn't available. Apologies for that.

After investigating how it flows through upjet: `r.OperationTimeouts` is applied in `BuildMainTF()` after `GetParameters()` has already populated the workspace file — it unconditionally overwrites any `timeouts` value set via `operationTimeout` on the CR, with no conditional check. We also confirmed it is a shared pointer set at provider startup (`Connector.config *config.Resource`), so mutating it per-CR at runtime would be a race condition. There is no hook that runs after it is applied.

One more constraint worth noting: crossplane-provider-btp's `ServiceInstance` uses `UseAsync=true`, which routes through `workspace.ApplyAsync()` in upjet ([`pkg/terraform/workspace.go:169`](https://github.com/crossplane/upjet/blob/v2.2.0/pkg/terraform/workspace.go#L169)). This sets a hard `context.WithDeadline` of `defaultAsyncTimeout = 1h` on the goroutine running `terraform apply`. Since `r.OperationTimeouts` is passed into `WaitForStateContext` which respects the parent context, the effective timeout is always `min(r.OperationTimeouts, 1h)`. Setting `r.OperationTimeouts` above 60m has no practical effect — the goroutine is cancelled at 1h regardless.

`r.OperationTimeouts` therefore acts as the final setting and overwrites any other timeout parameter, with 60m being the effective ceiling. Three options:

1. **`r.OperationTimeouts = 60m`** — covers all known plans (HANA Cloud large: 31–38m, Kafka large: 30m+) and hits the effective ceiling. Simple, no API change. Risk: no per-CR escape hatch if a plan type needs a value within the 10m–60m range, and any operation stuck in `in_progress` due to a non-failure condition will run until the 60m ceiling before surfacing.

2. **Add a default in `TfResource()` + keep `operationTimeout` as optional override** — set 60m as the default in the mapper when `operationTimeout` is not specified. Per-CR override within the 10m–60m range continues to work. Self-contained in this repo.

3. **Keep this PR + raise `DefaultTimeout` in `terraform-provider-btp`** — one-line change in [`internal/tfutils/tfutils.go:16`](https://github.com/SAP/terraform-provider-btp/blob/main/internal/tfutils/tfutils.go#L16). Fixes all three async resources (`btp_subaccount_service_instance`, `btp_subaccount_environment_instance`, `btp_subaccount_subscription`) for all Terraform and Crossplane users. Per-CR `operationTimeout` still works as an override within the 60m ceiling.

Options 2 and 3 preserve per-CR override capability. Option 1 is the simplest. cc @sdischer-sap — would appreciate your view on which option fits best here before we proceed.

sdischer-sap: I would go with 2.
It aligns with what terraform provider offers, which is usually what we try to do. 
Isn't that what you were already doing here?

udayakp: Done — updated to Option 2. The mapper now defaults to `60m` when `operationTimeout` is not set, with per-CR override still working. Rebased onto `main` (conflict with #714 resolved), generated files included, `make reviewable` clean.

xdj2e9 (approved):
LGTM

config/subaccount_service_instance/config.go
xdj2e9: Confirmed that @sdischer-sap suggestion was implemented.

PR #703, opened 2026-06-10
fix: move crank download to build step

sdischer-sap:
Crank downloads started to be an issue again. This PR moves it to the build step to reduce load.


SatabdiG (approved):
lgtm

PR #706, opened 2026-06-11
feat: implement external name annotation for RoleCollectionAssignment

### LINKED ISSUE #554: [TASK] `External-Name`(RoleCollectionAssignment): Ensure ADR compliance
**Part of Epic: #427** 
## Context

## Technical Tasks Checklist

**Acceptance Criteria:**

- [ ] **Migration Logic**:

**Acceptance Criteria:**
- [ ] **Observe() Tests**:

- [ ] **Create() Tests**:

- [ ] **Update() Tests**:

- [ ] **Delete() Tests**:

**Acceptance Criteria:**
- [ ] **Import Flow Test**:

### 5. Documentation

xdj2e9:
External-name support for `RoleCollectionAssignment` per [the external-name ADR](https://github.com/SAP/crossplane-provider-btp/blob/main/docs/contribution-notes/external-name-handling.md), plus two fixes uncovered while wiring the upgrade test.

Closes https://github.com/SAP/crossplane-provider-btp/issues/554

## Why

`RoleCollectionAssignment` still used the default `metadata.name`-as-external-name behavior, which doesn't satisfy the ADR: the identifier on the BTP side is the triple `(origin, user-or-group, roleCollection)`, not `cr.Name`. Without the compound key, imports are impossible and legacy clusters cannot be migrated cleanly.

## Decisions that need review by more experienced colleagues

These are conscious choices; each has a defensible alternative, so calling them out explicitly so reviewers can push back.

- **Three-part compound key merging user and group** (`origin/userOrGroup/roleCollection`) instead of a four-part key with the type as an explicit segment. `userName` / `groupName` are mutually exclusive and immutable via XValidation, so `IdentifierName(cr)` recovers the type from spec without ambiguity. Four-part is the stricter ADR reading.
- **Implicit migration of legacy CRs** (auto-rewrite the annotation when `externalName == cr.Name` and `HasRole` confirms the assignment) instead of failing loudly and asking users to update their resources. Compound key is fully spec-derivable, so the controller can do this deterministically with no risk of duplicate Creates. Caveat: GitOps tools that prune annotations (Argo `Replace=true`) defeat it.
- **Delete on legacy sentinel returns a parse error** rather than rebuilding the key from spec. The race window (legacy match in Observe → MR deleted before `kube.Update` lands) is narrow, self-healing on the next reconcile, and should be very rare or never occur in practice. Rebuilding from spec inside Delete would duplicate logic and silently mask unexpected state.
- **Reject inconsistent imports instead of reporting drift.** When the parsed external-name disagrees with the spec, Observe returns an error (`ErrExternalNameSpecMismatch`) listing which segments mismatch instead of populating an informational `Diff` field as the ADR §Observe step 3 sketches. All four spec fields are immutable, so a mismatch can only originate from a typo in an import manifest and the controller has no way to reconcile it on its own — fail-fast surfaces the problem more directly than a status diff. The "external-name set but assignment missing in XSUAA" case is treated separately as drift (`ResourceExists: false` → Create runs), matching ADR §Observe step 3 verbatim and the `RoleCollection` peer precedent; this preserves self-healing after external revoke.

## What changes

- **`zz_setup.go`**: `DefaultSetup` → `DefaultSetupWithoutDefaultInitializer` so fresh resources reach Observe with `externalName == ""`.
- **`external_name.go`** (new): `BuildExternalName(cr)` formats `origin/userOrGroup/roleCollection`; `ParseExternalName(s)` validates 3 non-empty, untrimmed segments, total length ≤ 512; `externalNameSpecMismatch(cr, parsedOrigin, parsedName, parsedRC)` returns a non-empty description when annotation segments disagree with spec.
- **Observe**: empty → `ResourceExists: false`; legacy sentinel `externalName == cr.Name` → `HasRole` by spec, rewrite the annotation via `kube.Update` on hit; populated → parse, then reject with `ErrExternalNameSpecMismatch` on spec/annotation disagreement, otherwise `HasRole`.
- **Create**: writes the compound key via `meta.SetExternalName` only on success.
- **Update**: replaces the `errNotImplemented` loop with no-op success — all four spec fields are immutable via `XValidation`.
- **Delete**: parses the annotation and revokes by parsed values; parse failure (including empty / legacy sentinel) surfaces as an error rather than silently leaking the XSUAA assignment.
- **Type doc-comment**: standard `External-Name Configuration:` block above `RoleCollectionAssignment` documenting the compound-key format and UI/CLI navigation paths for both user and group entries.
- **Upgrade test** `test/upgrade/rolecollectionassignment_external_name_upgrade_test.go`: baseCR with `externalName == cr.Name` from `v1.9.0`, asserts the annotation migrates to the compound key and the resource stays Ready.
- **E2E import test** `test/e2e/rolecollectionassignment_test.go`: user-only flow via `ImportTester` against a fresh subaccount with the technical user as admin, importing a `RoleCollectionAssignment` for the built-in `Subaccount Viewer` role collection in `sap.default`. Group flow remains unit-tested only.
- Unit tests rewritten for the new Observe / Create / Update / Delete shapes, including dedicated cases for each kind of spec-vs-annotation mismatch (origin, user, role collection, group-vs-user).

### Fixes uncovered while wiring the upgrade test

- **`SubaccountApiCredential` upgrade resource**: `forProvider.name` was *removed* from the fixture. With it set, the SAC's `NameAsExternalName` initializer skipped its rename path on post-upgrade, so the connection secret wasn't deleted and the dependent RCA couldn't connect. `subaccount_api_credential_external_name_upgrade_test.go` actually fails because of that.
- Allow writes for `SubaccountApiCredential` in the RCA upgrade fixtures so role-collection assignments can be exercised end-to-end.

### Build tooling: auto-install of `goimports`

Unrelated to the feature, but uncovered while running `make generate` from a fresh shell: the upjet generator shells out to `goimports`, which nothing in this repo was installing. Aligned with the [upjet provider template](https://github.com/crossplane/upjet-provider-template) by declaring `goimports` as a Go tool in `go.mod`, adding a `make install-tools` target (`go install tool`) wired into `generate.init` so `make generate` / `make reviewable` self-bootstraps, simplifying the matching CI step in `reviewable_check_diff.yaml`, and [noting](https://github.tools.sap/cloud-orchestration/cloud-orchestration.github.tools.sap/pull/1368) in `docs/contribution/1-local-setup.mdx` that `$(go env GOPATH)/bin` must be on PATH (the official Go installer doesn't add it).

## Test plan

- [x] `go build ./...`, `go vet ./...` pass
- [x] Unit tests pass
- [x] Upgrade test `Test_RoleCollectionAssignment_External_Name` green
- [x] `subaccount_api_credential_external_name_upgrade_test.go` green after the SAC fix
- [x] E2E test `TestRoleCollectionAssignmentImportFlow` green in CI
- [x] Manual smoke test against XSUAA (apply fresh MR, observe compound-key annotation, delete, observe revoke)


xdj2e9:
Issue for failing `Test_SubaccountTrustConfiguration_v1alpha1` E2E test: https://github.com/SAP/crossplane-provider-btp/issues/724

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/sap/crossplane-provider-btp
 
 go 1.25.9
 
+tool golang.org/x/tools/cmd/goimports
```
sdischer-sap, line 5: Why do we need that?

internal/controller/security/rolecollectionassignment/rolecollectionassignment.go
```diff
@@ -117,19 +120,78 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
...
+		return managed.ExternalObservation{ResourceExists: false}, nil
+	}
+
+	// Legacy sentinel: pre-ADR controller wrote `cr.Name` via the default
+	// initializer. ADR(external-name): Observe step 1 backwards-compat branch.
+	if externalName == cr.Name {
+		return c.observeLegacy(ctx, cr)
+	}
+
+	return c.observeCurrent(ctx, cr, externalName)
```
sdischer-sap: Personal nitpick: 
Don't really like to split methods like that where the context kind of stays the same, just to reduce size. 
I see that the legacy behaviour should be factored out as an edge case that at some point can be removed from the flow, but "current" really is just regular observing in that sense and does not really add any context other then "this is the other kind of observing". 
But yes, thats probably taste, so no need for changing.

sdischer-sap (dismissed):
Nice work! 
Agree to all decisions taken here. 
Just some questions / nitpick remarks.

go.mod
xdj2e9: When running `make generate` I got the error that `goimports` is not available, because I never installed it. When checking I realized it is not part of any repo config and will not be installed automatically. This adds this as part of this repos go definition and will be installed using `go install tool`. See also related changes in `.github/workflows/reviewable_check_diff.yaml` and `Makefile`.

internal/controller/security/rolecollectionassignment/rolecollectionassignment.go
xdj2e9: 👍  Inlined

go.mod
sdischer-sap: Ok great, I guess i must have installed it before already, because I never so such issues. Makes sense, thank you

sdischer-sap (approved):

PR #707, opened 2026-06-11
ci(e2e): harden test setup against transient flakes

Kukkerem:
## Harden the e2e test setup against transient CI failures

Mitigates three independent flake sources in the e2e workflow, plus runner action bumps.

### Problem
- `kind`/`kubectl`/crank were re-fetched from upstream CDNs on every `build` run; the crank download was un-retried and had been failing repeatedly (#697, #703).
- `kindest/node` is pulled from Docker Hub, whose per-IP rate limit is shared across GitHub-hosted runners, so cluster creation could fail with `failed to pull image`.
- A transient setup error (helm repo, node-image pull) failed the whole matrix shard with no retry.

### Fix
- Cache `.cache/tools` across runs (`actions/cache`, keyed on `k8s_tools.mk` hash + crank version); fetch only on a miss, save only on success. The retry/backoff helper now also wraps the crank download, not just the tool fetch.
- New step routes Docker Hub through `mirror.gcr.io` (`registry-mirrors` in `daemon.json` + `systemctl reload docker`, verified via `docker info`) before kind runs, so `kindest/node` pulls bypass the Hub rate limit.
- Wrap `make test-acceptance` in a 3-attempt loop that retries only on known transient setup signatures and fails fast on real test failures.
- Bump `actions/checkout` v4→v6 and `actions/setup-go` v5→v6 (SHA-pinned).

### DIFF
--- a/.github/workflows/e2e_test.yaml
+++ b/.github/workflows/e2e_test.yaml
@@ -35,6 +35,7 @@ env:
   GO_VERSION: '1.25'
   E2E_BUILD_REGISTRY: index.docker.io/e2e
   E2E_PROVIDER_IMAGE: index.docker.io/e2e/provider-btp-amd64
+  CROSSPLANE_CLI_VERSION: v2.0.2
 
 jobs:
   prepare-matrix:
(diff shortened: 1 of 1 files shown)


SatabdiG (approved):
🔥 🦸

n-losse (approved):

sdischer-sap (approved):

PR #708, opened 2026-06-11
refactor: consolidate connection-detail flatten + add __raw to ServiceBinding/ServiceManager

maximilianbraun:
This PR adds `__raw` to ServiceBinding (and ServiceManager) secrets. That lets consumers pass the binding data raw into other systems that expect to read it directly. CAP is one such consumer.

## What

Three near-duplicate flatten dialects mapped terraform `attribute.credentials` into Crossplane connection details. Two were identical copies of `flattenSecretData`. The CIS path preserved the original blob under `__raw`. ServiceBinding and ServiceManager threw it away.

This PR consolidates the JSON-aware path into one helper, `internal.FlattenConnectionDetails`. It writes `__raw` everywhere, so ServiceBinding and ServiceManager secrets now carry the original credentials blob too. Same convention CIS already uses.

Also drops the duplicate `ResourceRaw = "__raw"` constant in `apis/environment/v1alpha1/cfenvironment_types.go` and routes its only producer through `providerv1alpha1.RawBindingKey`.

## Why

Downstream consumers that need the full `attribute.credentials` JSON to call back into BTP APIs - CF env, Kyma env, Subscription, KymaEnvironmentBinding - already get it from CIS. External consumers like CAP expect the same: paste the secret into their config, done. Today that only works for CIS bindings. After this PR, the same pattern works for plain ServiceBinding and ServiceManager too.

The consolidation is a side benefit. Less drift, one place to fix bugs.

## Behaviour change

ServiceBinding and ServiceManager secrets now contain an extra `__raw` key with the original credentials JSON. Existing consumers that read individual keys are unaffected. Additive only.

## What's in the diff

Refactor + raw key:
- `internal/utils.go` - new `FlattenConnectionDetails` next to `Flatten`.
- `internal/clients/tfclient/tfcontroller.go` - drops local helper, calls shared one.
- `internal/controller/account/servicebinding/secretformat.go` - same (called via `processConnectionDetails`).
- `internal/clients/servicemanager/service_manager_tf_client.go` - appends `RawBindingKey: bindingAsBytes` to the typed map.
- `internal/clients/cfenvironment/client.go` (+ test) - switches to `providerv1alpha1.RawBindingKey`.
- `apis/environment/v1alpha1/cfenvironment_types.go` - drops duplicate `ResourceRaw` constant.

E2E coverage:
- ServiceManager secret check now asserts `__raw` is valid JSON with the expected fields.
- `TestServiceBinding_CreationFlow` adds an assertion for `__raw` shape.
- New `TestServiceBinding_SecretFormatSAPKubernetes` provisions a binding with `secretFormat: sap-kubernetes` and `secretKey: credentials`, verifies the SAP Kubernetes Service Binding metadata keys, the `.metadata` descriptor, and that the bundle key is marked `container: true`. Uses a dedicated `secretformat-env` to stay decoupled from the no-rotation teardown, mirroring `rotation-env`.

## Test plan

- [x] `go vet ./...` clean.
- [x] `go test ./internal/... ./apis/...` all pass.
- [x] `go vet -tags=e2e ./test/e2e/...` clean.
- [x] `grep -rn "flattenSecretData\|ResourceRaw" --include="*.go"` returns zero hits.
- [ ] CI green.
- [ ] E2E green (long-running, includes new sap-kubernetes flow).

WDYT? Happy to split the e2e additions into a follow-up if reviewers prefer keeping the refactor PR small.

### DIFF
--- a/apis/environment/v1alpha1/cfenvironment_types.go
+++ b/apis/environment/v1alpha1/cfenvironment_types.go
@@ -20,7 +20,6 @@ const (
 	ResourceAPIEndpoint = "apiEndpoint"
 	ResourceOrgId       = "orgId"
 	ResourceOrgName     = "orgName"
-	ResourceRaw         = "__raw"
 )
 
 // User identifies a user by username and origin
--- a/internal/clients/cfenvironment/client.go
+++ b/internal/clients/cfenvironment/client.go
@@ -11,6 +11,7 @@ import (
 	provisioningclient "github.com/sap/crossplane-provider-btp/internal/openapi_clients/btp-provisioning-service-api-go/pkg"
 
 	"github.com/sap/crossplane-provider-btp/apis/environment/v1alpha1"
+	providerv1alpha1 "github.com/sap/crossplane-provider-btp/apis/v1alpha1"
 	"github.com/sap/crossplane-provider-btp/internal"
 )
 
@@ -105,7 +106,7 @@ func GetConnectionDetails(instance *provisioningclient.BusinessEnvironmentInstan
 		return managed.ConnectionDetails{}, err
 	}
 	details := managed.ConnectionDetails{
-		v1alpha1.ResourceRaw: label,
+		providerv1alpha1.RawBindingKey: label,
 	}
 
 	if orgName, ok := cflabels["Org Name"]; ok {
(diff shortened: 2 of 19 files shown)


sdischer-sap (approved):
lgtm

sdischer-sap:
Retrigger workflows via close -> reopen

PR #713, opened 2026-06-15
fix(subaccount): surface BTP error bodies; treat 409 on delete as soft

maximilianbraun:
## Summary

- `Delete` on `Subaccount` previously wrapped a bare 409 from the BTP API without surfacing the response body, so logs could not distinguish *"delete already in progress"* from *"blocked by child resources"*. All BTP API error sites in this controller (`deleteBTPSubaccount`, `moveSubaccountAPI`, `updateSubaccountAPI`, `migrateExternalName`/`listSubaccounts`) now route through the existing `specifyAPIError` helper, which extracts the structured `ApiExceptionResponseObject` or raw body from `*accountclient.GenericOpenAPIError`.
- HTTP 409 on `DeleteSubaccount` is now soft-handled: the controller logs the body (capped at 1 KiB via a new `readBodySafe` helper) and returns `nil` so the next reconcile re-runs `Observe`. Repeated `DELETE`s amplify the conflict and pollute the error log; the BTP-side status will eventually flip to `DELETING` (the guard in `Delete` already short-circuits that) and `Observe` will eventually detect the final 404.
- Defensive nil-check added to the existing 404 guard (`raw != nil && raw.StatusCode == 404`) so a transport-level error that returns `(nil, nil, err)` no longer panics.

Addresses Pattern E in `docs/development/flakiness-analysis.md` (`TestCloudFoundryEnvironment` 409 on subaccount delete).

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -race -count=1 ./internal/controller/account/subaccount/...` — all existing tests pass; new `DeleteAPI409SoftHandled` and `DeleteAPI5xxWrappedWithSpecifyAPIError` cases pass.
- [x] `go test -race -count=1 ./...` — full suite green.
- [x] `gofmt -l` clean on both modified files.
- [ ] CI must exercise the real-BTP path; the actual 409-on-delete scenario can only be reproduced against a live global account.

### DIFF
--- a/internal/controller/account/subaccount/subaccount.go
+++ b/internal/controller/account/subaccount/subaccount.go
@@ -3,6 +3,7 @@ package subaccount
 import (
 	"context"
 	"fmt"
+	"io"
 	"net/http"
 	"reflect"
 	"strings"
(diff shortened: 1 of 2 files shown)


maximilianbraun:
Superseded by #716 — recreated as part of stacked chain (#716 → #717 → #718 → #719) to test combined effect on e2e flakiness.

PR #714, opened 2026-06-15
feat: add labels field to ServiceInstance

### LINKED ISSUE #715: feat: support labels on ServiceInstance
The BTP Service Manager API supports labels on service instances (key to list-of-values metadata). These labels are visible in the BTP cockpit and queryable via the SM API's labelQuery parameter. The BTP Service Operator already supports this.

Our provider does not currently expose a labels field on the ServiceInstance CRD, so there is no way to set BTP labels on instances created through the provider.

**Solution**

Add a `labels` field to the ServiceInstance `forProvider` spec. The field is a map of string keys to lists of string values, matching the SM API format. Labels are passed through on create and update.

PR: #714

jtucci:
Adds a labels field to the ServiceInstance forProvider spec. Labels are passed through to the BTP Service Manager API on create and update, matching the format used by the BTP Service Operator (map of string keys to list of string values).

Closes #715


sdischer-sap (approved):
Looks easy enough ;) 
Thank you

PR #716, opened 2026-06-15
fix(subaccount): surface BTP error bodies on delete via Synced condition

maximilianbraun:
## Summary

- All BTP API error sites in this controller (`deleteBTPSubaccount`, `moveSubaccountAPI`, `updateSubaccountAPI`, `migrateExternalName`/`listSubaccounts`) now route through the existing `specifyAPIError` helper, which extracts the structured `ApiExceptionResponseObject` (or raw body) from `*accountclient.GenericOpenAPIError`. Previously the bare transport error was wrapped, so logs and the resource's `Synced` condition could not distinguish *"delete already in progress"* from *"blocked by child resources"*.
- 409 on `DeleteSubaccount` is no longer special-cased — it falls through to the same `specifyAPIError` wrapping as any other failure, so the BTP message lands in `Synced: False` and is visible via `kubectl describe subaccount`. The existing `Status == DELETING` guard at the top of `Delete()` still short-circuits subsequent reconciles once `Observe` records the in-flight state, and crossplane-runtime applies exponential backoff between retries.
- Defensive nil-check added to the existing 404 guard (`raw != nil && raw.StatusCode == 404`) so a transport-level error that returns `(nil, nil, err)` no longer panics.

Addresses Pattern E in `docs/development/flakiness-analysis.md` (`TestCloudFoundryEnvironment` 409 on subaccount delete).

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -race -count=1 ./internal/controller/account/subaccount/...` — existing tests pass; new `DeleteAPI409SurfacesError` and `DeleteAPI5xxWrappedWithSpecifyAPIError` cover the wrapped-error contract.
- [x] `go test -race -count=1 ./...` — full suite green.
- [x] `gofmt -l` clean on both modified files.
- [ ] CI must exercise the real-BTP path; the actual 409-on-delete scenario can only be reproduced against a live global account.


internal/controller/account/subaccount/subaccount.go
```diff
@@ -334,18 +335,51 @@ func deleteBTPSubaccount(
...
+		// 409 Conflict: delete already in progress, or blocked by child resources
+		// (entitlements, subscriptions, environments, trust configs). Return nil so
+		// the controller polls Observe on the next reconcile instead of retrying
+		// DELETE — repeated DELETEs amplify the conflict and pollute the error log.
+		// The BTP-side state will eventually become DELETING, at which point the
+		// guard in Delete() short-circuits, and Observe will detect the final
+		// 404 (deleted).
+		ctrl.Log.Info("subaccount delete returned 409 Conflict (in-progress or blocked by children), waiting for next reconcile",
+			"subaccountId", subaccountId, "body", string(readBodySafe(err, raw)))
+		return nil
```
sdischer-sap: Doesn't that mean mean we treat 409 as successful deletions in general?
So we will have an "successfully deleted" event and no indication what so ever that something is not working. 
In case of "deleting in progress" that might be fine, but in case of blocked deletions due to existing child resources that doesn't seem right.

maximilianbraun: true, lets keep it to bubble up the message

sdischer-sap (approved):

sdischer-sap:
@maximilianbraun you put the do-not-merge label. 
Is there a reason for that?

Issue #521, opened 2026-02-20
Use v1.22 of btp terraform provider


sdischer-sap:
## Context
As we found in #520 the latest release of the btp terraform provider is incompatible with our provider. A lot of tests where failing with the error below:
```
observe failed: cannot run refresh: refresh failed: Missing Resource
      Identity After Read: The Terraform Provider unexpectedly returned no resource
      identity data after having no errors in the resource read. This is always an
      issue in the Terraform Provider and should be reported to the provider developers.
```
We reached out to the terraform folks and got some response back in https://github.com/SAP/terraform-provider-btp/issues/1532. Apparently the issues are arising from a new terraform feature called identities, which properly have to be used on the users side. Back then I didn't succeed in getting it work in upjet without no fork. 

## Scope
Ideally we should just update to the latest terraform provider here and make existing clients compatible with the changes with this terraform identity changes. Maybe I missed something in my previous checks, because I remember that identities generally can be configured in upjet. If not we will probably need to put all the additional changes of this no-fork implementation into on branch until everything is stable and we can merge to main.

## Technical Steps / Acceptance Criteria
- remove fix for the BUG  before migrating to the new TF  version (temporary solution which fixed in the new TF)-> https://github.com/SAP/crossplane-provider-btp/issues/724
- update to latest TF version (today -> 1.23.1) of terraform provider
- adapt identity changes and make it compatible
- **try first without no-fork** if no possible look next item
- if it won't work without no-fork clients we just need to keep the changes in a branch and keep it until we migrated all of the resources to no-fork (which are the subissues in this epic)

**Additional context**
- [BTP Terraform Release v1.22](https://github.com/SAP/terraform-provider-btp/releases/tag/v1.22.0) (use the latest version of TF provider -> today 1.23.1)
- we have a POC which contains most of the cases required for the upgrade - [POC link](https://github.com/SAP/crossplane-provider-btp/issues/201)

lechnerc77:
Please cross-check with release [1.20.1](https://github.com/SAP/terraform-provider-btp/releases/tag/v1.20.1). This error occured for OpenTofu that does not support import identities and we ensured the compatibility with PR https://github.com/SAP/terraform-provider-btp/pull/1384

sdischer-sap:
@lechnerc77 thanks for that hint. I just retried that with the latest 1.22.0 release and unfortunately found the same issues. Created some detailed issue in https://github.com/SAP/terraform-provider-btp/issues/1532.

PR #725, opened 2026-06-18
fix: skip description on SubaccountTrustConfiguration to avoid inconsistent-result error

### LINKED ISSUE #724: [BUG] SubaccountTrustConfiguration: Create fails with "inconsistent result after apply" when `description` is set
**Describe the bug**

Creating a `SubaccountTrustConfiguration` with `spec.forProvider.description` set fails on the first reconcile with `"Provider produced inconsistent result after apply"`. The managed resource enters a stuck state and every subsequent reconcile plans a destroy+recreate. The Kubernetes object cannot reach `Ready=True` and cannot be deleted cleanly via `kubectl delete` without manual intervention.

The error from the Terraform plugin framework is:

```
Provider produced inconsistent result after apply
.description: was cty.StringVal("<user value>"),
  but now cty.StringVal("IAS tenant <tenant>.accounts.ondemand.com (OpenID Connect)").
```

`GlobalaccountTrustConfiguration` is **not** affected (single-step Create on the BTP API side).

The bug originates in `terraform-provider-btp` (≤ v1.22). Subaccount trust configuration creation is implemented inside `terraform-provider-btp` as a two-step flow against the BTP backend API:

1. `CreateBySubaccount` — passes `description` from the plan.
2. `UpdateBySubaccount` — `description` was **omitted** from the update request, so the BTP backend's auto-generated default (`"IAS tenant <tenant>.accounts.ondemand.com (OpenID Connect)"`) overwrites the user-provided value between the two calls.

`terraform-plugin-framework` then detects that the post-apply state diverges from the plan and raises the inconsistent-result error. This is reported upstream as [SAP/terraform-provider-btp#1578](https://github.com/SAP/terraform-provider-btp/issues/1578) and fixed by [PR #1579](https://github.com/SAP/terraform-provider-btp/pull/1579) in upstream **v1.23.1** (2026-06-09).

v2.0.0-rc1 (latest release at the time of writing). All released versions of `crossplane-provider-btp` are affected: v1.0.0–v1.3.0 pin `terraform-provider-btp` v1.7.0 and v1.4.0 onwards pin v1.15.1, and both upstream pins contain the same buggy two-step Create flow (only fixed in upstream v1.23.1). The current pin on `main` is `Makefile: TERRAFORM_PROVIDER_VERSION ?= 1.15.1`. Upjet v2.2.0 is in use, configured with the Terraform CLI-forking client (the "fork" variant).

**To Reproduce**

1. Install the provider on a Kubernetes cluster.
2. Apply a `SubaccountTrustConfiguration` with a non-empty `description`, for example:

   ```yaml
   metadata:
     name: e2e-trustconfiguration
   spec:
     forProvider:
       name: pr-706-e2e-simple-trust
       description: "my-descriptions"
       subaccountRef:
         name: sub-trust-test
       identityProvider: "opensourcetest.accounts.ondemand.com"
   ```

3. Observe that the resource never reaches `Ready=True`. Inspect the managed resource's events / conditions and see the `"Provider produced inconsistent result after apply"` error referencing `.description`.
4. Attempt to `kubectl delete` the resource and observe that it cannot be cleanly deleted without manual intervention.

The same scenario fails our e2e test `Test_SubaccountTrustConfiguration_v1alpha1`. The fixture is at `test/e2e/testdata/crs/SubaccountTrustConfiguration/trustconfiguration.yaml`.

**Expected behavior**

`SubaccountTrustConfiguration` reconciles successfully when `description` is set: the resource reaches `Ready=True`, the user-provided `description` is honored (or, if dropped silently as a quick-fix interim, the resource still reconciles successfully and is deletable).

**Screenshots**

n/a

**Additional context**

The proper fix is to bump `TERRAFORM_PROVIDER_VERSION` from `1.15.1` to `1.23.1` (or later), which is already tracked in [#521](https://github.com/SAP/crossplane-provider-btp/issues/521).

A quick fix should land before that bump so users can create `SubaccountTrustConfiguration` resources at all in the meantime. Because `description` is informational only (not used for matching, references, or wiring), silently dropping the user-provided value is considered acceptable for users in the interim.

The quick fix registers a managed-resource [`Initializer`](https://github.com/crossplane/crossplane-runtime/blob/v2.2.1/pkg/reconciler/managed/reconciler.go#L218-L220) that nils `spec.forProvider.description` (and the `initProvider` counterpart) in-process on every reconcile, before upjet serializes the spec into `main.tf.json`. The mutation is **not** persisted to the API server, so the user-provided value remains intact in etcd; on the next reconcile the value reappears and is nilled again. Terraform plans against a nil description; combined with `Optional+Computed` in the upstream schema, the BTP-computed value materializes from nil and the framework's plan-vs-state consistency check accepts it. The same Initializer hook is already used in this file by `OriginInitializer`.

Concrete change in `config/subaccount_trust_configuration/config.go`, inside the existing `AddResourceConfigurator`:

```go
r.InitializerFns = append(r.InitializerFns, func(_ client.Client) managed.Initializer {
    return &descriptionStripper{}
})
```

with a small helper type defined alongside:

```go
type descriptionStripper struct{}

func (descriptionStripper) Initialize(_ context.Context, mg resource.Managed) error {
    cr, ok := mg.(*securityv1alpha1.SubaccountTrustConfiguration)
    if !ok {
        return errors.New(errTypeAssertion)
    }
    cr.Spec.ForProvider.Description = nil
    cr.Spec.InitProvider.Description = nil
    return nil
}
```

In addition, the field's [`ArgumentDocs`](https://github.com/crossplane/upjet/blob/v2.2.0/pkg/registry/resource.go#L55-L57) entry is overridden so the limitation is surfaced on the user-facing CR — the override flows into the generated CRD field description visible via `kubectl explain` and the published reference docs.

> A previously proposed mechanism — using upjet's `TerraformConfigurationInjector` to delete the `description` key from the Terraform parameters map — turned out to be a no-op on this provider's runtime. The injector is only honored by upjet's no-fork SDKv2/plugin-framework controller paths; this provider runs on the CLI-forking controller, which has no equivalent hook. The Initializer-based approach is the working alternative.

Verification and follow-up:

- The e2e test `Test_SubaccountTrustConfiguration_v1alpha1` should pass with the fixture's `description` value retained.
- The `SubaccountTrustConfiguration` CRD documentation already notes that `description` is currently ignored, via the `ArgumentDocs` override.
- Once #521 lands and `TERRAFORM_PROVIDER_VERSION` is at v1.23.1 or later, the Initializer registration, the `descriptionStripper` type, and the `ArgumentDocs` override should be removed.

References:

- Existing tracking issue: [SAP/crossplane-provider-btp#521](https://github.com/SAP/crossplane-provider-btp/issues/521)
- Upstream bug: [SAP/terraform-provider-btp#1578](https://github.com/SAP/terraform-provider-btp/issues/1578)
- Upstream fix: [SAP/terraform-provider-btp#1579](https://github.com/SAP/terraform-provider-btp/pull/1579)
- Fixed upstream release: [terraform-provider-btp v1.23.1 (2026-06-09)](https://github.com/SAP/terraform-provider-btp/releases/tag/v1.23.1)
- Upjet release used here: [crossplane/upjet v2.2.0](https://github.com/crossplane/upjet/releases/tag/v2.2.0)

xdj2e9:
Fixes #724.
Backport into `release/v1_10_x`: https://github.com/SAP/crossplane-provider-btp/pull/730

## Summary

`SubaccountTrustConfiguration` creation fails with *"Provider produced inconsistent result after apply"* whenever `spec.forProvider.description` is set, due to a bug in `terraform-provider-btp` ≤ v1.22 (fixed upstream in v1.23.1; the corresponding bump is tracked in #521). Until that bump lands, this PR adds a temporary workaround so users can create the resource at all.

See #724 for the full root cause analysis, the upstream references, and the rationale for the workaround mechanism.

## What I changed

- Registered an `Initializer` (`descriptionStripper`) on the `btp_subaccount_trust_configuration` resource configuration that nils `spec.forProvider.description` and `spec.initProvider.description` in-process on every reconcile. The mutation is not persisted to the API server, so the user-provided value remains intact in etcd.
- Overrode the field's `ArgumentDocs` so the limitation is surfaced on the user-facing CRD field description (visible via `kubectl explain` and the published reference docs).
- Both pieces are grouped behind a single `applyDescriptionInconsistencyWorkaround` helper carrying a `TODO` referencing #521 so the workaround is removed in one place when the upstream bump lands.

`GlobalaccountTrustConfiguration` is intentionally untouched — upstream confirmed the single-step Create flow is not affected.

## Test Plan

- [x] `go build ./...` — clean
- [x] `make generate` — generated CRD and Go types pick up the overridden field description on `SubaccountTrustConfiguration` only; `GlobalaccountTrustConfiguration` is unchanged
- [x] e2e `Test_SubaccountTrustConfiguration_v1alpha1` — passes with the fixture's `description: "my-descriptions"` retained

### DIFF
--- a/apis/security/v1alpha1/zz_subaccounttrustconfiguration_types.go
+++ b/apis/security/v1alpha1/zz_subaccounttrustconfiguration_types.go
@@ -35,7 +35,7 @@ type SubaccountTrustConfigurationInitParameters struct {
 	// Determines that end users can choose the trust configuration for login. If not set, the trust configuration can remain active, however only application users that explicitly specify the origin key can use if for login.
 	AvailableForUserLogon *bool `json:"availableForUserLogon,omitempty" tf:"available_for_user_logon,omitempty"`
 
-	// (String) Description of the trust configuration.
+	// (String) Description of the trust configuration. NOTE: currently ignored due to an upstream bug (see #724); the BTP backend assigns its own default.
 	// Description of the trust configuration.
 	Description *string `json:"description,omitempty" tf:"description,omitempty"`
 
@@ -86,7 +86,7 @@ type SubaccountTrustConfigurationObservation struct {
 	// Determines that end users can choose the trust configuration for login. If not set, the trust configuration can remain active, however only application users that explicitly specify the origin key can use if for login.
 	AvailableForUserLogon *bool `json:"availableForUserLogon,omitempty" tf:"available_for_user_logon,omitempty"`
 
-	// (String) Description of the trust configuration.
+	// (String) Description of the trust configuration. NOTE: currently ignored due to an upstream bug (see #724); the BTP backend assigns its own default.
 	// Description of the trust configuration.
 	Description *string `json:"description,omitempty" tf:"description,omitempty"`
 
@@ -146,7 +146,7 @@ type SubaccountTrustConfigurationParameters struct {
 	// +kubebuilder:validation:Optional
 	AvailableForUserLogon *bool `json:"availableForUserLogon,omitempty" tf:"available_for_user_logon,omitempty"`
 
-	// (String) Description of the trust configuration.
+	// (String) Description of the trust configuration. NOTE: currently ignored due to an upstream bug (see #724); the BTP backend assigns its own default.
 	// Description of the trust configuration.
 	// +kubebuilder:validation:Optional
 	Description *string `json:"description,omitempty" tf:"description,omitempty"`
--- a/config/subaccount_trust_configuration/config.go
+++ b/config/subaccount_trust_configuration/config.go
@@ -23,12 +23,7 @@ func Configure(p *config.Provider) {
 		r.ShortGroup = "security"
 		r.Kind = "SubaccountTrustConfiguration"
 
-		// Workaround for inconsistent description on create, see #724.
-		// TODO: remove when terraform-provider-btp is bumped to >= 1.23.1 (#521).
-		r.TerraformConfigurationInjector = func(jsonMap, tfMap map[string]any) error {
-			delete(tfMap, "description")
-			return nil
-		}
+		applyDescriptionInconsistencyWorkaround(r)
 
 		r.References["subaccount_id"] = config.Reference{
 			Type:              "github.com/sap/crossplane-provider-btp/apis/account/v1alpha1.Subaccount",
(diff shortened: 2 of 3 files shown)


xdj2e9:
~Please do not merge. Behavior of existing CRs has to be checked.~ ✅ 

It is confirmed that existing resources will not change their description as the `null` field in the terraform plan will be treated as "no change, leave it as it is".

sdischer-sap (approved):
lgtm, great work! thank you+

Issue #726, opened 2026-06-19
[BUG] Isolate long running e2e tests in own make goal


sdischer-sap:
## Context
Long running e2e tests are actually running all e2e tests right now, not just the ones marked as long running tests. 

## Why is that a problem
Its rather difficult to troubleshoot those, because 
A - running all tests at ones might take two hours and 
B - overall e2e tests are flaky, which raises the chance of not 

## Solution 
Not sure if that was intentional, but since we do run all other e2e tests as part of every PR and release I would prefer to isolate long runing e2e tests into a dedicated make goal. 

## Implementation proposal
Currently we rely on testing.Skip() for manually skipping tests based on a go test flag. 
Using that for excluding other e2e tests from long running tests would require adding similar checks in each and every test. 

Better approach would be to use golang build tags for this. 

## Rollout requirements
This should be backported to our v1.10.x release branch. (see #685)

sdischer-sap:
Merged to main and backported. to release branch

PR #728, opened 2026-06-19
fix: dedicated e2e_long build tag for long-running tests

sdischer-sap:
Relates to: #726 

## Summary
Changes make goal `test-e2e-long` to only execute long running tests. 

### Side Quests:
- stream loggs for visibility in CI
- filter long running tests in CI instead of executing them in skip mode, declutters PR checks

## Claude details
- replaced the `testing.Short()` pattern for skipping long-running e2e tests with a dedicated `e2e_long` build tag
- `resource_utils.go`, `import_utils_test.go`, and `main_test.go` now accept both `e2e` and `e2e_long` tags so shared helpers are available to both suites
- `Makefile` updated: `test-e2e-long` uses `-tags=e2e_long` (was `-tags=e2e -short`); `test-acceptance` drops the `-short` flag (long tests are simply not compiled in); debug target aligned accordingly
- prepare-matrix GH action adapted to exclude long running tests
- Timeout for `test-e2e-long` reduced from 240m to 150m, since now its we do not run that much anymore


.github/workflows/e2e_test.yaml
```diff
@@ -70,7 +70,7 @@ jobs:
       - name: Set matrix with test names
         id: set-matrix
         run: |
-          echo "test-names=$(grep -rh --include='**_test.go' -o -E 'func (Test[A-Za-z0-9_]+)\(t \*testing.T\)' test/e2e | sed 's/func //;s/(t \*testing.T)//' | jq -R . | jq -cs .)" >> $GITHUB_OUTPUT
+          echo "test-names=$(grep -rhl --include='*_test.go' 'go:build e2e' test/e2e | xargs grep -L 'go:build e2e_long' | xargs grep -oh -E 'func (Test[A-Za-z0-9_]+)\(t \*testing\.T\)' | sed 's/func //;s/(t \*testing.T)//' | jq -R . | jq -cs .)" >> $GITHUB_OUTPUT
```
n-losse: nitpick: the -h does nothing with the -l i believe

minor: the "xargs grep -L 'go:build e2e_long'" depends on the order of "e2e || e2e_long" if a test is for both, that might be surprising when adding more tests later

xdj2e9: From what I understand the test tags are a go native mechanism (build constraints). would it be benefitial for us to rely on that to get the list of tests to execute (using something like `go test -list [...]`)?

The runner would need go tooling available in that case though.

sdischer-sap: Nice catch @n-losse, we need those kind of reviews! ;) 

Yes the test cases actually only have either e2e or e2e_long right now and I would vote for keeping it that way semantically. But yes to make it technically robust, I changed it to:

Just exclude `go:build e2e_long`, any test should have e2e... already in there. So that should make sure that we only exclude the ones that ONLY are marked for e2e_long. So we can keep it open to include them into PRs with putting `go:build e2e || e2e_long` in the future.

sdischer-sap: @xdj2e9 didn't see your suggestion earlier.
Nice idea. I had no clue about that possibility. Just tested it though and it looks like unfortunately its not possible in this scenario. 

The command compiles and runs the go code and just skips the individual test cases. In our case, since most of the pluming for kind cluster creation and such is in the `TestMain` all of that is performed on that command call. So yeah I would say that defeats the purpose a little.

xdj2e9: Thank you for looking into it. It was a long shot.

n-losse: but i see two tags in places - so those will still be included 

image

whereas a test with `go:build e2e_long || e2e` would still be dropped - right?

sdischer-sap: @n-losse 
Those are only helper functions and do not contain test cases on their own. They need to be available in both build scenarios, because compiling fails other wise. 

This github action step only garthers the names of test cases. So we do not need to take are about compiling or things like that here.

n-losse (approved):

PR #731, opened 2026-06-20
fix(directory): surface BTP error bodies via specifyAPIError

maximilianbraun:
## Summary
- Route `DeleteDirectory` and `UpdateDirectory` non-404 paths through the existing `specifyAPIError` helper so 409/5xx BTP error bodies surface in the `Synced` condition instead of bare transport errors.
- Mirrors PR #716 (subaccount).

## Test plan
- [x] `go build ./... && go vet ./...`
- [x] `go test -race -count=1 ./internal/clients/directory/... ./internal/controller/account/directory/...`
- [x] New `DeleteAPI409SurfacesError` unit test exercising a `GenericOpenAPIError` with `ApiExceptionResponseObject` model.
- [x] Existing 404 short-circuit and other delete/update tests still pass.


maximilianbraun (commented):
internal/clients/directory/directory_test.go:880 — Drop the local `create409Error`; it is a verbatim copy of subaccount_test.go's helper (same `unsafe`-pointer reflection over `GenericOpenAPIError`'s unexported fields). Move one copy to `internal/testutils` (or wherever the BTP test helpers live) and import it from both packages — duplicating a reflect/unsafe trick guarantees one of them rots silently when the SDK regenerates.

maximilianbraun (commented):
Helper is appropriately minimal: every parameter has a caller using it, the empty-message branch lets the original code-only call site stay one-liner, and the regeneration-rot warning is the right ceiling note for unsafe-reflection. Subaccount_test.go edit is the deduplication itself, not scope creep. Lean.

sdischer-sap (approved):

PR #732, opened 2026-06-20
fix(servicemanager): surface BTP error bodies in api_plan_resolver

maximilianbraun:
## Summary

- `PlanIDByName` in `internal/clients/servicemanager/api_plan_resolver.go` returned the bare transport error from `GetServiceOfferings` and `GetAllServicePlans`, hiding the BTP-side message carried in `*smclient.GenericOpenAPIError`. Both call-sites now route through a new SM-flavoured `specifyAPIError` helper.
- New `internal/clients/servicemanager/errors.go` mirrors the subaccount helper landed in #716: prefer the structured `smclient.Error` model, fall back to raw body, otherwise return the original error untouched. The helper lives in its own file (not next to `PlanIDByName`) so PR5 (`service_instance_proxy_client.go`, same package) can reuse it without an import cycle or a copy.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -race -count=1 ./internal/clients/servicemanager/...` — new `TestSpecifyAPIError` covers nil, non-OpenAPI passthrough, structured-model surfacing, raw-body fallback, and no-model/no-body passthrough.
- [x] `gofmt -l internal/clients/servicemanager/` clean.

### DIFF
--- a/internal/clients/servicemanager/api_plan_resolver.go
+++ b/internal/clients/servicemanager/api_plan_resolver.go
@@ -124,7 +124,7 @@ func (sm *ServiceManagerClient) PlanIDByName(ctx context.Context, offeringName, 
 
 	execute, _, err := sm.GetServiceOfferings(ctx).FieldQuery(offeringQuery).Execute()
 	if err != nil {
-		return "", err
+		return "", specifyAPIError(err)
 	}
 	if len(execute.Items) == 0 {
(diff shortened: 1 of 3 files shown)


maximilianbraun (commented):
Lean. No over-engineering findings.

sdischer-sap (approved):

PR #733, opened 2026-06-20
fix(servicemanager): surface BTP error bodies + nil-guard in proxy client

maximilianbraun:
## Summary

- Wrap transport errors in `service_instance_proxy_client.go` (`describeAdminBinding`, `createAdminBinding`, `deleteAdminBinding`) through a local `specifyAPIError` helper so the BTP accounts-service API body lands in the error message instead of the opaque transport string. Mirrors the helper added in #716 for subaccount.
- Nil-guard `raw` before reading `raw.StatusCode == 404` in `describeAdminBinding` (previously panicked on transport errors that returned a nil `*http.Response`).
- Unit test asserts the surfaced body for `createAdminBinding`.

## Helper duplication (intentional)

This is the 4th file-local copy of essentially the same helper (subaccount, directory, entitlement, here). Two reasons it stays local in this PR:
1. The proxy client uses the **accounts-service** OpenAPI client (`accountsserviceclient`), not the service-manager OpenAPI client, so PR4's upcoming `servicemanager/errors.go` helper (typed for `*smclient.GenericOpenAPIError`) does not apply.
2. Extracting a shared accounts-service helper across packages is a real refactor and out of scope for this bugfix fan-out.

A follow-up consolidation PR is warranted once all six error-body PRs land.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -race -count=1 ./internal/clients/servicemanager/...`
- [x] `gofmt -l internal/clients/servicemanager/`


maximilianbraun (commented):
Lean. The 4th copy of `specifyAPIError` is the team-lead's deferred consolidation, not a finding.

sdischer-sap (dismissed):

sdischer-sap:
There are some compiler errors in make reviewable in this one.

sdischer-sap (approved):

Issue #161, opened 2025-04-30
Remove deprecated `GlobalAccount`


GenosseOtt:
## Context
As we have deprecated GlobalAccount some time ago, lets remove completely to keep docs and end user complexity as little as possible

We evaluate upfront: https://github.com/SAP/crossplane-provider-btp/issues/371

## Acceptance criteria
- Remove all code and CRD from `GlobalAccount` resource
- Remove references to the CRD from schemas

jn-av:
Note: let's check how deletion of a resource works with our upgrade tests

PR #734, opened 2026-06-20
fix(globalaccount): surface BTP error bodies via specifyAPIError

maximilianbraun:
## What

Route `GetGlobalAccount().Execute()` errors through a local `specifyAPIError` helper (mirroring `internal/controller/account/subaccount/subaccount.go`) before wrapping with `errGetGlobalAccount`.

## Why

Without this, when the BTP accounts API returns an error, the controller surfaces only the generic OpenAPI client error string. The actual BTP-side message — carried in `*accountclient.GenericOpenAPIError`'s `Body()` or `ApiExceptionResponseObject` model — never reaches the `GlobalAccount`'s `Synced=False` condition, so operators have to scrape controller logs to figure out what BTP is complaining about.

This mirrors PR #716 for `Subaccount`.

## How

- Added a local `specifyAPIError` helper to `globalaccount.go` (same shape as subaccount's: model branch first, then raw `Body()`).
- Wrapped the only `.Execute()` call in this file (`GetGlobalAccount` in `Observe`) through it. No other `.Execute()` calls exist in `globalaccount.go`.
- Added a unit test asserting the `"API Error"` prefix lands when `GetGlobalAccount` returns a 5xx `GenericOpenAPIError` with an `ApiExceptionResponseObject` model.
- Added a small `mockGlobalAccountClient` (interface satisfier with `panic("not implemented")` stubs for the 4 unused methods).

## Test

`go build ./... && go vet ./... && go test -race -count=1 ./internal/controller/account/globalaccount/...` — all green; `gofmt` clean.


maximilianbraun (commented):
internal/controller/account/globalaccount/fakes.go — Rename to `fakes_test.go`. The file is currently `package globalaccount` (not `_test`) so `mockGlobalAccountClient` plus its four `panic("not implemented")` stubs compile into the production binary; the `_test.go` suffix gates it to test builds at zero cost.

internal/controller/account/globalaccount/globalaccount.go:163 — Replace `errors.New(fmt.Sprintf("API Error: %v, Code %v", ...))` with `fmt.Errorf("API Error: %v, Code %v", ...)`. The next branch already does this; one idiom is enough.

sdischer-sap (approved):
This is a bit pointless, because we plan to remove the globalaccount with the next release (see #161), however since we have the PR we might as well merge it.

PR #735, opened 2026-06-20
fix(kymabinding): surface BTP error bodies on Describe + Delete

maximilianbraun:
## Summary

- `DescribeInstance` and `DeleteInstances` in `internal/clients/kymaenvironmentbinding/binding.go` now route their API errors through the existing package-level `specifyAPIError` helper (already used by `CreateInstance`). Previously the bare transport error was wrapped, so the structured `ApiExceptionResponseObject` body never reached the resource's `Synced` condition.
- The deliberate 404 swallow in `DeleteInstances` is preserved (covered by a regression test).
- Fixes a pre-existing copy/paste bug where `DescribeInstance`'s wrap message was `errKymaBindingCreateFailed`; introduces `errKymaBindingDescribeFailed`.
- Adds unit tests mirroring the subaccount `DeleteAPI5xxWrappedWithSpecifyAPIError` pattern (PR #716).

Part of the BTP error-body fan-out series.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -race -count=1 ./internal/clients/kymaenvironmentbinding/...`
- [x] `go test -race -count=1 ./...` — full suite green
- [x] `gofmt -l internal/clients/kymaenvironmentbinding/` clean


maximilianbraun (commented):
Lean. No over-engineering findings.

sdischer-sap (approved):

PR #736, opened 2026-06-20
fix(security): surface XSUAA error bodies via shared helper

maximilianbraun:
## Summary

Mirrors PR #716 (subaccount) for the three XSUAA security clients. Adds a single `SpecifyAPIError` helper at `internal/clients/security/errors.go` and routes every `.Execute()` error in:

- `internal/clients/security/rolecollection/xsuaa_rolecollection.go`
- `internal/clients/security/rolecollectiongroupassignment/xsuaa_group_assigner.go`
- `internal/clients/security/rolecollectionuserassignment/xsuaa_user_assigner.go`

through it. The helper type-asserts to `*xsuaa.GenericOpenAPIError` and surfaces the parsed model (or, failing that, the raw response body) so users see the actual XSUAA error payload in `Synced` conditions instead of a bare `500 Internal Server Error`.

## Why a separate helper

The XSUAA OpenAPI client lives in its own module (`internal/openapi_clients/btp-xsuaa-service-api-go/pkg`) with its own `GenericOpenAPIError` type, so the existing helper in `internal/controller/account/subaccount/subaccount.go` cannot be reused as-is. The new helper has the same shape but targets the XSUAA error type, with `map[string]any` as the model.

The existing 404 short-circuits in `rolecollection.Delete` and `userassignment.HasRole` are deliberately preserved.

## Test plan

- [x] `go build ./... && go vet ./...` clean
- [x] `go test -race -count=1 ./internal/clients/security/...` green (helper unit tests + one per-file integration test asserting the response body is surfaced)
- [x] `gofmt -l internal/clients/security/` clean

release-notes/bugfix


maximilianbraun (commented):
internal/clients/security/{rolecollection,rolecollectiongroupassignment,rolecollectionuserassignment}/*_test.go — The four-line `reflect.NewAt(...UnsafeAddr()).Elem().Set(reflect.ValueOf(body))` dance to set the unexported `body` field on `*xsuaa.GenericOpenAPIError` is now inlined 3× across the new sub-package tests, plus a 4th copy as `newGenericErr` in `errors_test.go`. Same shape PR #731 just consolidated for the account-side: drop these copies, add `internal/testutils/openapi_errors.go::NewXsuaaAPIError(body []byte) error`, call it from all four sites. Each test shrinks to one line and the regeneration-rot risk lives in one place.

maximilianbraun (commented):
Helper is minimal — single `body []byte` param, `nil` short-circuit matches the empty-returns-self case, the `IsValid()` guard handles SDK regeneration. All 4 inline copies collapsed to one-liners, no scope creep. Resolved.

sdischer-sap (approved):

PR #738, opened 2026-06-22
fix: deprecate KymaEnvironmentBindingId and remove broken reference resolver (backport to v1_10_x)

### LINKED ISSUE #727: [BUG] `KymaModule` fails creating due to not resolved `KymaEnvironmentBindingId`
**Describe the bug**
When creating a `KymaModule` it never starts creating the external resource, but rather gets stuck showing an event like: 

```
cannot resolve references: mg.Spec.KymaEnvironmentBindingId: referenced field was empty (referenced resource may not yet be ready)                                              
```
This came up during running long e2e tests, but should be the same for all `KymaModule` resources.

in dev on main as well as release v1.10.x branch

**To Reproduce**
1. Create `KymaModule` resource (and all required resources like `KymaEnvironmentBinding`
2. Check error events and see as described

**Expected behavior**
`KymaModule` gets healthy and is activated in the Kyma Environment.

**Additional context**
Checking the code I found that this Id is missing an extractor annotation: 

Also it looks like this ID is not used anywhere at all. 

So the easy fix should be to just remove this from the schema. Since this would cause an incompatibility it would probably better to just remove annotations and ignore (mark as deprecated). 

** Rollout Requirements **
Backport to v1.10.x branch (see #685)

sdischer-sap:
Backport of #737 to `release/v1_10_x`.

Fixes #727

### DIFF
--- a/apis/environment/v1alpha1/kymamodule_types.go
+++ b/apis/environment/v1alpha1/kymamodule_types.go
@@ -38,9 +38,8 @@ type KymaModuleParameters struct {
 type KymaModuleSpec struct {
 	xpv1.ResourceSpec `json:",inline"`
 	ForProvider       KymaModuleParameters `json:"forProvider"`
-	// +crossplane:generate:reference:type=github.com/sap/crossplane-provider-btp/apis/environment/v1alpha1.KymaEnvironmentBinding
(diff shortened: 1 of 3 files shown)


xdj2e9 (approved):
lgtm

PR #744, opened 2026-06-24
perf: cache BTP client per credential, dedupe + narrow entitlement reads

maximilianbraun:
## Summary

Cuts BTP API call volume by ~75% and response bandwidth by ~99% on the entitlement reconcile path, measured against a real cluster with ~40 Entitlement CRs (back-to-back runs, same instrumentation, baseline toggled via a local-only env var that's not in this PR).

| Metric | Before | After | Δ |
|---|---:|---:|---:|
| `POST /oauth/token` | 539 / 2 min | **2 / 2 min** | **−99.6%** |
| `GET /entitlements/v1/assignments` | 347 / 2 min | **218 / 2 min** | **−37%** |
| Response bytes received | ~390 MB / 2 min | **~5 MB / 2 min** | **−98.7%** |
| BTP wait time (sum of latencies) | ~181 s / 2 min | **~38 s / 2 min** | **−79%** |
| Avg GET latency | 278 ms | 172 ms | −38% |

## What changed

**1. Client cache per credential bundle** (`btp/cisclient.go`)
Process-wide `sync.Map` cache keyed by SHA-256 of the credential bundle. The providerconfig connector was building a fresh `btp.Client` per reconcile, throwing away the oauth2 token cache each time. The three sub-clients (accounts/entitlements/provisioning) now share one `oauth2.ReuseTokenSource` via a single `*http.Client`.

**2. Singleflight + 30 s TTL on `DescribeInstance`** (`internal/clients/entitlement/entitlement.go`)
Concurrent sibling reconciles share one in-flight `GetDirectoryAssignments`; back-to-back ones within the cache window share a result. Keyed by `(subaccount, service, plan)`. `UpdateInstance` invalidates its key on successful writes.

**3. Use `serviceName`+`planName` filter instead of `assignedServiceName`** (regen + `internal/clients/entitlement/entitlement.go`)
The new filter pair narrows BOTH `entitledServices` AND `assignedServices` server-side. Per-call response payload drops from ~1.18 MB to ~15-23 KB. The openapi client was regenerated from the latest BTP spec; the `ApiExceptionResponseObject` JSON patch op is changed from \`replace\`→\`add\` because upstream dropped that schema.

**Bonus: \`CountingRoundTripper\`** (\`btp/count_http_client.go\`)
Always-on instrumentation recording per-(method, host, bucketed-path) call count, latency, and req/resp bytes via atomic ops. Inspect with \`HTTPCallStats()\` / \`LogAndResetHTTPCallCounts(logger, label)\`. No automatic output — data is there for future observability hookups.

## Test plan

- [x] Full test suite passes (\`go test ./...\`)
- [x] Build clean (\`go build ./...\`)
- [x] Measured against a real cluster end-to-end (numbers above)
- [x] Token POST count stable at ~2 per ProviderConfig per session
- [x] Cache invalidation after writes verified by tracing
- [ ] Review by maintainers — esp. the openapi regen patch-op change and cache key composition


sdischer-sap:
@maximilianbraun lets add some documentation into the contribution docs about those new caching approaches and what APIs it does affect.

`docs/contribution_notes` seems to be a good place for it.

maximilianbraun:
> @maximilianbraun lets add some documentation into the contribution docs about those new caching approaches and what APIs it does affect.
> 
> `docs/contribution_notes` seems to be a good place for it.

Done, tried to keep it concise. Does that work for you?

sdischer-sap (approved):

PR #730, opened 2026-06-19
fix: skip description on SubaccountTrustConfiguration to avoid inconsistent-result error (backport to v1_10_x)

xdj2e9:
Backport of #725 into `release/v1_10_x`

Relates to #724

Issue #637, opened 2026-04-24
[FEATURE] Add disaster recover subaccount pairing as a resources


sdischer-sap:
**Context**
There is a new feature for disaster recovery in BTP. It allows you to pair a subaccount with another one. 
Looks like it got added to the BTP CLI and Terraform provider by now:
https://registry.terraform.io/providers/SAP/btp/latest/docs/resources/disaster_recovery_subaccount_pair

**Scope**
Add new Upjet based resources for `disaster_recovery_subaccount_pair`

**Technical Steps** 
- new resource available in provider
- one e2e test for the resource

**Additional context**
Link to terraform resource: https://registry.terraform.io/providers/SAP/btp/latest/docs/resources/disaster_recovery_subaccount_pair

antopraveen:
Team any ETA for this feature to be available

mpechkurov:
@antopraveen we are planning to do this in Q3
Let us know if you have any questions.

PR #750, opened 2026-06-25
feat: upgrade terraform-provider-btp v1.15.1 → v1.23.1 (#521, #724)

### LINKED ISSUE #521: Use v1.22 of btp terraform provider

xdj2e9:
Closes #521.

## Summary

This PR upgrades the bundled Terraform provider used by the fork-based client. Four changes are coupled together. All four are required to make the upgrade compile, generate, and reconcile correctly:

1. **`terraform-provider-btp` v1.15.1 → v1.23.1** (`Makefile`).
2. **`controller-tools` v0.20.0 → v0.20.1** (`go.mod`). This is needed so that `make generate` produces valid deepcopy code. In provider v1.23.1, `btp_subaccount_service_broker.password` became Optional. As a result, upjet now emits `*v1.SecretKeySelector`, and `controller-tools` v0.20.0 generated incorrect code for pointers to type aliases. The fix is in [controller-tools#1331](https://github.com/kubernetes-sigs/controller-tools/pull/1331).
3. **Identity injector** (`internal/clients/tfclient/identity_injector.go` plus the post-generate patcher in `cmd/zz-inject-identity/`). Starting with provider v1.19.0, every BTP managed resource declares `IdentitySchema()`. The `terraform-plugin-framework` then requires `resp.NewIdentity` to be set after every `Read`. Upjet's pinned `StateV4` struct has no `identity` field, so the on-disk `tfstate` never contains one. The first reconcile fails with `Missing Resource Identity After Read`. The injector wraps upjet's `controller.Store` and patches an `identity` block into `terraform.tfstate` between `WorkspaceStore.Workspace()` and the CLI subprocess read. The patched values come from a static identity table defined in the injector. The framework's post-Read check accepts any identity block, including one with placeholder values. See [terraform-provider-btp#1532](https://github.com/SAP/terraform-provider-btp/issues/1532) for more context.
4. **Terraform CLI 1.3.9 → OpenTofu 1.12.3** (`Makefile`, `cluster/images/provider-btp/Dockerfile`). The identity feature requires a CLI of version 1.12 or higher. Below that, the CLI removes the `identity` block before forwarding state to the provider, and the injector has no effect on the wire. OpenTofu (MPL-2.0) is installed as `/usr/local/bin/terraform` because upjet's executor expects that exact binary name.

## User-visible changes

The following changes are relevant for upgrade notes and release notes:

|---|---|---|
| `SubaccountServiceBroker.spec.forProvider` | `passwordSecretRef` and `username` are no longer Required. The CEL `XValidation` rules that required them have been removed. | The new mTLS path makes basic auth optional. Existing CRs that set both fields continue to validate. |
| `SubaccountServiceBroker.spec.forProvider` | New optional fields: `certSecretRef`, `keySecretRef`, `mtls` | These enable mTLS authentication against the broker. The cert and the key must always be supplied together. Set `mtls: true` to use credentials provided by Service Manager. In that case, cert and key must NOT be supplied. |
| `SubaccountServiceBroker.status.atProvider` | New `mtls` field that reflects the effective mTLS state | — |
| `SubaccountTrustConfiguration.spec.forProvider` | New optional `origin` field | This field was previously only present in the status. It is now also allowed in the spec. |
| `SubaccountTrustConfiguration.spec.forProvider.description` | The field is now persisted as set. | The workaround from #724, which silently set the field to nil, has been removed. The upstream fix is in provider v1.23.1. |
| `DirectoryEntitlement.planUniqueIdentifier` | Description clarified. | The new wording is: "...required only if you need to differentiate between identical plans that have different pricing." The schema itself is unchanged. |
| Bundled CLI | Terraform 1.3.9 → OpenTofu 1.12.3 | This is a drop-in replacement at the CLI level. Users who override `TERRAFORM_VERSION` or who ship custom Dockerfiles should review the Dockerfile changes for the plugin-mirror path layout. |

## Removal of the `SubaccountTrustConfiguration` `description` workaround

Related to #724.

The `description` skip that was introduced in #725 (and backported in #730) is removed in this PR. The upstream fix in the Terraform provider has landed in v1.23.1, so the workaround is no longer needed.

## Implementation details that may not be obvious

- **The wrap is applied only at `NewInternalTfConnector` (the hand-written upjet wrapper path).** `ServiceManager` reached Ready, but the template-generated controllers (`SubaccountTrustConfiguration` and others) still produced the identity error. The reason is that `UpjetOptions.WorkspaceStore` has a concrete type, because `NewWorkspaceFinalizer` requires it. The wrap therefore has to be injected at the `NewConnector` call site in each `zz_controller.go`. This is why we need a post-generate patcher.
- **Identity field names equal state attribute names, and empty values are not skipped.** For the trust configuration, the `origin` identity field is sourced from the state attribute `id`. The BTP provider treats them as the same logical identifier. On the first reconcile, `id` is empty, so the injector previously skipped exactly the case where the framework check fires. The fix has two parts: (1) the identity table now has the structure `map[tfType]map[identityField]stateAttr` with explicit aliases (for example `origin → id`), and (2) the identity block is always emitted, with placeholder values when needed.

## Decisions to confirm before merge

The following three points are the most important trade-offs in this PR. Please confirm them explicitly before we merge:

1. **The injector is a workaround.** The injector is functionally correct and has been validated end-to-end on both the template path and the hand-written wrapper path. However, it patches upjet's on-disk state from the outside instead of fixing the root cause (upjet's `StateV4` predates Terraform identity). The surface area is limited (about 140 lines for the injector, 130 lines for the patcher, plus one `//go:generate` directive in `apis/generate.go`). A removal plan is described below. Are we OK with shipping this as a temporary solution on `main`?
2. **Effort versus the upcoming no-fork migration.** Epic #207 will remove the fork-based client and the CLI entirely. At that point, the injector, the patcher, the OpenTofu install, and the mirror layout all become unused code (we estimate less than one hour of work to delete them). The benefit of merging now is that this PR unblocks #521, #724, #637, #480, and #282 today, instead of waiting for no-fork. Is this short-term benefit worth the temporary complexity? Note: upjet's identity support has landed in [PR #623](https://github.com/crossplane/upjet/pull/623) and is shipped in [v2.3.0](https://github.com/crossplane/upjet/releases/tag/v2.3.0), but it only covers the no-fork (TF Plugin Framework) path. The fork-based `StateV4` struct does not carry identity in any released version, and no upstream issue tracks adding it. The injector is therefore the only bridge on the fork-based client until the no-fork migration is complete.
3. **OpenTofu as the bundled CLI.** OpenTofu avoids the Terraform CLI 1.6 BSL license boundary and gives us identity support. It is a drop-in replacement at the CLI level for our workload (`init`, `refresh`, and `apply -refresh-only` have been validated). The risks are: (a) the maintainer and release cadence may diverge from HashiCorp, and (b) there may be behaviour differences at the boundary. Are we comfortable with taking the OpenTofu dependency, or should we revisit this (for example, by shipping native Terraform 1.12 and accepting the BSL license)? This alternative would need further investigation.

## Test plan

- [x] `make reviewable` — schema generation, `go generate`, `go mod tidy`, the identity injector patcher, `golangci-lint`, and unit tests.
- [x] `make test-acceptance` — full e2e sweep, including:
  - `Test_SubaccountTrustConfiguration_v1alpha1` (template path, with a non-empty `description` — this confirms that the #724 workaround can be removed).
  - `TestServiceManagerCreationFlow` (hand-written wrapper path via `NewInternalTfConnector`).
  - `DirectoryEntitlement`, `SubaccountServiceInstance`, `SubaccountServiceBinding`, `SubaccountServiceBroker`, `GlobalaccountTrustConfiguration`, and `SubaccountApiCredential` (the no-identity skip path).
- [x] Multi-reconcile loop — the injector is idempotent (`mutateState` returns `(raw, false)` when the state is already patched).
- [x] CLI-revert diagnostic — reverting `TERRAFORM_VERSION` to 1.3.9 while keeping everything else in place reproduces `Missing Resource Identity After Read` within seconds. This confirms that a CLI version of 1.12 or higher is required.
- [x] Upgrade tests.

## Removal plan

When no-fork lands, the following items can be removed (estimated effort: less than one hour):

- `internal/clients/tfclient/identity_injector.go` and its tests.
- `cmd/zz-inject-identity/` and its `//go:generate` directive in `apis/generate.go`.
- The `NewConnector` wrap in `internal/clients/tfclient/setup.go`.
- The OpenTofu install and the entire `provider-mirror` tree in the Dockerfile.
- The `--terraform-version` and `TERRAFORM_VERSION` plumbing (or this can be left in place for no-fork to ignore).


cmd/provider/main.go
```diff
@@ -131,6 +131,16 @@ func setupTerraformControllers(mgr manager.Manager, log logging.Logger, maxRecon
 			Provider: config.GetProvider(),
 			// use the following WorkspaceStoreOption to enable the shared gRPC mode
 			// terraform.WithProviderRunner(terraform.NewSharedProvider(log, os.Getenv("TERRAFORM_NATIVE_PROVIDER_PATH"), terraform.WithNativeProviderArgs("-debuggable")))
+			// NOTE(issue #521): identity-injection for template-generated upjet
+			// controllers (DirectoryEntitlement, SubaccountTrustConfiguration, …)
+			// is NOT applied here. Options.WorkspaceStore is typed
+			// *terraform.WorkspaceStore (concrete) — we cannot substitute our
+			// Store wrapper, and NewWorkspaceFinalizer also needs the concrete
+			// pointer. Resources whose lifecycle goes through this store still
+			// hit "Missing Resource Identity After Read" on refresh. Covered for
```
sdischer-sap: Are there any implications in practice? Didn't fully understand the PR descriptions. Will there any errors been shown to the user at any point? Or is this patching of the WorkspaceStore fixing that entirely?

test/e2e/testdata/crs/kyma_env/kyma.yaml
```diff
@@ -10,7 +10,7 @@ spec:
     parameters:
       region: westeurope
       administrators:
-        - $TECHNICAL_USER_EMAIL
+        - dl_673f35f54ffd090136dc821c@global.corp.sap
```
sdischer-sap: Needs to be removed!

Makefile
```diff
@@ -173,6 +179,20 @@ install-tools:
...
+# Patch upjet-generated zz_controller.go files to wrap o.WorkspaceStore in
+# tfclient.NewIdentityInjectingStore at the NewConnector call site. Workaround
+# for issue #521 — upjet's controller template has no public override hook, and
+# Options.WorkspaceStore must stay *terraform.WorkspaceStore (NewWorkspaceFinalizer
+# needs the concrete type). Idempotent. Remove when no-fork (PR #680 / issue
+# #207) lands. See cmd/zz-inject-identity/main.go.
+generate.done: inject-identity-into-zz-controllers
+
+.PHONY: inject-identity-into-zz-controllers
+inject-identity-into-zz-controllers:
```
sdischer-sap: Not super important, since it will be removed soon, but I think it might be a bit cleaner to run that generation as part of the `apis/generate.go`. We keep all generator scripts (related to the build itself) there.

sdischer-sap (changes requested):

cmd/provider/main.go
xdj2e9: this comment was outdated and already removed on my local branch. will push the commits asap

test/e2e/testdata/crs/kyma_env/kyma.yaml
xdj2e9: same as above. will be pushed asap

Makefile
xdj2e9: yes that looks like the way better place. will move it there.

cmd/provider/main.go
sdischer-sap: Great, but just to confirm regarding my question: 

> ServiceManager reached Ready, but template-generated controllers (SubaccountTrustConfiguration etc.) still hit the identity error — UpjetOptions.WorkspaceStore is typed concrete because NewWorkspaceFinalizer needs it, so the wrap has to be injected at the NewConnector call site in each zz_controller.go. Hence the post-generate patcher.

That part is a little hard to read. So there won't be any errors right? This only means that those inner changes affect workings of ServiceManager as well I guess.

xdj2e9: there won't be errors no. that comment represents an intermediate state of my journey (with claude 🙁 ).

xdj2e9:
Upgrade tests confirmed to be green except for https://github.com/SAP/crossplane-provider-btp/issues/760

sdischer-sap (approved):
Nice, looks good to me.

PR #751, opened 2026-06-25
fix: skip ResourceUsage tracking when MR is being deleted

schlingemann:
### What this fixes

When an MR is being deleted and one of its referenced upstream MRs has already been removed, `Connect()` aborts with:

> `connect failed: cannot track ResourceUsage: servicemanager.account.btp.sap.crossplane.io "sm-…" not found`

Because `Connect()` errors, `Delete()` is never called. The BTP-side resource stays alive, the Crossplane finalizer is never removed, and the MR is stuck on the cluster indefinitely. We hit this in canary with a `CloudManagement` MR after its `ServiceManager` was removed.

### The fix

Wrap each `resourcetracker.Track(ctx, mg)` call in `Connect()` with a `!meta.WasDeleted(mg)` guard. Applied to 11 controllers (`cloudmanagement`, `servicemanager`, `serviceinstance`, `servicebinding`, `subscription`, `rolecollection`, `rolecollectionassignment`, `kymamodule`, `kymaenvironmentbinding`, `environment/kyma`, `environment/cloudfoundry`).

### Why the guard is safe

`Track()` exists to *maintain* "we depend on X" `ResourceUsage` objects. During deletion we're tearing the dependency down, not strengthening it. Existing `ResourceUsage` records are owned by the MR with `BlockOwnerDeletion: false`, so they're garbage-collected by Kubernetes when the MR finally goes. The reverse direction ("we are depended upon by Y") is enforced by `DeleteShouldBeBlocked` reading the `InUse` condition (`internal/tracking/resourcetracker.go:303-308`), unchanged by this PR.

### Recovery for already-stuck CRs

Re-create the missing upstream MR with the existing external-name annotation so `Connect()` succeeds → `Delete()` fires → finalizer is removed.

### Tests

- Regression test `TestConnect_SkipTrackingWhenDeleted` in `cloudmanagement` reproducing the exact reported failure mode (Track set to return an error, MR with DeletionTimestamp → Connect must not propagate the error, and Track must not be called).
- `SkipTrackingWhenDeleted` cases added to existing `TestConnect_ResourceTracking` tables in `servicemanager` and `serviceinstance`.

### Note on generated files

Two of the patched files (`internal/controller/environment/kyma/zz_connect.go`, `internal/controller/kymaenvironmentbinding/zz_connect.go`) are upjet-generated. The same delete-loop bug occurs there, so the guard belongs in those files for now. A follow-up should push the pattern into the upjet template so it isn't lost on regeneration.

### Verification

- `go build ./...` clean
- `go vet ./...` clean
- `go test ./...` all green

### DIFF
--- a/internal/controller/account/cloudmanagement/cloudmanagement.go
+++ b/internal/controller/account/cloudmanagement/cloudmanagement.go
@@ -67,8 +67,14 @@ func (c *connector) Connect(ctx context.Context, mg resource.Managed) (managed.E
 		return nil, errors.Wrap(err, errTrackPCUsage)
 	}
 
-	if err := c.resourcetracker.Track(ctx, mg); err != nil {
-		return nil, errors.Wrap(err, errTrackRUsage)
+	// Skip ResourceUsage tracking when the MR is being deleted: Track() walks
+	// references and Gets each upstream MR; if any was deleted out from under
+	// us, that Get returns NotFound and Connect would abort before Delete()
+	// runs, leaving the BTP-side instance and the finalizer in place forever.
+	if !meta.WasDeleted(mg) {
+		if err := c.resourcetracker.Track(ctx, mg); err != nil {
+			return nil, errors.Wrap(err, errTrackRUsage)
+		}
 	}
 
 	if cr.Spec.ForProvider.ServiceManagerSecret == "" || cr.Spec.ForProvider.ServiceManagerSecretNamespace == "" {
--- a/internal/controller/account/cloudmanagement/cloudmanagement_test.go
+++ b/internal/controller/account/cloudmanagement/cloudmanagement_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/sap/crossplane-provider-btp/internal"
 	cmclient "github.com/sap/crossplane-provider-btp/internal/clients/cis"
 	"github.com/sap/crossplane-provider-btp/internal/clients/servicemanager"
+	"github.com/sap/crossplane-provider-btp/internal/testutils"
 	test2 "github.com/sap/crossplane-provider-btp/internal/tracking/test"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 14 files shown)


internal/controller/account/cloudmanagement/cloudmanagement.go
```diff
@@ -67,8 +67,14 @@ func (c *connector) Connect(ctx context.Context, mg resource.Managed) (managed.E
 		return nil, errors.Wrap(err, errTrackPCUsage)
 	}
 
-	if err := c.resourcetracker.Track(ctx, mg); err != nil {
-		return nil, errors.Wrap(err, errTrackRUsage)
+	// Skip ResourceUsage tracking when the MR is being deleted: Track() walks
+	// references and Gets each upstream MR; if any was deleted out from under
+	// us, that Get returns NotFound and Connect would abort before Delete()
+	// runs, leaving the BTP-side instance and the finalizer in place forever.
+	if !meta.WasDeleted(mg) {
```
maximilianbraun, line 74: The same guard ends up in 10 other controllers. Since every caller flows through `resourcetracker.Track()`, one `if meta.WasDeleted(mg) { return nil }` at the top of `Track()` in `internal/tracking/resourcetracker.go:65` would replace all 11 of these blocks and remove the divergence risk. Happy to sketch the diff in a follow-up commit if useful.

maximilianbraun (commented):
### Right diagnosis, consider moving the fix one layer down

The incident analysis is solid and the safety argument holds: `ResourceUsage` records are owned by the deleting MR with `BlockOwnerDeletion: false`, the `InUse` block reads from a separate condition path, and skipping `Track()` during deletion does not weaken either. So **the change works.**

What I'd push back on is *where* the guard lives. The same six lines land in 11 controllers, and two of them are upjet-generated (`zz_connect.go`) — which means the next `make generate` reverts the fix in those two unless someone also patches the template. The TODO is real and easy to forget.

All 11 call sites hit the same function: `resourcetracker.Track()`. Putting the guard there makes the property hold for every present and future caller, generated or not:

```go
// internal/tracking/resourcetracker.go:65
func (r *DefaultReferenceResolverTracker) Track(ctx context.Context, mg resource.Managed) error {
    if meta.WasDeleted(mg) { return nil }
    if hasIgnoreAnnotation(mg) { return nil }
    // … unchanged
}
```

Effect:
- 11 controller files revert to their pre-PR state.
- The two `zz_connect.go` cases are fixed for free; no upjet template follow-up.
- The three test additions collapse into one table case in `resourcetracker_test.go` (Track with a deleted MR → nil, no side effects).

Net diff vs. current: roughly **−214 / +1**.

The principle: a guard that says "skip this during deletion" is a property of the tracker, not of each caller. Pushing it down to the tracker keeps the call sites identical and makes the invariant a single place to read and reason about.

### Separate follow-up (out of scope for this PR)

Worth a tracking issue: today `tracker.SetConditions(ctx, mg)` only runs inside each upstream's `Delete()` (e.g. `servicemanager.go:184`). That means the `ResourceUsage` condition does not exist on a steady-state upstream — `DeleteShouldBeBlocked` returns `false` until the *first* delete attempt has already run. Moving `SetConditions` into each upstream's `Observe()` would arm the block before anyone tries to delete, which is the root cause of the scenario you describe in the PR body (upstream removed while downstream still depended on it).

Happy to take that one as a follow-up.

internal/controller/kymaenvironmentbinding/zz_connect.go
```diff
@@ -34,8 +35,14 @@ func (c *connector) Connect(ctx context.Context, mg resource.Managed) (managed.E
 		return nil, errors.Wrap(err, errTrackPCUsage)
 	}
 
-	if err := c.resourcetracker.Track(ctx, mg); err != nil {
-		return nil, errors.Wrap(err, errTrackRUsage)
+	// Skip ResourceUsage tracking when the MR is being deleted: Track() walks
+	// references and Gets each upstream MR; if any was deleted out from under
+	// us, that Get returns NotFound and Connect would abort before Delete()
+	// runs, leaving the BTP-side binding and the finalizer in place forever.
+	if !meta.WasDeleted(mg) {
```
maximilianbraun, line 42: Heads up: this file is generated by upjet. The TODO in the PR body about pushing the guard into the template is real — without it, the next `make generate` silently removes the fix here. If the guard moves into `Track()` itself, this file does not need to be touched at all, and the template TODO goes away.

internal/controller/account/cloudmanagement/cloudmanagement_test.go
```diff
@@ -271,6 +272,60 @@ func TestConnect(t *testing.T) {
 	}
 }
 
+// TestConnect_SkipTrackingWhenDeleted is a regression test for the case where
+// a CloudManagement MR is being deleted and its referenced ServiceManager MR
+// has already been removed from the cluster. Before the fix, Connect() would
+// abort with "cannot track ResourceUsage: servicemanager … not found", which
+// kept Delete() from ever running — leaving the BTP-side CIS instance and the
+// finalizer in place forever.
+func TestConnect_SkipTrackingWhenDeleted(t *testing.T) {
```
maximilianbraun, line 281: Two small things if this stays:

1. The assertion `errors.Cause(err).Error() == "servicemanager … not found"` matches a hardcoded string from the test fixture, not real-world error text. The table-style cases you added in `servicemanager_test.go` and `serviceinstance_test.go` assert `err == nil && trackCalled == false`, which is what we actually care about and survives refactors. Consider folding this into the same table form here for consistency.

2. If the guard moves into `Track()`, one test in `internal/tracking/resourcetracker_test.go` covers all 11 controllers, and these three controller-level tests can go away.

sdischer-sap:
Isn't that generally hiding other potential errors. In what scenario can would the referenced resources not be found? 
Isn't it the point of the resourceUsage, that those references are ensured?

sdischer-sap:
@schlingemann this was probably already addressed with https://github.com/SAP/crossplane-provider-btp/pull/771 right? I guess we can close it?

PR #752, opened 2026-06-25
fix(servicemanager): handle numeric metadata in service plan responses

### LINKED ISSUE #747: [BUG]  ServiceInstance plan resolution fails when broker returns numeric supportedMaxOSBVersion in plan metadata
## Description

ServiceInstance creation fails for `feature-flags` with:

```
connect failed: while initializing service plan: cannot resolve plan ID:
json: cannot unmarshal number into Go struct field
ServicePlanMetadata.items.metadata.supportedMaxOSBVersion of type string
```

## Root Cause

The `feature-flags` broker returns `supportedMaxOSBVersion` as a JSON number (`2.14`) instead of a string (`"2.14"`) in the SM API response at `GET /v1/service_plans`:

```json
{
  "name": "standard",
  "metadata": {
    "supportedMaxOSBVersion": 2.14,
    "supportedMinOSBVersion": 2.13
  }
}
```

The provider's Go struct types this field as `string`, so `json.Unmarshal` fails during plan ID resolution.

Other services (xsuaa, identity) return the same field as a string. Only the feature-flags broker serializes it as a number. Checked against `service-manager.cfapps.ap21.hana.ondemand.com`.

## Expected Behavior

Plan resolution should succeed regardless of whether the broker returns the field as a string or number. The metadata struct should use `interface{}` or `json.Number` for fields that brokers may serialize inconsistently.

## Affected

- provider-btp v1.9.0
- Service: `feature-flags` (both `lite` and `standard` plans)
- Region: verified in ap21, likely all regions

## Reproduction

1. Entitle `feature-flags` / `standard` in a subaccount
2. Create a ServiceInstance CR targeting `feature-flags` with `planName: standard`
3. Observe ReconcileError on the Synced condition

jtucci:
## Summary

Patches the Service Manager OpenAPI spec to treat `ServicePlanResponseObject.metadata` as an untyped object (`map[string]interface{}`) instead of a strictly-typed struct. Some brokers (e.g. `feature-flags`) return `supportedMaxOSBVersion` as a JSON number rather than a string, causing plan resolution to fail on unmarshal.

Fixes #747


jtucci:
Closing in favor of a new PR from the origin repo (not the fork).

PR #756, opened 2026-06-25
fix: enforce BTP label limits on Subaccount spec

### LINKED ISSUE #740: [BUG] `Subaccount` label value max length not reflected in CRD
**Describe the bug**
When setting a `label` in the `Subaccount` CRD there is a length limit of 63 characters. However, the CRD allows more resulting in reconciliation errors due to a `400 Bad Request` to the BTP API

v1.9.0

**To Reproduce**
add a label to a subaccount with a value of >63 character

**Expected behavior**
CRD should have been rejected/dry run should have failed

**Screenshots**

**Additional context**

maximilianbraun:
## Summary

`Subaccount.spec.forProvider.labels` is documented in the Go type as "up to 10 user-defined labels", "up to 10 values per label", and "Keys and values are each limited to 63 characters" (see `apis/account/v1alpha1/subaccount_types.go:28-31`), but **none of those limits were translated into CRD validation**. Anything malformed sailed through admission and failed mid-reconcile with `400 Bad Request` from BTP, which is exactly what #740 reports.

This PR adds all four limits at admission time. The first attempt expressed everything as CEL `XValidation`, which is the style already in use across this provider (e.g. `apis/security/v1alpha1/rolecollectionassignment_types.go:13`, `apis/account/v1beta1/cloudmanagement_types.go:51`). That approach failed CRD install with:

> spec.validation.openAPIV3Schema.properties[spec].properties[forProvider].properties[labels].x-kubernetes-validations[2].rule: Forbidden: **estimated rule cost exceeds budget by factor of 6.3x** (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)

The CEL cost analyzer can't compute a tight cost bound when the inner array length and inner string length are unconstrained in the schema, even if the rule itself does a size check. The fix is to push the array and string bounds into native OpenAPI properties, which has the side benefit of replacing the costliest CEL rule with the equivalent native validation.

## What changed

**New named type for the label value list** (`apis/account/v1alpha1/subaccount_types.go`):

```go
// SubaccountLabelValueList is a list of values for one Subaccount label key.
// Bounded so the CEL value-length rule on Subaccount.spec.forProvider.labels stays inside the API server's per-rule cost budget.
// +kubebuilder:validation:MaxItems=10
// +kubebuilder:validation:items:MaxLength=63
type SubaccountLabelValueList []string
```

**`Labels` field on `SubaccountParameters`** now uses that type, plus two field-level markers:

```go
// +kubebuilder:validation:MaxProperties=10
// +kubebuilder:validation:XValidation:rule="self.all(k, size(k) <= 63)",message="label keys must be at most 63 characters"
Labels map[string]SubaccountLabelValueList `json:"labels,omitempty"`
```

**`package/crds/account.btp.sap.crossplane.io_subaccounts.yaml`** — regenerated. Gains `maxProperties: 10` on the labels map, `maxItems: 10` on each value array, and `maxLength: 63` on each value string, plus one CEL rule for key length (OpenAPI has no native marker for map keys). No changes to the status-side `atProvider.labels` — no point validating what the controller writes.

**Call-site updates.** `map[string][]string` and `map[string]SubaccountLabelValueList` are not implicitly convertible in Go even though the underlying slice type is the same. Three call sites that passed plain `map[string][]string` to fields or helpers that now expect the new map type were updated with explicit copies / signature changes:
- `cmd/exporter/resources/subaccount/convert.go`: explicit copy when assigning to `Subaccount.Spec.ForProvider.Labels`.
- `internal/controller/account/subaccount/subaccount.go`: `changedLabels` signature now takes the new map type and compares per-key via `reflect.DeepEqual([]string(v), …)`; `addOperatorLabel` builds its `map[string][]string` result with an explicit copy.
- Test files (`convert_test.go`, `subaccount_test.go`) updated to use the new map type in struct literals.

## What each rule catches

| Manifest mistake | Reported as | Where |
|---|---|---|
| 11+ keys | `spec.forProvider.labels: Too many: must have at most 10 items` | `MaxProperties: 10` (native) |
| Key with 64+ chars | `label keys must be at most 63 characters` | CEL `self.all(k, size(k) <= 63)` |
| 11+ values for a key | `spec.forProvider.labels[<key>]: Too many: must have at most 10 items` | `MaxItems: 10` on `SubaccountLabelValueList` (native) |
| Value with 64+ chars | `spec.forProvider.labels[<key>][<i>]: Too long: may not be more than 63 bytes` | `items.MaxLength: 63` on `SubaccountLabelValueList` (native) |

All surface as standard `kubectl apply` errors / `Synced=False` admission rejections instead of opaque BTP 400s mid-reconcile. The native messages are slightly better than the CEL form (they name the offending index/key), which is a bonus from the redesign.

## Notes

- `size()` in CEL and `MaxLength` in OpenAPI both count UTF-8 bytes, which is how BTP measures the limit too. Same byte count on both sides.
- Existing `Subaccount` resources with malformed labels would start failing admission on update. That's a feature, not a regression — they were silently broken before. Worth a one-liner in the release notes.
- Limits sourced from the existing Go doc comment, which matches the BTP accounts-service API contract (`CreateSubaccountRequestPayload.properties.labels` and `UpdateSubaccountRequestPayload.properties.labels` at `accounts-service.cfapps.<region>.hana.ondemand.com/v3/api-docs`). The Go doc-comment now cites the source inline. If BTP later relaxes these, the markers move with the doc.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./apis/account/v1alpha1/... ./cmd/exporter/resources/subaccount/... ./internal/controller/account/subaccount/...` green
- [x] CRD regenerated with `controller-gen v0.20.0` (same version the repo uses)
- [x] CRD passes `kubectl-validate` (CEL cost budget check included) — first attempt failed this check, current one passes
- [ ] Maintainer review

Fixes #740.


maximilianbraun:
Confirmed the limits against the live source. Pulled \`https://accounts-service.cfapps.eu10.hana.ondemand.com/v3/api-docs\` and the relevant schemas are \`CreateSubaccountRequestPayload.properties.labels\` and \`UpdateSubaccountRequestPayload.properties.labels\`. Both carry the same constraint verbatim:

> JSON array of up to 10 user-defined labels to assign as key-value pairs to the subaccount. Each label has a name (key) that you specify, and to which you can assign up to **10 corresponding values or leave empty**. **Keys and values are each limited to 63 characters.**

So 10 / 10 / 63 / 63 is authoritative on both create and update. The CEL rules in this PR encode it exactly.

| Constraint | Source phrase | Rule |
|---|---|---|
| \`\`MaxProperties: 10\`\` | "up to 10 user-defined labels" | native marker |
| ≤10 values per key | "up to 10 corresponding values" | \`self.all(k, size(self[k]) <= 10)\` |
| Empty list allowed | "or leave empty" | implicit (empty list trivially satisfies the size rules) |
| Keys ≤63 chars | "Keys and values are each limited to 63 characters" | \`self.all(k, size(k) <= 63)\` |
| Values ≤63 chars | same | \`self.all(k, self[k].all(v, size(v) <= 63))\` |

One side-observation: the BTP OpenAPI documents these limits only in the human-readable \`description\` text, not as \`maxLength\` / \`maxItems\` / \`maxProperties\` JSON Schema fields. That's why client tooling never picked them up automatically. Worth a friendly nudge to the accounts-service team to formalize them in the schema — once that happens, regenerated clients pick up the validation for free.

sdischer-sap (approved):

PR #762, opened 2026-06-26
fix(cfenvironment): skip tech user when assigning initial org managers

### LINKED ISSUE #758: [BUG] Failing `CloudFoundryEnvironment` creation if technical user is part of `initialOrgManagers`
**Describe the bug**
This issue was created based off an internal support ticket:
We learned, that there is an issue when setting the technical user as part of the `initialOrgManager` list of the cf environment. 

This leads to the provider attempting to assign manager rights, which fails, because the creating user is already manager by default. Since all of this happens as part of the Creation it fails the overall process and the external-name is not yet set, which in turn leads to collisions when attempting to create in the next reconcilations. 

**Error shown in CloudFoundryEnvironment**
> while creating instance: could not create CloudFoundryEnvironment: {"error":{"code":41001,"message":"Cannot create the environment instance on the specified landscape [cf-ae │
│ 1[]. Reason: there is already an environment instance of the same type on that landscape.","target":"/provisioning/v1/environments","correlationID":"ab3a383e-9a87-4e63-4aa0-2cfaf3b96624"}}: 409 Conflict

v1.9.0

**To Reproduce**
1. Set User A as Technical User credential in a secret and reference it in ProviderConfig A
2. Create a `CFEnvironment` with reference to ProviderConfig A and User A as part of the `initialOrgManager` list
3. See the creation failing eventually ending in the state where collisions are detected in continous creation attempts (see error above)

**Expected behavior**
It should not matter wether or not the techuser is part of the `initialOrgManager`. Whoever is set in the initialOrgManager should have manager rights. Technical details need to be hidden from the user. 

**Suggested solution**
- filtering the user out when assigning manager permissions
- make it more clear in the documentation of `initialOrgManager` CRD that the technical user will be defined as manager by default

**Additional context**
A workaround of course is not set the test user in the `initialOrgManager` list.

lennardrother:
Do not add the user referenced in the `ProviderConfig` as Org Manager since it's already added when creating the `CloudFoundryEnvironment` (the creating user is added by default).

An alternative would be a filter function which I could test separately.

### DIFF
--- a/internal/clients/cfenvironment/cfenvironment.go
+++ b/internal/clients/cfenvironment/cfenvironment.go
@@ -162,6 +162,10 @@ func (c CloudFoundryOrganization) CreateInstance(ctx context.Context, cr v1alpha
 	}
 
 	for _, managerEmail := range cr.Spec.ForProvider.Managers {
+		if managerEmail == adminServiceAccountEmail {
+			// Skip adding the admin service account email as org manager since it's added by default during creation
+			continue
+		}
 		if err := cloudFoundryClient.addManager(ctx, managerEmail, defaultOrigin); err != nil {
 			return "", errors.Wrap(err, instanceCreateFailed)
 		}
--- a/internal/clients/cfenvironment/cfenvironment.go
+++ b/internal/clients/cfenvironment/cfenvironment.go
@@ -161,11 +161,7 @@ func (c CloudFoundryOrganization) CreateInstance(ctx context.Context, cr v1alpha
 		return "", errors.Wrap(err, instanceCreateFailed)
 	}
 
-	for _, managerEmail := range cr.Spec.ForProvider.Managers {
-		if managerEmail == adminServiceAccountEmail {
-			// Skip adding the admin service account email as org manager since it's added by default during creation
-			continue
-		}
+	for _, managerEmail := range filterOutUser(cr.Spec.ForProvider.Managers, adminServiceAccountEmail) {
 		if err := cloudFoundryClient.addManager(ctx, managerEmail, defaultOrigin); err != nil {
 			return "", errors.Wrap(err, instanceCreateFailed)
 		}
@@ -187,6 +183,16 @@ func FormOrgName(orgName string, subaccountId string, crName string) string {
 		return subaccountId + "-" + crName
 	}
 	return orgName
+}
+
+func filterOutUser(users []string, exclude string) []string {
+	result := make([]string, 0, len(users))
+	for _, u := range users {
+		if u != exclude {
+			result = append(result, u)
+		}
+	}
+	return result
 }
 
 type organizationClient struct {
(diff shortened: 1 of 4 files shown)


internal/clients/cfenvironment/cfenvironment.go
```diff
@@ -162,6 +162,10 @@ func (c CloudFoundryOrganization) CreateInstance(ctx context.Context, cr v1alpha
 	}
 
 	for _, managerEmail := range cr.Spec.ForProvider.Managers {
+		if managerEmail == adminServiceAccountEmail {
```
xdj2e9: The added behavior should be unit tested.

xdj2e9 (commented):
Is this behaviour for the CF instance creation documented somewhere? If not, I think it would be the safer implementation to have a check if the email is already set as manager and only add it if missing.

Otherwise if the undocumented behavior is changed our provider will end up not setting the managers as requested because we would still have the implicit expectation of it being added automatically.

lennardrother:
> Is this behaviour for the CF instance creation documented somewhere? If not, I think it would be the safer implementation to have a check if the email is already set as manager and only add it if missing.
> 
> Otherwise if the undocumented behavior is changed our provider will end up not setting the managers as requested because we would still have the implicit expectation of it being added automatically.

Yes, it's documented [here](https://help.sap.com/docs/btp/sap-business-technology-platform/about-roles-in-cloud-foundry-environment) and [here](https://help.sap.com/docs/btp/sap-business-technology-platform/add-org-members-using-cockpit). We could of course go the safe route regardless. What do you think?

xdj2e9:
> > Is this behaviour for the CF instance creation documented somewhere? If not, I think it would be the safer implementation to have a check if the email is already set as manager and only add it if missing.
> > Otherwise if the undocumented behavior is changed our provider will end up not setting the managers as requested because we would still have the implicit expectation of it being added automatically.
> 
> Yes, it's documented [here](https://help.sap.com/docs/btp/sap-business-technology-platform/about-roles-in-cloud-foundry-environment) and [here](https://help.sap.com/docs/btp/sap-business-technology-platform/add-org-members-using-cockpit). We could of course go the safe route regardless. What do you think?

I think in that case the implementation is perfectly fine. Maybe we can add a comment with these links to document why we behave like that?

sdischer-sap (commented):
Maybe linking the docs in the CRD comment as well (if its public). 

I think it should be safe for us to assume that the techuser that created the resource has access on that resource. If it wouldn't we also couldn't add them as manager. So I don't think we need that check.

internal/clients/cfenvironment/cfenvironment.go
lennardrother: Done with https://github.com/SAP/crossplane-provider-btp/pull/762/commits/5f62092341d2eeaeb794fa20a374fb16af4b2d55, extracted into a function for simple testing

xdj2e9 (dismissed):
LGTM

xdj2e9 (approved):

PR #770, opened 2026-06-29
fix(entitlement): retry and surface PROCESSING_FAILED with no assigned quota

### LINKED ISSUE #391: [BUG] Wrong State and Missing Event for failing Hana Entitlement
**Describe the bug**
The hana entitlement was shown as Ready, even though it was not created, because of quota issues. There was no Event shown in the resource. The error message was only visible in `Status.At Provider`. I would have expected the state 'Failed' with an event.

ghcr.io/sap/crossplane-provider-btp/crossplane/provider-btp:v1.3.0

**To Reproduce**
1. Create a entitlement `hana-cloud` with plan `hana`that hits the quota
2. Check the Status of the entitlement resource

**Expected behavior**
The entitlement should have an error state, when it can not be created in BTP.

**Screenshots**
Image

Image

**Additional context**
-

### LINKED ISSUE #769: [BUG] Entitlement reports Ready=Available when BTP assign actually failed (assigned.entityState=PROCESSING_FAILED, amount=0)
**Describe the bug**

The `Entitlement` controller maps `status.atProvider.assigned.entityState == PROCESSING_FAILED` to the Crossplane `Ready=Available` condition unconditionally, even when the assignment never succeeded (i.e. `assigned.amount == 0` and the BTP `stateMessage` indicates a hard assign failure such as *"Failed to call Provisioning service to assign quota."*).

As a result, an entitlement that is **not actually provisioned in BTP** reports as healthy:

```
NAME                READY   SYNCED   VALIDATION                EXTERNAL-NAME           AGE
aicore-<sub>        True    True     NoValidationIssuesFound   aicore-<sub>            30h
malware-scanner-<sub>  True True     NoValidationIssuesFound   malware-scanner-<sub>   30h
```

while `kubectl get -o yaml` shows:

```yaml
status:
  atProvider:
    assigned:
      amount: 0
      entityState: PROCESSING_FAILED
      stateMessage: "Failed to call Provisioning service to assign quota."
  conditions:
  - type: Ready
    status: "True"
    reason: Available
  - type: Synced
    status: "True"
    reason: ReconcileSuccess
```

This makes any orchestration / composition that gates on `Ready=True` think the entitlement is good, but downstream `ServiceInstance` reconciles immediately fail with:

```
connect failed: while initializing service plan:
  cannot resolve plan ID: no service offering found for '<service>'
```

because the catalog never received the assign.

There are actually two compounding defects in the controller:

1. **Status conflation.** `internal/controller/account/entitlement/entitlement.go` lines 139-156:
   ```go
   switch cr.Status.AtProvider.Assigned.EntityState { //nolint:exhaustive
   case apisv1alpha1.EntitlementStatusOk:
       cr.Status.SetConditions(xpv1.Available())
   // ... comment justifying this for delete-time failures ...
   case apisv1alpha1.EntitlementStatusProcessingFailed:
       cr.Status.SetConditions(xpv1.Available())          // ← always reports Available
   ...
   ```
   The accompanying comment makes clear the intent was to avoid flapping on a *delete-time* `PROCESSING_FAILED` against an entitlement that BTP still considers valid (e.g. the "quota cannot be lower than consumed" pattern). That intent is reasonable, but the implementation also covers the *assign-time* `PROCESSING_FAILED` case where there is no live assignment at all (`amount=0`, requested but never reserved). Both cases are reported identically as `Ready=Available`.

2. **No retry path for enable-style entitlements.** `internal/controller/account/entitlement/entitlement.go` lines 289-309:
   ```go
   func (c *external) needsUpdate(cr *apisv1alpha1.Entitlement) bool {
       autoAssign := cr.Status.AtProvider.Assigned.AutoAssign
       if autoAssign { return false }
       if unlimitedAmountAssigned { return false }
       if cr.Spec.ForProvider.Amount != nil {                              // amount-style only
           return !reflect.DeepEqual(cr.Status.AtProvider.Required.Amount,
                                     cr.Status.AtProvider.Assigned.Amount)
       }
       return false                                                        // enable-style: never
   }

   func (c *external) needsCreate(cr *apisv1alpha1.Entitlement) bool {
       return cr.Status.AtProvider.Assigned == nil                          // Assigned != nil here
   }
   ```
   For enable-style entitlements (e.g. `aicore/extended`, `malware-scanner/clamav` — `category=ELASTIC_SERVICE`, `provisioningMethod=SERVICE_BROKER`, spec `enable: true`, no `amount`), after the first failed assign:
   - `Assigned` is non-nil (it carries the failure record), so `needsCreate` returns `false`.
   - `Spec.ForProvider.Amount` is `nil`, so `needsUpdate` returns `false`.
   - The switch hits `PROCESSING_FAILED → Available`.
   - `Observe` returns `ResourceExists=true, ResourceUpToDate=true` → Crossplane never calls `Create` or `Update` again.

   So a transient outage of the BTP Provisioning service at the very first assign turns the CR into a permanent "green but broken" state with no self-healing.

This appears to be the same underlying issue reported by @LenaHoinkis in #391 (closed for lack of reporter follow-up). The defect is still in current `main` — `jj log -r '::main' internal/controller/account/entitlement/entitlement.go` shows the last touch to that file is #653 (upjet v2 upgrade), which doesn't affect this code path. The `PROCESSING_FAILED → Available` mapping and the missing enable-style retry are both unchanged.

Reproduced on `ghcr.io/sap/crossplane-provider-btp/crossplane/provider-btp:v0.0.0-50c457265f27d5f6bab1fe063f7be4c9ba5a003e` (commit `50c4572` on `main`, between `v2.0.0-rc1` and HEAD).

Code citations above are against current `main` (`851d246a`).

**To Reproduce**

1. Apply two enable-style `Entitlement` CRs in the same subaccount, e.g. `aicore/extended` and `malware-scanner/clamav` (`spec.forProvider.enable: true`, no `amount`).
2. Cause the very first BTP assign call to fail. The simplest path is a transient outage of the Provisioning service or any condition under which the Entitlements service responds with the `PROCESSING_FAILED` stateMessage `Failed to call Provisioning service to assign quota.`. (We have seen this happen unprompted in our shared landscape.)
3. After the transient condition passes, observe the `Entitlement` CR:
   - `kubectl get entitlement <name>` → `READY=True, SYNCED=True`.
   - `kubectl get entitlement <name> -o yaml` → `status.atProvider.assigned.entityState=PROCESSING_FAILED`, `amount=0`, `stateMessage="Failed to call Provisioning service to assign quota."`.
   - Quota on BTP side stays at 0 (`btp list accounts/entitlement --subaccount <guid>` shows no row for that service/plan).
4. Try to create a matching `ServiceInstance` (e.g. `aicore/extended`). The SI reconciles fail with `cannot resolve plan ID: no service offering found for '<service>'` because the catalog never received the assignment.
5. Wait an arbitrary amount of time. The Entitlement controller never retries; the CR remains `Ready=True` forever.

**Expected behavior**

- An entitlement whose assigned state is `PROCESSING_FAILED` and whose live assigned amount is `0` (i.e. nothing actually reserved on BTP) should not report `Ready=Available`. It should surface as `Ready=Unavailable` (or `Synced=False`) with the BTP `stateMessage` propagated as the condition message, and an event should be emitted.
- The controller should retry the assign for enable-style entitlements in this state, not stop forever. Either `needsCreate` should treat "`Assigned.EntityState=PROCESSING_FAILED && Assigned.Amount=0` (and amount was never positive)" as "still needs to be created", or `needsUpdate` should grow a clause for the enable-style path (`Spec.Enable != nil && Assigned.Amount == 0 && EntityState == PROCESSING_FAILED`) returning `true` so `Update` reissues the assign.
- The legitimate concern in the existing comment (don't flap on a *delete-time* `PROCESSING_FAILED` against an entitlement that BTP still has assigned) should be preserved by gating the `Available` mapping on `assigned.amount > 0` (or, for enable-style, on the entitlement actually being present in the Entitlements/Provisioning catalog).

**Additional context**

- Related, previously closed issue: #391 — same symptom shape, closed because the reporter didn't respond to a reproduction request. The code path is unchanged since.
- This is independent of, but stacks badly with, a separate deletion-leak issue we've been investigating: when this state happens during a teardown, the never-actually-assigned entitlement still has an `assigned` record, so `needsCreate=false`, and the subaccount/entitlement teardown logic can't reason about it correctly either.

dattito:
**What this fixes**

Fixes #769 (closely related: previously-closed #391).

When the BTP Entitlements / Provisioning service rejects the initial assign call, the `Entitlement` controller currently maps the resulting `assigned.entityState=PROCESSING_FAILED` with `amount=0` (or `nil`) to `Ready=Available` and never retries. Downstream resources gating on `Ready=True` (compositions, `ServiceInstance`) treat the entitlement as healthy when nothing is actually reserved on BTP. `ServiceInstance` reconciles then fail with `connect failed: while initializing service plan: cannot resolve plan ID: no service offering found for '<service>'` because the catalog never received the assignment.

**What this changes**

* **`needsCreate`** is broadened to also return `true` when `assigned.entityState=PROCESSING_FAILED && assigned.amount<=0` (or `nil`). The managed reconciler then drives `Create` → `CreateInstance` (which is `UpdateInstance` in `internal/clients/entitlement/entitlement.go`), retrying the assign on every poll. `Ready` is now honestly `Creating` instead of `Available`.
* The existing tolerance for delete- or update-time `PROCESSING_FAILED` with `amount > 0` is preserved (that was the intent of the original comment on the switch case — failing a *change* on a still-assigned entitlement should not flap dependents). The new `assignFailedNoQuota` helper makes the discriminator explicit.
* A defensive `Unavailable` mapping is added inside the existing `switch` `PROCESSING_FAILED` branch, so the green-but-broken state is unreachable even if the early `needsCreate` short-circuit is bypassed in a future refactor.

**Discriminator**

| Field | Assign-time failure (this fix) | Delete-/update-time failure (existing tolerated case) |
|---|---|---|
| `Assigned.EntityState` | `PROCESSING_FAILED` | `PROCESSING_FAILED` |
| `Assigned.Amount` (`*int`) | `nil` or `*amount == 0` | `> 0` |
| What's actually in BTP | nothing reserved; catalog has no offering | entitlement is still assigned; just the last change op failed |

Negative checks:
* Numeric quota with `EntityState=OK, Amount=0` (someone disabled the entitlement): unaffected, `EntityState != PROCESSING_FAILED`.
* Enable-style with `Enable=false, EntityState=OK`: unaffected for the same reason.
* Brand-new entitlement before first BTP response (`Assigned == nil`): already caught by the existing `needsCreate`.

**Tests**

Three new cases under `TestObserve`:
1. `Assign-time PROCESSING_FAILED with amount=0 (enable-style) -- retries via Create` — asserts `ResourceExists=false` so the managed reconciler retries.
2. `Assign-time PROCESSING_FAILED with nil amount (enable-style) -- retries via Create` — same, with `nil` instead of `0`.
3. `Assign-time PROCESSING_FAILED with amount=0 -- Ready must not be Available` — asserts the `Ready` condition is not set to `Available` for this shape.

The existing test `"Simple Case, All up-to-date, unavailable condition"` (which asserts `amount=1, PROCESSING_FAILED → Ready=Available`) is preserved and still passes — it's the delete-/update-time tolerance case, and the new helper specifically excludes it.

```
$ go test ./internal/controller/account/entitlement/... -run TestObserve -count=1
ok    github.com/sap/crossplane-provider-btp/internal/controller/account/entitlement   0.679s
```

`go vet ./...` clean, `go build ./...` clean.

**Diff is confined to two files**

* `internal/controller/account/entitlement/entitlement.go` (helper + `needsCreate` broaden + defensive switch update)
* `internal/controller/account/entitlement/entitlement_test.go` (three new cases)

No CRD / API / generated-code changes.

**Out of scope**

* Surfacing the BTP `stateMessage` directly on the `Ready` condition. Crossplane's managed reconciler overwrites `Ready` to `Creating` after our Observe returns, so the message would need either an event recorder injected into the `external` client (precedent exists in other controllers in this repo) or a custom condition type. Happy to follow up with that in a separate PR if you'd like.
* The deletion-time leak path observed at the same time (cascading `ServiceBinding`/`ServiceInstance`/`ServiceManager` leak when upjet `Observe` mis-reports `ResourceExists=false`). Separate root cause, separate fix.

### DIFF
--- a/internal/controller/account/entitlement/entitlement.go
+++ b/internal/controller/account/entitlement/entitlement.go
@@ -139,14 +139,28 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 	switch cr.Status.AtProvider.Assigned.EntityState { //nolint:exhaustive
 	case apisv1alpha1.EntitlementStatusOk:
 		cr.Status.SetConditions(xpv1.Available())
-	// the state relates to the last operation, not the current state
-	// at this point we already verified that the entity does match the desired state for all we know
-	// so we will ignore this state to avoid blocking healthy resources
-	// example case: attempting to delete an entitlement which is already consumed will fail in that way
-	// -> the only resolution is to discard and recreate the cr, in this case we just need to observe the still existing entitlement
-	// despite the last operation being a failure
+	// PROCESSING_FAILED reflects the *last operation* on the entitlement,
+	// not whether something is currently assigned. Two distinct shapes hit
+	// this branch:
+	//   1. Delete- or update-time failure on a still-assigned entitlement
+	//      (amount > 0). The entitlement is still in use; reporting
+	//      Available avoids flapping orchestration that depends on this CR.
+	//      Example: attempting to delete an entitlement that is already
+	//      consumed will fail in this way; the only resolution is to
+	//      discard and recreate the CR, in this case we just need to
+	//      observe the still existing entitlement despite the last operation
+	//      being a failure.
+	//   2. Assign-time failure with nothing reserved (amount == 0 / nil).
+	//      This is normally short-circuited above by needsCreate() so the
+	//      managed reconciler can retry via Create. The branch below is a
+	//      defensive fallback in case any future change bypasses that
+	//      short-circuit: reporting Available here would be a lie.
 	case apisv1alpha1.EntitlementStatusProcessingFailed:
-		cr.Status.SetConditions(xpv1.Available())
+		if c.assignFailedNoQuota(cr) {
+			cr.Status.SetConditions(xpv1.Unavailable())
+		} else {
+			cr.Status.SetConditions(xpv1.Available())
+		}
 	case apisv1alpha1.EntitlementStatusProcessing:
 		cr.Status.SetConditions(xpv1.Creating())
 	case apisv1alpha1.EntitlementStatusStarted:
@@ -304,8 +318,33 @@ func (c *external) needsUpdate(cr *apisv1alpha1.Entitlement) bool {
 	return false
 }
 
+// assignFailedNoQuota returns true when BTP reports a PROCESSING_FAILED
+// assignment for this entitlement and the reported amount is zero or unset,
+// i.e. nothing is actually reserved on the BTP side. This is distinct from
+// PROCESSING_FAILED with a non-zero amount, which typically reflects a
+// delete- or update-time failure on an entitlement that is still assigned
+// (and which should remain marked as Available so siblings/orchestration are
+// not flapped).
+func (c *external) assignFailedNoQuota(cr *apisv1alpha1.Entitlement) bool {
+	if cr.Status.AtProvider == nil || cr.Status.AtProvider.Assigned == nil {
+		return false
+	}
+	if cr.Status.AtProvider.Assigned.EntityState != apisv1alpha1.EntitlementStatusProcessingFailed {
+		return false
+	}
+	amount := cr.Status.AtProvider.Assigned.Amount
+	return amount == nil || *amount <= 0
+}
+
 func (c *external) needsCreate(cr *apisv1alpha1.Entitlement) bool {
-	return cr.Status.AtProvider.Assigned == nil
+	if cr.Status.AtProvider.Assigned == nil {
+		return true
+	}
+	// Previous assign attempt failed and BTP reserved nothing — treat as
+	// not-yet-created so Crossplane re-issues the assign via Create (which
+	// calls CreateInstance == UpdateInstance) under the managed reconciler's
+	// rate-limited retry.
+	return c.assignFailedNoQuota(cr)
 }
 
 // deletionComplete checks whether this CR's portion has already been removed from BTP.
--- a/internal/controller/account/entitlement/entitlement_test.go
+++ b/internal/controller/account/entitlement/entitlement_test.go
@@ -250,6 +250,89 @@ func TestObserve(t *testing.T) {
 				err: nil,
 			},
 		},
+		"Assign-time PROCESSING_FAILED with amount=0 (enable-style) -- retries via Create": {
+			args: args{
+				kube: &test.MockClient{
+					MockStatusUpdate: noopStatusUpdate,
+					MockList:         test.NewMockListFn(nil, ListEntitlements(entitlement(withEnabled(true)))),
+				},
+				client: fake.MockClient{MockDescribeCluster: func(ctx context.Context, input v1alpha1.Entitlement) (*entitlement2.Instance, error) {
+					return &entitlement2.Instance{
+						EntitledServicePlan: &entclient.ServicePlanResponseObject{
+							Category: internal.Ptr("ELASTIC_SERVICE"),
+						},
+						Assignment: &entclient.AssignedServicePlanSubaccountDTO{
+							Amount:       internal.Ptr(float32(0)),
+							EntityState:  internal.Ptr("PROCESSING_FAILED"),
+							StateMessage: internal.Ptr("Failed to call Provisioning service to assign quota."),
+						},
+					}, nil
+				}},
+				cr: entitlement(withEnabled(true)),
+			},
+			want: want{
+				// Observe returns ResourceExists=false so the managed reconciler
+				// drives Create -> CreateInstance, which re-issues the assign.
+				o:   managed.ExternalObservation{ResourceExists: false},
+				err: nil,
+			},
+		},
+		"Assign-time PROCESSING_FAILED with nil amount (enable-style) -- retries via Create": {
+			args: args{
+				kube: &test.MockClient{
+					MockStatusUpdate: noopStatusUpdate,
+					MockList:         test.NewMockListFn(nil, ListEntitlements(entitlement(withEnabled(true)))),
+				},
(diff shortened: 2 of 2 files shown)


internal/controller/account/entitlement/entitlement.go
```diff
@@ -139,14 +139,28 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 	switch cr.Status.AtProvider.Assigned.EntityState { //nolint:exhaustive
 	case apisv1alpha1.EntitlementStatusOk:
 		cr.Status.SetConditions(xpv1.Available())
-	// the state relates to the last operation, not the current state
-	// at this point we already verified that the entity does match the desired state for all we know
-	// so we will ignore this state to avoid blocking healthy resources
-	// example case: attempting to delete an entitlement which is already consumed will fail in that way
-	// -> the only resolution is to discard and recreate the cr, in this case we just need to observe the still existing entitlement
-	// despite the last operation being a failure
+	// PROCESSING_FAILED reflects the *last operation* on the entitlement,
```
sdischer-sap, line 142: General remark: 
This is a trivial fix and does not need to walls of text of explanation both in code as well as PR. No need to change, but general remark about our claude era ;)

sdischer-sap (approved):

PR #753, opened 2026-06-25
docs: explain ResourceUsage tracking mechanism

maximilianbraun:
## Summary

Adds `docs/contribution-notes/resource-usage-tracking.md` — a contribution note describing the provider's in-house dependency tracker:

- **Context**: why BTP-side cleanup is order-sensitive and why Crossplane's own model (owner refs between MRs) doesn't cover it.
- **Note: upstream Crossplane `Usage` came later**: how the in-house `ResourceUsage` compares to the framework-layer `protection.crossplane.io/v1beta1.Usage` introduced in Crossplane v1.14 (auto-stamp via reflection vs. user-declared, advisory `Delete()` check vs. admission webhook, no replay vs. `replayDeletion`).
- **Mechanism**: the five stages — reference discovery via `reflectwalk` on `reference-*` struct tags, RU creation in `createTracking`, condition wiring via `SetConditions`, the `DeleteShouldBeBlocked` check, and the enforcement surface.

Pure documentation. No code changes, no behaviour changes.

## Why now

The mechanism is load-bearing for every cross-MR delete in this provider but isn't written down anywhere. Recent incidents around stuck finalizers (and the discussion in #751) made the gap obvious. This note is the reference future contributors can read before touching `Connect()` / `Delete()` or proposing changes to dependency tracking.

## Test plan

- [x] Markdown renders (`docs/contribution-notes/` is already populated with sibling notes that follow the same H1+sections style)
- [x] No code changes, no build/test surface affected

PR #771, opened 2026-06-29
fix: refresh ResourceUsage condition in Track().

### LINKED ISSUE #755: Refresh ResourceUsage condition on Observe(), not only Delete()
## Problem

Today `tracker.SetConditions(ctx, mg)` runs only inside each upstream's `Delete()` flow. Example: `internal/controller/account/servicemanager/servicemanager.go:184`.

That means a steady-state upstream MR (`ServiceManager`, `CloudManagement`, `ServiceInstance`) has **no `ResourceUsage` condition at all** until the first delete attempt runs. The block at `internal/tracking/resourcetracker.go:303`:

```go
func (r *DefaultReferenceResolverTracker) DeleteShouldBeBlocked(mg resource.Managed) bool {
    if hasIgnoreAnnotation(mg) { return false }
    return mg.GetCondition(v1alpha1.UseCondition).Reason == v1alpha1.InUseReason
}
```

checks `Reason`, not `Status`. A missing condition returns `Reason == ""`, so `DeleteShouldBeBlocked` returns `false`. The first time anyone tries to delete the upstream, the block is not yet armed.

## Why it matters

The first delete attempt arms the condition, because `Delete()` calls `SetConditions` before checking the block. In a healthy steady state, the block works on the *second* reconcile. The race window is:

1. User issues `kubectl delete servicemanager …`.
2. Reconcile 1: `Delete()` runs. `SetConditions` stamps `ResourceUsage=True/ResourceUsagesFound`. Block engages, returns `ErrResourceInUse`. MR enters `Terminating`.
3. SM is now stuck in `Terminating` with the condition set correctly.
4. Operator gets impatient, patches the finalizer off, SM is gone.
5. Downstream MR's next reconcile fails forever (see #751 for the symptom).

Moving `SetConditions` into `Observe()` closes the gap. The condition stays current at all times. The block engages on the very first delete attempt, with no `Terminating` state in between. The operator never has a reason to reach for `--force`.

## Proposed fix

Add one call to each upstream's `Observe()`:

```go
func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.ExternalObservation, error) {
    // existing observe logic …
    c.resourcetracker.SetConditions(ctx, mg)
    // …
}
```

Upstream MRs that need this: `ServiceManager`, `CloudManagement`, `ServiceInstance`. Anything else that appears as a `reference-*` target should get it too. Grep `reference-kind` to enumerate them.

The existing `SetConditions` call in `Delete()` can stay as a safety net, or go. Either works.

## Tests

One table case per upstream controller in the existing `TestObserve` tables. Assert the condition is set when RUs exist, and cleared when they don't. No new test files needed.

## Out of scope

- The `Connect()`-during-deletion trap (#751) is a separate failure mode and is addressed there.
- Migrating to upstream Crossplane `Usage` (#754) would replace this entirely, but is a much larger change. This fix is a small, contained improvement to the current mechanism in the meantime.

## Asks

- Anyone see a reason `SetConditions` *should* stay in `Delete()` only? (Cost is one extra Get-and-Patch per Observe, which is cheap.)
- Should we drop the `Delete()` call once `Observe()` covers it, or keep both as a safety net?

Kukkerem:
## Summary

`tracker.SetConditions` ran only in each controller's `Delete()`, so a steady-state upstream MR carried no `ResourceUsage` condition until its first delete attempt. `DeleteShouldBeBlocked` keys on `Reason == ResourceUsagesFound`; a missing condition reads as empty, so the in-use block armed only on the *second* reconcile. That is the window in #755: the MR briefly enters `Terminating`, an operator reaches for `--force`, and a still-referenced resource is orphaned.

`Track()` runs in every `Connect()`, so moving the `SetConditions` call into `Track()` keeps the condition current in steady state - fresh before the first `Delete()`, so the block engages on the first delete attempt. Because it lives in `Track()`, every tracked resource is covered by its own reconcile, with no per-controller change.

## Changes

- Call `SetConditions` at the end of `Track()` in `internal/tracking/resourcetracker.go`.
- Serve `Track()`'s new `List` call in tests: add `MockList` to `FakeKubeClientBuilder.Build()` (`internal/testutils/kube_test_utils.go`) and to the tfclient setup test (`internal/clients/tfclient/setup_test.go`).
- Add a `Test_Track` case asserting the `InUse` condition is set when the resource is referenced by another.
- (The pre-existing `SetConditions` calls in each `Delete()` are unchanged and remain as a redundant refresh before the block check.)

## Test plan

- [x] `go test ./internal/tracking/... ./internal/clients/tfclient/...`

Closes #755


internal/controller/account/cloudmanagement/cloudmanagement.go
```diff
@@ -179,6 +179,7 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 	if !ok {
 		return managed.ExternalObservation{}, errors.New(errNotCloudManagement)
 	}
+	c.tracker.SetConditions(ctx, cr)
```
sdischer-sap: This adds just another condition of "how to implement a proper resource". 
I wasn't even aware that this conditions is whats required when checking to prevent deletions. 

Can't we build wire this logic into the Track() method? This one is called in each Connect(). It might as well deal with everything related to tracking reconcilations, not just creating ResourceUsage entities itself. This would reduce the surface error of proper implementation and you would only need a templated track call in the Connect and the Check for blocking deletion.

internal/testutils/tracker_real.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/sap/crossplane-provider-btp/internal/tracking"
+)
+
+// NewRealResourceTracker returns a real DefaultReferenceResolverTracker backed by a
+// controller-runtime fake client. When inUse is true the client is seeded with a single
+// ResourceUsage whose source-UID label equals mg's UID, so tracker.SetConditions(ctx, mg)
+// stamps an InUse condition; when false, no usages exist and it stamps NotInUse.
+//
+// Set mg's UID before calling.
+func NewRealResourceTracker(t *testing.T, mg resource.Managed, inUse bool) tracking.ReferenceResolverTracker {
```
sdischer-sap: Is it required to have yet another fake for this tracker? 
I think we have a ton of those already. Maybe we can unify?

sdischer-sap (commented):
Maybe we can simplify?

sdischer-sap (approved):
Haha nice, how few code changes that actually turned out to take. Great work. 
Code is great, just asking for two follow up tasks:
Could you update the PR description to reflect those changes (Can be super simply, just remove the false statements)?
Could you please update the documentation in https://github.com/SAP/crossplane-provider-btp/pull/753 to reflect those changes. Should be just some minor tweaks I think in the last sections. 
Thanks for your efforts!

Kukkerem:
> Haha nice, how few code changes that actually turned out to take. Great work. Code is great, just asking for two follow up tasks: Could you update the PR description to reflect those changes (Can be super simply, just remove the false statements)? Could you please update the documentation in #753 to reflect those changes. Should be just some minor tweaks I think in the last sections. Thanks for your efforts!

Thanks for the review! Fixed the description, and modified the doc in #753, changed some references as well.

PR #777, opened 2026-06-30
feat: backport terraform-provider-btp v1.23.1 to release/v1_10_x

### LINKED ISSUE #521: Use v1.22 of btp terraform provider

### LINKED ISSUE #724: [BUG] SubaccountTrustConfiguration: Create fails with "inconsistent result after apply" when `description` is set

xdj2e9:
**Note:** `Pending decision to postpone the v2.x.x release. If we do a v1.11.0 release from current main, this backport is not necessary anymore.`

Backports the v1.23.1 provider bump (#521, PR https://github.com/SAP/crossplane-provider-btp/pull/750) and the trust-configuration description workaround removal (#724) to `release/v1_10_x`.

Closes #521 on v1_10_x
Closes #724 on v1_10_x

## Why this is not a direct cherry-pick

The v1.23.1 work on `main` was built on top of the crossplane-runtime + upjet v2 upgrade (#653). The `release/v1_10_x` branch is on crossplane-runtime v1.20.0 / upjet v1.11.0, and the v1.10.x release line should stay on the v1 runtime — pulling in v2 as part of a backport would be a larger change than appropriate for a patch release on this branch.

This PR therefore reproduces the v1.23.1 payload on the v1 stack instead of cherry-picking or merging the feature branch.

## What is the same as the original PR

- Same identity injector and post-generate patcher (#521).
- Same CLI swap to OpenTofu 1.12.3.
- Same `controller-tools` v0.20.1 pin (fixes the pointer-`SecretKeySelector` deepcopy bug introduced by the new provider schema).
- Same removal of the `SubaccountTrustConfiguration` description workaround (#724).

Static comparison of upjet v1.11.0 and v2.2.0 showed that the parts the injector depends on (`controller.Store`, `tjcontroller.NewConnector`, the `StateV4` shape, `FileProducer.EnsureTFState()`) are unchanged between the two majors. The mechanism works the same way on v1.11.0 as it does on v2.2.0.

## What differs from the original PR

- **Module paths** stay on v1: `github.com/crossplane/upjet` (v1.11.0) and `github.com/crossplane/crossplane-runtime` (v1.20.0) instead of the `/v2` variants.
- **`CallbackProvider` parameter type**: v1 uses `string`, v2 uses `types.NamespacedName`. The v1 signature is restored in `internal/clients/tfclient/tf_async_callback.go`.
- **No `LegacyManaged` adjustments**: v1 has no split between `Managed` and `LegacyManaged`, so the casts and the `NewLegacyProviderConfigUsageTracker` / `fake.LegacyManaged` usages from the feature branch are not needed.
- **PR #708 (`__raw` connection-detail semantics) is not part of this backport**: the change is a behaviour change and should land on `release/v1_10_x` separately if needed. The release-branch inline `flattenSecretData` is kept.
- **Unrelated changes from the feature branch are excluded**: the Makefile additions (`local-deploy-prebuilt`, `BUILD_REGISTRY`, `ACCEPTANCE_DEPLOY`) and the Dockerfile cleanup (dropping the `:3.23.3` alpine tag).

Other v1 → v2 differences in upjet (`tjcontroller.Options.SecretStoreConfigGVK` / `ESSOptions`, `handler.EventHandler.RequestReconcile`, v2-only `FileProducer.hasTFID`) do not affect any code path used by this provider.

## Test plan

- [x] `go build ./...`, `go vet ./...` pass
- [x] `go test ./...` passes, including the identity injector unit tests
- [x] Post-generate patcher wraps `o.WorkspaceStore` in all 5 affected generated controllers (`directoryentitlement`, `subaccountservicebroker`, `globalaccounttrustconfiguration`, `subaccountapicredential`, `subaccounttrustconfiguration`)
- [x] E2E (`make test-acceptance`) green on the higher-risk paths: `Test_SubaccountTrustConfiguration_v1alpha1`, `TestServiceManagerCreationFlow`, `TestServiceManagerImport`, `Test_DirectoryEntitlement_v1alpha1` (three-field identity), `TestSubaccountApiCredentialsStandalone`, `TestSubaccountApiCredentialOrphanImport`
- [x] Long-running E2E tests (`make test-e2e-long`) executed
- [x] Baseline upgrade test (`make upgrade-test testFilter=Test_BaselineUpgradeProvider`) executed against the previous tagged release
- [x] Per-resource external-name upgrade tests (the remaining `*_external_name_upgrade_test.go` cases under `test/upgrade/`) executed against the previous tagged release


sdischer-sap (approved):
LGTM

PR #778, opened 2026-06-30
chore(test): render e2e fixtures into .work/ instead of overwriting templates

xdj2e9:
## What & why

`make generate-test-crs` used to run `envsubst` over `test/e2e/testdata/crs/` and write the result back on top of the committed templates. This left the working tree dirty. Same issue for `test/upgrade/testdata/`.

## How

Templates are now copied into `.work/rendered-crs/{e2e,upgrade}/` (gitignored) and `envsubst` runs there, with an explicit allowlist of the four vars actually used. Go tests resolve fixture paths via two small helpers (`crsPath`, `upgradeCRsPath`) that read the rendered dir from env, with a fallback to the in-tree templates so plain `go test` still works. Layout and approach match `crossplane-provider-cloudfoundry`.


Makefile
```diff
@@ -414,6 +429,11 @@ upgrade-test-debug: $(KIND) check-upgrade-test-vars build-upgrade-test-images pu
 	 esac
 
 .PHONY: upgrade-test-restore-crs
+# Restores `test/upgrade/testdata/baseCRs` after `pull-upgrade-test-version-crs`
```
sdischer-sap: I don't understand this comment.

xdj2e9: That was confusing indeed. What do you think now?

sdischer-sap: Thank you , now I understand it, but this raises the question than if those are still required. 
If we do not use this directory anymore I guess we do not need a goal to restore its contents right?

xdj2e9: The upgrade tests check out the old state of the template files. For that we actually need to restore to current HEAD (of the branch). So that case still remains. Only the substitution was adapted to write to a separate gen dir.

sdischer-sap: Ahh yes of course makes sense. Didn't recall that part.

sdischer-sap (approved):

PR #783, opened 2026-07-02
fix(cache): concurrency hardening + PR #744 verification tests

maximilianbraun:
## Summary

Follow-up to #744: audit of the two caches introduced there surfaced three concurrency hazards on the entitlement reconcile hot path. This PR fixes them and adds end-to-end tests that verify #744's headline call-count claims on the wire.

| # | Where | Hazard | Fix |
|---|---|---|---|
| CTL-1 | `internal/clients/entitlement/entitlement.go` `describeCacheGet` | `sync.Map.Delete` on TTL expiry can wipe a fresh entry stored by a concurrent writer between our Load and Delete. Probe: **98–127 wipes / 5000 iters**. | `describeCache.CompareAndDelete(key, e)` |
| CTL-2 | same file | Cache never sweeps: keys not re-Loaded leak for process lifetime. Probe: **200 expired entries survive** with no janitor. | New `describeSweep` + lazy `startDescribeSweeper` goroutine (ticks every `describeCacheT`). |
| CTL-3 | `btp/cisclient.go` `NewServiceClientWithCisCredential` | Plain `Load` + `LoadOrStore` lets N concurrent first-callers each run `createClient()` and discards N-1. Probe: **5/32 builds** under contention. | `golang.org/x/sync/singleflight.Group.Do` keyed by `credentialCacheKey`. |

All three probes are checked in and now assert the fixed behavior (**0/5000 wipes**, **200→0 evicted**, **1/32 builds**) across 3 stress runs under `-race`.

## Verifying the original PR #744 claims

Adds `internal/httpcount` — a test-only counting `http.RoundTripper` (atomic total + per-`(method, path)` and per-host counts under `sync.RWMutex`) — plus five end-to-end tests that wire a real openapi client against an `httptest.Server` and count the actual HTTP calls that leave the process:

- `TestClientCache_SharedTokenSource_OnePOSTAcrossManyGETs` — 10 GETs across sub-clients → **1 POST `/oauth/token`** (PR #744 §1: shared `oauth2.ReuseTokenSource`).
- `TestDescribeInstance_CachedWithinTTL` — 10 sequential describes → **1 GET `/entitlements/v1/assignments`** (PR #744 §2: 30s TTL).
- `TestDescribeInstance_SingleflightConcurrent` — 32 concurrent describes → **1 GET** (PR #744 §2: singleflight).
- `TestDescribeInstance_DistinctKeys_NoSharing` — 5 distinct keys → **5 GETs**.
- `TestUpdateInstance_InvalidatesDescribeCache` — describe → describe → PUT → describe → **2 GETs + 1 PUT** (PR #744 §2: post-write invalidation).

## Production surface

Two small test seams added to `btp/cisclient.go`:

```go
var buildClientFn = createClient
var newBaseHTTPClientFn = func() *http.Client { return DebugPrintHTTPClient() }
```

Zero runtime behavior change; they exist solely so tests can count build invocations and inject a counting roundtripper between oauth2 and the network without rewiring the production paths.

## Commits

- `fix(cache):` — the three prod fixes.
- `test(cache):` — `internal/httpcount` package + race probes + call-count tests.
- `docs(caching):` — updates `docs/contribution-notes/caching.md` to reflect the hardening and point contributors at the new test hooks.

Bisect-safe: each commit builds and each test suite is green under `-race`.

## Test plan

- [x] `go build ./...` clean
- [x] `go test -race -count=1 ./...` green
- [x] Race probes stable across `-count=3` under `-race` (0 wipes, 0 leaks, 1 build)
- [x] Call-count tests deterministic across 3 runs
- [ ] Review by maintainers — especially the test-seam pattern (`buildClientFn`, `newBaseHTTPClientFn`) if you'd prefer a different injection point


xdj2e9 (dismissed):
The injection points for testing are fine.

Thank you for the hardening!

@maximilianbraun Could you please rebase and fix the conflicts so we can merge this?

xdj2e9:
Rebased onto latest main.

Changes that are left:
  - Ensure `cisclient` is created exactly once per credential (fixing first creators race)
  - Ensure freshly cached directory assignments are not deleted by a concurrent reconilation.

PR #795, opened 2026-07-06
feat: add external-name ADR compliance for GlobalaccountTrustConfiguration

### LINKED ISSUE #546: [TASK] `External-Name`(GlobalaccountTrustConfiguration): Ensure ADR compliance
**Part of Epic: #427** 
## Context

## Technical Tasks Checklist

**Acceptance Criteria:**

- [ ] **Migration Logic**:

**Acceptance Criteria:**
- [ ] **Observe() Tests**:

- [ ] **Create() Tests**:

- [ ] **Update() Tests**:

- [ ] **Delete() Tests**:

**Acceptance Criteria:**
- [ ] **Import Flow Test**:

### 5. Documentation

Kukkerem:
Implements ADR-compliant external-name handling for `GlobalaccountTrustConfiguration` as part of Epic #427.

The **origin key** (e.g. `sap.custom`) is used as the external-name identifier. Unlike `SubaccountTrustConfiguration` (#671), no compound key is needed: the global account scope is fully determined by the provider credentials, so the origin key alone satisfies the ADR requirement that the external-name is self-contained. It also matches the Terraform import ID of `btp_globalaccount_trust_configuration`, so the existing `config.IdentifierFromProvider` setup already produces the correct value — no `GetExternalNameFn`/`GetIDFn`/initializer changes are required.

## Changes

- `apis/security/v1alpha1/globalaccounttrustconfiguration_types.go` — external-name documentation comment
- `docs/end-user-guides/import-landscape/external-name.md` — regenerated via `make docs.generate-external-name`
- `test/e2e/globalaccount_trust_configuration_v1alpha1_test.go` — import flow e2e test
- `test/upgrade/` — upgrade test verifying the external-name (origin key) is a real origin key before the upgrade and survives the provider upgrade unchanged (no format migration). `fromVersion` is `v1.3.0`: the earliest tag where the CRD and the `identityProvider` field exist.

## Import example

External-name = origin key. Required `spec.forProvider` fields must still be set in an import scenario (see ADR):

​```yaml
kind: GlobalaccountTrustConfiguration
metadata:
  annotations:
    crossplane.io/external-name: sap.custom
  name: ga-trustconfiguration-observe
spec:
  forProvider:
    identityProvider: <idp-url>
  providerConfigRef:
    name: account-provider-config
​```

Observe-only import (`managementPolicies: [Observe]`) was tested manually and works.

Closes #546


sdischer-sap (dismissed):
Looks good to me. 
I think the custom upgrade test isn't really required if we have the baseline upgrade tests. But yes generally doesn't hurt and baseline tests are still pending, we can still remove later.

sdischer-sap:
@Kukkerem could you please resolve those conflicts?

Kukkerem:
> Looks good to me. I think the custom upgrade test isn't really required if we have the baseline upgrade tests. But yes generally doesn't hurt and baseline tests are still pending, we can still remove later.

Is baseline enough when the external-name handling changes? It asserts health only, so it can't tell a silently recreated resource from one that survived, and the external-name format (preservation/migration) isn't checked at all.

sdischer-sap:
> Is baseline enough when the external-name handling changes? It asserts health only, so it can't tell a silently recreated resource from one that survived, and the external-name format (preservation/migration) isn't checked at all.

Did the external-name handling change? The external-name stays the same doesn't it so I guess this custom upgrade tests just verifys that and health status as well. In this case its the origin, which will always be `sap.custom` or something like that.

Kukkerem:
> > Is baseline enough when the external-name handling changes? It asserts health only, so it can't tell a silently recreated resource from one that survived, and the external-name format (preservation/migration) isn't checked at all.
> 
> Did the external-name handling change? The external-name stays the same doesn't it so I guess this custom upgrade tests just verifys that and health status as well. In this case its the origin, which will always be `sap.custom` or something like that.

I meant it as a regression guard for future upgrades. But as you said, we could remove it later if needed.

sdischer-sap:
> I meant it as a regression guard for future upgrades. But as you said, we could remove it later if needed.

Valid point actually. I always thought that the upgrade tests are testing specifically from one version jump to the next, but since its to version is open ended thats an good addition. Thank you!

sdischer-sap (approved):

PR #812, opened 2026-07-06
[Bug] add `environment` to distinguish service plan ID resolver

dwlou:
## Problem

The service plan ID resolver queries the Service Manager API for plans matching `catalog_name eq '<plan>'`. Without an environment filter, the SM API can return plans from multiple environments (`sapbtp`, `kubernetes`, `cloudfoundry`). When more than one plan is returned the resolver silently picks `Items[0]`, which may be the wrong plan for the intended environment.

## Fix

- Adds `spec.forProvider.environment` to the `ServiceInstance` CR
  - Enum: `sapbtp` | `kubernetes` | `cloudfoundry`
  - Default: `sapbtp`
- Passes the environment as a filter to the SM `GetAllServicePlans` query via the existing `.Environment()` parameter
- Updated error message includes the environment when no plan is found
- `cloudmanagement` and `service_instance_proxy_client` callers hardcoded to `sapbtp` (not user-facing)

## Test plan

- [ ] Unit tests pass: `go test ./internal/clients/servicemanager/... ./internal/controller/account/serviceinstance/... ./internal/controller/account/cloudmanagement/...`
- [ ] Verify `environment` field appears in CRD after `make generate`
- [ ] Manual test: create `ServiceInstance` with `environment: kubernetes` and confirm correct plan ID is resolved
- [ ] Existing `ServiceInstance` resources without `environment` field continue to work (default `sapbtp`)

### DIFF
--- a/internal/clients/servicemanager/api_plan_resolver.go
+++ b/internal/clients/servicemanager/api_plan_resolver.go
@@ -24,7 +24,7 @@ const ErrMissingXsappname = "Xsappname (xsappname) is missing"
 
 // PlanIdResolver used as its own resolval implementation downstream
 type PlanIdResolver interface {
-	PlanIDByName(ctx context.Context, offeringName string, servicePlanName string, dataCenter string) (string, error)
+	PlanIDByName(ctx context.Context, offeringName string, servicePlanName string, dataCenter string, environment string) (string, error)
 }
 
 // NewCredsFromOperatorSecret creates a new BindingCredentials from a secret data
@@ -119,8 +119,11 @@ func NewServiceManagerClient(ctx context.Context, creds *BindingCredentials) (*S
 	}, nil
 }
 
-func (sm *ServiceManagerClient) PlanIDByName(ctx context.Context, offeringName, planName, dataCenter string) (string, error) {
-	offeringQuery := fmt.Sprintf("catalog_name eq '%s' and data_center eq '%s'", offeringName, dataCenter)
+func (sm *ServiceManagerClient) PlanIDByName(ctx context.Context, offeringName, planName, dataCenter, environment string) (string, error) {
+	offeringQuery := fmt.Sprintf("catalog_name eq '%s'", offeringName)
+	if dataCenter != "" {
+		offeringQuery = fmt.Sprintf("catalog_name eq '%s' and data_center eq '%s'", offeringName, dataCenter)
+	}
 
 	execute, _, err := sm.GetServiceOfferings(ctx).FieldQuery(offeringQuery).Execute()
 	if err != nil {
@@ -134,13 +137,13 @@ func (sm *ServiceManagerClient) PlanIDByName(ctx context.Context, offeringName, 
 	}
 
 	planQuery := fmt.Sprintf("catalog_name eq '%s' and service_offering_id eq '%s'", planName, *execute.Items[0].Id)
-	object, _, err := sm.GetAllServicePlans(ctx).FieldQuery(planQuery).Execute()
+	object, _, err := sm.GetAllServicePlans(ctx).FieldQuery(planQuery).Environment(environment).Execute()
 	if err != nil {
 		return "", specifyAPIError(err)
 	}
 
 	if len(object.Items) == 0 {
-		return "", errors.Errorf("No service plan '%s' found for offering '%s'", planName, offeringName)
+		return "", errors.Errorf("No service plan '%s' found for offering '%s' in environment '%s'", planName, offeringName, environment)
 	}
 
 	servicePlanID := *object.Items[0].Id
--- a/apis/account/v1alpha1/serviceinstance_types.go
+++ b/apis/account/v1alpha1/serviceinstance_types.go
@@ -36,6 +36,13 @@ type ServiceInstanceParameters struct {
 	// +kubebuilder:validation:Optional
 	ServicePlanID string `json:"servicePlanID,omitempty"`
 
+	// Environment for which to resolve the service plan. Valid values: sapbtp, kubernetes, cloudfoundry.
+	// Defaults to sapbtp if not set.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=sapbtp;kubernetes;cloudfoundry
+	// +kubebuilder:default=sapbtp
+	Environment string `json:"environment,omitempty"`
+
 	// Whether the service instance is shared or not
 	// +kubebuilder:validation:Optional
 	Shared *bool `json:"shared,omitempty"`
--- a/internal/clients/servicemanager/api_plan_resolver_test.go
+++ b/internal/clients/servicemanager/api_plan_resolver_test.go
@@ -193,7 +193,7 @@ func TestPlanIDByName(t *testing.T) {
 				OfferingServiceFake{tc.args.listOfferingsMockFn},
 				PlansServiceFake{listPlansMockFn: tc.args.listPlansMockFn},
 			}
-			planID, err := smClient.PlanIDByName(context.TODO(), "Not relevant, since mocked", "Not relevant, since mocked", tc.args.dataCenter)
+			planID, err := smClient.PlanIDByName(context.TODO(), "Not relevant, since mocked", "Not relevant, since mocked", tc.args.dataCenter, "sapbtp")
 
 			if tc.wantErr != (err != nil) {
 				t.Errorf("Unexpected error return; Expected error: %v, Returned: %v", tc.wantErr, err)
--- a/internal/clients/servicemanager/service_instance_proxy_client.go
+++ b/internal/clients/servicemanager/service_instance_proxy_client.go
@@ -71,7 +71,7 @@ func (t ServiceManagerInstanceProxyClient) resolveServicePlan(ctx context.Contex
 			return "", err
 		}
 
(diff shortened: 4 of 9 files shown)


internal/clients/servicemanager/api_plan_resolver.go
```diff
@@ -119,28 +119,24 @@ func NewServiceManagerClient(ctx context.Context, creds *BindingCredentials) (*S
...
 	}
 	if len(execute.Items) == 0 {
-		if dataCenter != "" {
-			return "", errors.Errorf("no service offering '%s' found in data center '%s'", offeringName, dataCenter)
-		}
-		return "", errors.Errorf("no service offering found for '%s'", offeringName)
+		return "", errors.Errorf("API returned no service plan for offering %s", offeringName)
 	}
 
 	planQuery := fmt.Sprintf("catalog_name eq '%s' and service_offering_id eq '%s'", planName, *execute.Items[0].Id)
```
dwlou, line 139: somehow fieldQuery looks strange to me. take a look at the example https://help.sap.com/docs/service-manager/sap-service-manager/filtering-parameters-and-operators?locale=en-US#query-task-examples. this would make more sense  `Sprintf("service_name eq '%s' and plan_name eq '%s'", offeringName, planName)`

sdischer-sap:
Maybe @jtucci might help reviewing this one, since he worked on a recent fix in that area (https://github.com/SAP/crossplane-provider-btp/pull/687)

jtucci:
> Maybe @jtucci might help reviewing this one, since he worked on a recent fix in that area (#687)

Yup, I've been out sick but will take a look at this soon

jtucci:
@dwlou 
>The service plan ID resolver queries the Service Manager API for plans matching catalog_name eq '<plan>'. Without an environment filter, the SM API can return plans from multiple environments (sapbtp, kubernetes, cloudfoundry). When more than one plan is returned the resolver silently picks Items[0], which may be the wrong plan for the intended environment.

Do you have an example of this that is reproducible? I tried testing the SM API against several services and couldn't find a case where the same catalog_name + service_offering_id query returns a different plan ID depending on environment. 

### xsuaa 

GET /v1/service_plans?fieldQuery=catalog_name eq 'application' and service_offering_id eq '9563a892-...'
```json
{ "num_items": 1, "items": [{ "id": "160eae09-be8b-4c7d-8e78-da9d2fb8bacd", "catalog_name": "application" }] }
```


GET /v1/service_plans?environment=sapbtp&fieldQuery=catalog_name eq 'application' and service_offering_id eq '9563a892-...'
```json
{ "num_items": 1, "items": [{ "id": "160eae09-be8b-4c7d-8e78-da9d2fb8bacd", "catalog_name": "application" }] }
```

GET /v1/service_plans?environment=cloudfoundry&fieldQuery=catalog_name eq 'application' and service_offering_id eq '9563a892-...'
```json
{ "num_items": 1, "items": [{ "id": "160eae09-be8b-4c7d-8e78-da9d2fb8bacd", "catalog_name": "application" }] }
```

Same plan ID (`160eae09...`) in all cases. The environment filter doesn't change the result.

### hana-cloud "hana" plan (multi-datacenter subaccount) 
GET /v1/service_plans?fieldQuery=catalog_name eq 'hana' and service_offering_id eq 'd0ec53c9-...'

```json
{ "num_items": 1, "items": [{ "id": "0f744562-8038-4f35-9567-08496ddb8730", "catalog_name": "hana" }] }
```

GET /v1/service_plans?environment=sapbtp&fieldQuery=catalog_name eq 'hana' and service_offering_id eq 'd0ec53c9-...'

```json
{ "num_items": 1, "items": [{ "id": "0f744562-8038-4f35-9567-08496ddb8730", "catalog_name": "hana" }] }
```


GET /v1/service_plans?environment=kubernetes&fieldQuery=catalog_name eq 'hana' and service_offering_id eq 'd0ec53c9-...'

```json
{ "num_items": 1, "items": [{ "id": "0f744562-8038-4f35-9567-08496ddb8730", "catalog_name": "hana" }] }
```

The CF environment does expose additional plans that don't exist in sapbtp (e.g., xsuaa has `space` and `apiaccess` plans only visible with `environment=cloudfoundry`). But for plans that exist in both, I think the plan ID is the same. 


GET /v1/service_plans?fieldQuery=catalog_name eq 'space' and service_offering_id eq '9563a892-...'
```json
{ "num_items": 0, "items": [] }
```


GET /v1/service_plans?environment=cloudfoundry&fieldQuery=catalog_name eq 'space' and service_offering_id eq '9563a892-...'
```json
{ "num_items": 1, "items": [{ "id": "4c5440cc-e232-4d3f-9e10-7649aa8b3ded", "catalog_name": "space" }] }
```

So the environment filter is useful for surfacing CF-only plans, but I'm not sure it disambiguates plans that already show up in the default query. I know that setting the environment param can return plans that only exist in certain environments, but wouldn't those be handled by their respective providers (e.g. a Cloud Foundry provider)? The SM API's `POST /v1/service_instances` is meant for creating subaccount-level instances so if you needed to create Cloud Foundry instances for example, wouldn't that be done through the CF provider?




Separately, this PR also removes the `data_center eq ''` clause from the offering query when `dataCenter` is empty. This is a regression, `data_center eq ''` was added intentionally in #687 to handle multi-datacenter services like hana-cloud. Without it, the offering query returns multiple items and the resolver picks `Items[0]` arbitrarily. The comments in that PR describe why it is needed.

dwlou:
> Do you have an example of this that is reproducible?

@jtucci Can you test using SM offering/plan `postgresql-db`/`standard`?  This is a issue for provisioning postgresql and it seems that sm api resolve to a different plan id than the BTP cli. More in this [ticket](https://github.tools.sap/openmcp/support/issues/579).

jtucci:
> > Do you have an example of this that is reproducible?
> 
> @jtucci Can you test using SM offering/plan `postgresql-db`/`standard`? This is a issue for provisioning postgresql and it seems that sm api resolve to a different plan id than the BTP cli. More in this [ticket](https://github.tools.sap/openmcp/support/issues/579).

Hey @dwlou, I tested with `postgresql-db` as you suggested. The results are consistent with the previous test, some offerings **only exist** under a specific environment type and aren't visible at all under the default `sapbtp` environment. In this case, postgres standard plan is only available under k8s environment. 


```bash
# Default (sapbtp) - offering not found
GET /v1/service_offerings?fieldQuery=catalog_name eq 'postgresql-db'
→ {"num_items": 0, "items": []}

# With environment=kubernetes - offering found
GET /v1/service_offerings?fieldQuery=catalog_name eq 'postgresql-db'&environment=kubernetes
→ {"num_items": 1, "items": [{"id": "488aa064-089e-477e-abc5-33faa8790a12", "catalog_name": "postgresql-db"}]}

# Plans only resolve under kubernetes environment
GET /v1/service_plans?fieldQuery=service_offering_id eq '488aa064-...'&environment=kubernetes
→ standard:          46e9b037-c507-4654-938a-b5dc964b99f6
→ premium:           ede2e2c4-98f9-4cfd-bcc1-7590d643c5e9
→ shared-postgresql: ce1bfee0-aa48-4823-9254-45b4e605bc23
→ reference:         e3f08e82-e4b7-4fe1-924b-fe39c1ea3d4e
```

Same behavior through the BTP CLI:
```bash
btp list services/offering --environment kubernetes --fields-filter "name eq 'postgresql-db'"
→ 1 result

btp list services/offering --fields-filter "name eq 'postgresql-db'"
→ 0 results
```

Interestingly, in my subaccount I see the offering *completely absent* under the default environment (0 results), whereas in the linked issue (#579) it sounds like the SM API *does* return a plan ID under the default environment just a different (invalid) one? I haven't been able to reproduce that dual-plan-ID scenario myself yet. Do you know if the behavior varies by landscape or if the postgres team's recent "Other environment" support change could explain that difference?



Regarding the fix approach: I'm wondering whether exposing `environment` as a user-facing field is the best UX here, it expects users to know which SM environment their service is registered under. Would a fallback strategy (try default, retry with `kubernetes` if no offering found) be more user-friendly? Though if it turns out some subaccounts return *different* (wrong) plan IDs rather than empty results under the default environment, a simple fallback might not be sufficient. 

What do you think?

As a temporary workaround, have you tried setting `servicePlanID` directly? When set, the plan name resolution is skipped entirely so the environment lookup isn't involved. You can get the plan ID from the BTP CLI with `--environment kubernetes` and pass it straight through:

```yaml
spec:
  forProvider:
    servicePlanID: "46e9b037-c507-4654-938a-b5dc964b99f6"  # postgresql-db/standard
```

dwlou:
> In this case, postgres standard plan is only available under k8s environment.

postgres added sapbtp recently. This ticket is currently reopened because of this issue. 

'environment' is exposed by sm api as a parameter. As long as the same service offering/plan is available in more than one environment, we will need to require/allow user to disambiguate this because we cannot determine this for the user or assume other/sapbtp because it is the service-side default. 
image

btw, sap terraform provider added an enhancement: https://github.com/SAP/terraform-provider-btp/pull/1498.. which would eliminate the need for this plan id resolver ..

dwlou:
Won't merge, because `environment` filter is effectively a no-op in the SM API. The `credential` type (depends on which plan used for service manger instance/binding) alone controls the API visibility. e.g., `service-operator-access` sees only `kubenetes`, `subaccount-admin` sees only `sapbtp`.

PR #815, opened 2026-07-07
fix: make SubaccountApiCredential external-name compound-key based

### LINKED ISSUE #674: [TASK] External-Name (SubaccountApiCredential): Use ADR-compliant compound key
Follow-up to #553 (`[TASK] External-Name (SubaccountApiCredential): Ensure ADR compliance`).

## Problem

`SubaccountApiCredential` currently uses only the `name` field as the external-name identifier (e.g. `my-credential`). However, the Terraform BTP provider requires **both** `subaccount_id` and `name` to read, update, and delete the resource — making the external-name not self-contained.

The ADR requires a compound key with `/` as delimiter: https://github.com/SAP/crossplane-provider-btp/blob/7df0c8a3000500c58937c960ec520de3d5507e8d/docs/contribution-notes/external-name-handling.md?plain=1#L38-L43

The correct ADR-compliant external-name format would be: `<subaccount-id>/<name>` (e.g. `abc-123-def-456/my-credential`).

For an example implementation see #547 (`[TASK] External-Name (SubaccountTrustConfiguration): Ensure ADR compliance`).

## Additional Validation

In the same scope:
- Validate that the import flow works correctly end-to-end with the compound key
- Change the export CLI external format to match this format

gergely-szabo-sap:
## Summary

Closes #674.

This PR makes `SubaccountApiCredential` external-name handling ADR-compliant by switching from the legacy name-only annotation to the compound key format:

```text
<subaccount-id>/<name>
```

## Changes

- Reconstruct `crossplane.io/external-name` from Terraform state using both `subaccount_id` and `name`.
- Ensure Terraform still receives only the credential name in the `name` argument, not the full compound external-name.
- Add legacy migration for existing name-only annotations to `<subaccount-id>/<name>`.
- Preserve the existing credential-name segment during migration to avoid retargeting the managed external resource.
- Avoid deleting connection secrets during external-name migration.
- Keep import/adoption of existing `SubaccountApiCredential` resources unsupported, matching Terraform provider limitations.
- Update external-name documentation, generated API comments, and CRD docs.
- Extend unit, E2E, and upgrade coverage for the new external-name behavior.

## Notes

`btp_subaccount_api_credential` does not support Terraform import, and the credential secret cannot be fully reconstructed after creation. This PR therefore does not add adoption/import support for existing API credentials.

## Testing

- Added/updated config unit tests for compound external-name parsing, reconstruction, and legacy migration.
- Updated E2E coverage to assert the post-create external-name format.
- Updated upgrade coverage to verify migration from legacy name-only annotations.


apis/security/v1alpha1/subaccountapicredential_types.go
```diff
@@ -2,8 +2,9 @@ package v1alpha1
 
 // External-Name Configuration:
 //   - Resource: SubaccountApiCredential
-//   - Follows Standard: no (uses credential name as identifier, not a GUID)
-//   - Format: Credential Name (string)
+//   - Follows Standard: no (compound key; Terraform identifies credentials by subaccount ID and credential name)
+//   - Format: `<subaccount-id>/<name>` (e.g. "abc-123-def-456/my-credential")
+//   - Note: Existing name-only annotations are migrated automatically to the compound-key format; importing/adopting existing credentials is unsupported by the Terraform provider.
```
xdj2e9: Can we make the information about import / adopting not being supported part of the CRD doc?

config/btp_subaccount_api_credential/config.go
xdj2e9, line 1: I see potential problems if the user provides malformed external names. E.g. `/foo`, `foo/` and `a/b/c` will end up providing `/foo`, `foo/` and `b/c` to the `name` property of the terraform resource. Depending on how this is handled in the terraform provider or by the BTP API we could end up with internal errors from terraform / BTP API or in worst case with malformed resources created on BTP.

If that is the case I would like to fail as early as possible (probably in the `Initialize` of `CompoundExternalNameInitializer`) with a user facing error.

apis/security/v1alpha1/subaccountapicredential_types.go
```diff
@@ -2,8 +2,9 @@ package v1alpha1
 
 // External-Name Configuration:
 //   - Resource: SubaccountApiCredential
-//   - Follows Standard: no (uses credential name as identifier, not a GUID)
-//   - Format: Credential Name (string)
+//   - Follows Standard: no (compound key; Terraform identifies credentials by subaccount ID and credential name)
```
sdischer-sap: Lets remove "terraform" here. The User does not need to know about inner dependencies

xdj2e9 (commented):
Would be nice to have the import limitation as part of the CRD doc.

For the second comment I'm actually unsure if this is somehow handled differently and a non-issue.

test/e2e/subaccount_api_credential_v1alpha1_test.go
```diff
@@ -92,15 +93,12 @@ func assertApiCredentialSecret(t *testing.T, ctx context.Context, cfg *envconf.C
...
 //
-// Note: the BTP Terraform provider does not implement ImportState for this resource type.
-// Setting the external-name annotation before creation would trigger Upjet's terraform-import
-// path, which always fails for this resource. The external-name is therefore set by the
-// provider after successful creation, not by the user before creation.
-//
-// See: https://github.com/SAP/crossplane-provider-btp/issues/553
+// Note: the BTP Terraform provider does not implement ImportState for this resource type,
+// so this test does not validate adoption/import of an existing credential.
 func TestSubaccountApiCredentialOrphanImport(t *testing.T) {
```
sdischer-sap: I am not really sure anymore what our goal has been with this test. 
Do you remember? I thought we added something that rejects import attempts with an error, but looks like not.

sdischer-sap (changes requested):
Nicely done. 
Just one minor request to change comments.

apis/security/v1alpha1/subaccountapicredential_types.go
gergely-szabo-sap: Addressed in https://github.com/SAP/crossplane-provider-btp/pull/815/commits/6af92e07de3108cc334e2ff13113c9ee3338949a

config/btp_subaccount_api_credential/config.go
gergely-szabo-sap: Addressed in https://github.com/SAP/crossplane-provider-btp/pull/815/commits/2203f2c147a90b132167c92872d211e9f084f01e

apis/security/v1alpha1/subaccountapicredential_types.go
gergely-szabo-sap: addressed in https://github.com/SAP/crossplane-provider-btp/pull/815/commits/bd86e138c5ed606cf9d36ef59d7f1efd220a20ab

test/e2e/subaccount_api_credential_v1alpha1_test.go
gergely-szabo-sap: I have no clue what the reason was for the original test case.

Made a change in https://github.com/SAP/crossplane-provider-btp/pull/815/commits/f78984a4ae410d71f2382110bc0c0f2574e48c63

The test was renamed because its previous name, TestSubaccountApiCredentialOrphanImport, was misleading.                                                                                     
                                                                                                                                                                                              
 SubaccountApiCredential does not support Terraform import/adoption: the upstream BTP Terraform provider has no ImportState implementation for btp_subaccount_api_credential, and existing credentials cannot be fully adopted because client_secret is only available at creation time.                                                                                                
                                                                                                                                                                                              
 The test is not validating orphan import behavior. It creates a new SubaccountApiCredential, waits for provisioning, and verifies that the controller writes back an ADR-compliant compound  
 external-name annotation in the format:                                                                                                                                                      
                                                                                                                                                                                              
 ```text                                                                                                                                                                                      
   <subaccount-id>/<credential-name>                                                                                                                                                          
 ```                                                                                                                                                                                          
                                                                                                                                                                                              
 It also checks that the created connection secret contains client_secret.                                                                                                                    
                                                                                                                                                                                              
 So the test name and comment were updated to reflect its actual purpose: validating external-name ADR compliance for newly provisioned credentials, while explicitly avoiding any implication that import/adoption of existing credentials is supported.

xdj2e9 (approved):
My points have been addressed. LGTM

Thank you!

PR #819, opened 2026-07-08
fix(e2e): derive crossplane version from chart tarball filename

sdischer-sap:
We introduced an issue in https://github.com/SAP/crossplane-provider-btp/pull/785 which breaks all of our e2e tests. (This only came up on main, because the GH action workflows are pulled from main rather than the PR itself.)

Renovate bumped the workflow to `2.3.3` but missed the Go constant (still `2.1.3`), causing the e2e helm install to use the wrong version

## Changes
- When `CROSSPLANE_CHART_PATH` is set (CI path), the version is now parsed from the tarball filename (`crossplane-X.Y.Z.tgz`) — the two sources cannot drift
- Introduce `crossplaneVersionDefault`, which is only used for local dev 

## Notes on testing
- PR should prove its working, because no GH action changes are done here, its all catched in code

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -5,6 +5,8 @@ package e2e
 import (
 	"context"
 	"os"
+	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/crossplane-contrib/xp-testing/pkg/envvar"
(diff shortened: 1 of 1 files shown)


sdischer-sap:
Turned out not to be the issue. I am not sure right now if binding the versions like that is a good solution, so I will close this one

PR #822, opened 2026-07-08
fix: Test_SubaccountTrustConfiguration_External_Name upgrade test

### LINKED ISSUE #760: [BUG] SubaccountTrustConfiguration upgrade test asserts external-name byte-equality but controller migrates legacy form to compound form
**Describe the bug**

`Test_SubaccountTrustConfiguration_External_Name` (added in #671) fails on `main`. The pre-upgrade external-name is the legacy `<origin>` form (e.g. `sap.custom`); the post-upgrade controller migrates it to the compound `<subaccount-id>/<origin>` form. The resource stays `Synced: True` / `Ready: True` after the upgrade, but the test asserts byte-equality across the upgrade and therefore fails.

`main` HEAD.

**To Reproduce**

```
make upgrade-test \
  UPGRADE_TEST_FROM_TAG=v1.3.0 \
  UPGRADE_TEST_TO_TAG=local \
  testFilter='Test_SubaccountTrustConfiguration_External_Name'
```

Fails with:

```
subaccounttrustconfiguration_external_name_upgrade_test.go:80: External name changed during upgrade:
    before="sap.custom", after="<sa-uuid>/sap.custom" (expected no change)
```

**Expected behavior**

Upgrade test passes. The likely fix is to update the assertion to validate the migration outcome (post-upgrade annotation is in `<subaccount-id>/<origin>` form with the same `origin` segment) rather than byte-equality. The right approach should be determined as part of the investigation.

xdj2e9:
Closes #760.

## Summary

- **Rewrite `Test_SubaccountTrustConfiguration_External_Name` to assert the migration outcome** — v1.3.0 writes the annotation as `<origin>`; the current controller reconstructs `<subaccount-id>/<origin>` on the first Observe via `GetExternalNameFn` (`config/subaccount_trust_configuration/config.go`). Post-upgrade the external-name must be in compound form, with segments matching `status.atProvider.origin` and `status.atProvider.subaccountId`.

## Additional

The `--backoff-base` / `--backoff-max` flags in `test.DeploymentRuntimeConfig` will now only be set on provider versions supporting them. These flags were introduced in [#578](https://github.com/SAP/crossplane-provider-btp/pull/578) and are unavailable on binaries older than `v1.8.0`, which made the initial provider install crash-loop with `unknown long flag '--backoff-base'` for any upgrade test starting from `v1.3.0`.

## Test plan

- [x] `make upgrade-test testFilter=Test_SubaccountTrustConfiguration_External_Name UPGRADE_TEST_FROM_TAG=v1.3.0 UPGRADE_TEST_TO_TAG=local`


sdischer-sap (approved):
Nice, lgtm

PR #828, opened 2026-07-10
fix: schema-driven drift detection to prevent circuit breaker (#682)

### LINKED ISSUE #682: [BUG] Circuit Breaker hits after BTP Cockpit update
**Describe the bug**
When doing a manual update in the BTP Cockpit the provider attempts to continously update the KymaEnvironment until the circuit breaker sets it in an unhealthy state. 

After some investigation in an internal support issue we found that those attributes are added by (empty) Cockpit updates.

```
...
"accessControlList": {},
"gvisor": {"enabled": false},
"ingressFiltering": false
...
```

v1.8.0

**To Reproduce**
1. Create `KymaEnvironment` with provider
2. Go to BTP cockpit and apply a manual fix (don't add any changes)
3. Provider detects a drift and attempts to update to its spec several times
4. `KymaEnvironemnt` ends up in unhealthy state eventually

**Expected behavior**
If no real change is supplied in the BTP Cockpit, the provider should not attempt to update the resource

**Implementation Details**
In the drift detection we need to make sure, that default values of fields are not causing a drift. This can probably be achieved with doing a 

**Additional contexnt**
**Workaround:** Add those 3 attributes to your spec. It will ensure no false drift is detected

justin-luong:
## What

Reworks `KymaEnvironment` drift detection to compare against the BTP `updateSchema` instead of doing a naive equality check between `spec.forProvider.parameters` and `status.atProvider.parameters`.

Fixes #682.

## Why

The controller compared desired (spec) against current (BTP-reported) parameters with a plain `cmp.Diff`. Two properties of the BTP Kyma provisioning schemas make that comparison produce false drift, which never resolves and eventually trips the update circuit breaker, leaving the resource `Synced=False`.

**Failure mode 1 — BTP materialises `updateSchema` defaults on update.** The Kyma `updateSchema` declares defaults for `ingressFiltering` (`false`), `gvisor.enabled` (`false`), and an effectively-empty `accessControlList` (`{}`). Any update — including an "empty" update from the BTP Cockpit — causes BTP to stamp these into the stored parameters. They then appear in current but never in the user's spec, so the diff reports drift forever. The provider updates to remove them, BTP re-materialises them, and the loop repeats until the circuit breaker trips. This is the reported symptom.

**Failure mode 2 — create-only fields aren't in the update contract.** `region`, `networking`, `modules`, and `colocateControlPlane` exist in `createSchema` but not `updateSchema`. They are valid (often required) in spec at creation time, but the update API does not accept them. Sending them on update makes BTP reject or ignore the request, and comparing them for drift produces false positives.

## How

Fetch the BTP `updateSchema` for the resource's `(environmentType, planName)` and use it as the contract for both drift detection and update payloads:

- **Fetch + cache (`schema.go`):** `SchemaFetcher` pulls `availableEnvironments` and caches the parsed `updateSchema` with a 24h TTL. The cache is created once per controller and shared across all reconciles (previously it was rebuilt each reconcile and never persisted). It is keyed by `(region, environmentType, planName)` — the `updateSchema` structure is constant across subaccounts within a region but region-specific, and one provider reconciles across regions. On fetch failure with a warm cache the cached copy is used; with a cold cache an error is returned.
- **Set `Authorization("")` on the `availableEnvironments` fetch:** the generated provisioning client requires the `Authorization` header parameter (a nil value fails with `authorization is required and must be specified` before any request is sent); the real bearer token is injected by the oauth2 HTTP transport. This matches the other `ProvisioningServiceClient` calls (e.g. `GetEnvironmentInstances`).
- **Restrict the diff surface (`diff.go`):** `DiffAgainstUpdateSchema` compares only keys present in the `updateSchema`, so create-only fields can't drift (failure mode 2).
- **Normalise schema defaults:** for a key the user didn't set, an observed value equal to the schema's effective default is treated as absent (failure mode 1). A field the user does set keeps full drift detection, and a non-default observed value (e.g. `ingressFiltering: true`) is still real drift.
- **Keep the circuit breaker consistent with the diff surface:** `updateCircuitBreakerStatus` now hashes the same schema-normalised maps the diff is computed from, not the raw `desired`/`current`. Otherwise BTP-materialised defaults (which are dropped from the diff but remain in the raw maps) churn the hashes while the real drift is stable, resetting the retry counter every reconcile so the breaker never trips on genuine, non-converging drift.
- **Filter update payloads:** `FilterToUpdateSchema` drops create-only fields from the outgoing update payload so the provider never sends fields the update API rejects.
- **Fail closed:** both `needsUpdateWithDiff` and `UpdateInstance` return an error if the schema can't be fetched, rather than falling back to the naive comparison that caused the bug. `Observe` runs before `Update` and primes the cache. On a drift-check error, `Observe` returns no observation so the reconciler requeues, rather than reporting a possibly-stale up-to-date status.

## Why schema-driven rather than a hard-coded field list

The reported symptom involves three specific fields, and a spec workaround exists (mirror them into the spec). But hard-coding them breaks the next time BTP adds a defaulted field, and doesn't address failure mode 2. The schemas already describe exactly which fields are updatable and their defaults, so deferring to them is robust: if BTP stops materialising a default the normalisation becomes a no-op; if it changes the schema the next cache refresh picks it up; if it adds a new defaulted field it's handled with no code change.

## Reproduction & verification

The reported **symptom** was reproduced end-to-end against a live MCP-managed `KymaEnvironment` (azure / westeurope) before writing the fix:

- Healthy baseline: `Synced=True`, `atProvider.parameters` had none of the three default fields.
- Enriched the spec with `ingressFiltering`/`gvisor` → provider updated → BTP persisted them into stored parameters.
- Reverted the spec → provider detected drift on the now-materialised fields, `updateRetryStatus.count` climbed 1→2→3, `circuitBreaker: true`, `Synced=False` with `"update failed: circuit breaker is on"`.
- Applied the reporter's workaround → recovered to `Synced=True`.

The **fix** is covered by:

- **Unit tests** encoding this scenario (`BugRepro_DefaultsMaterialisedInCurrent`) plus real-drift, create-only, workaround-still-works, empty/nil-schema, and payload-filter cases.
- A **regression test** (`TestObserve_CircuitBreakerTripsOnPersistentSchemaDrift`) that pins the circuit-breaker/diff consistency: it fails against the pre-fix raw-map hashing and passes with the normalised hashing.
- A **long-e2e assess step** (`Verify materialised defaults do not trip the circuit breaker (#682)`) that forces an update round-trip against real BTP, then watches `updateRetryStatus` over a sustained window and asserts the breaker stays off and the resource stays `Available`. The long e2e now passes on this branch.

## Notable decisions for review

- **Create-only mismatches are now invisible to drift detection.** Dropping `region`/`networking`/`modules`/`colocateControlPlane` from both sides of the diff means a spec-vs-observed mismatch on these is no longer surfaced. This is intended: they're immutable post-create and the update API can't reconcile them, so surfacing them only ever produced an unfixable loop.
- **`CreateInstance` is intentionally unfiltered while `UpdateInstance` filters** — create needs the full `createSchema` surface, update must not send create-only fields.
- **Unusable schema (nil or empty properties) degrades to the pre-fix naive diff** rather than suppressing all drift, so a malformed/empty schema can never silently disable updates.
- **Fixed an in-passing latent bug in `UpdateInstance`:** the `UnmarshalRawParameters` error was checked after `AddKymaDefaultParameters` wrote to the (possibly nil) map; reordered the check.
- **Cache lifetime:** the schema cache is long-lived (per-controller) and region-scoped, so it survives reconciles; the per-reconcile BTP client is bound to it only for the fetch, avoiding stale-credential concerns. (Addresses @maximilianbraun's review comment.)

## Follow-ups (out of scope)

- The Kyma `accessControlList` schema declares `required: [allowedCIDRs]` yet BTP materialises `{}`, which is invalid against its own schema. Handled here via the "empty object = default" rule, but worth reporting upstream to the Kyma broker team. Mentioned by @maximilianbraun in [[BUG] Circuit Breaker hits after BTP Cockpit update #682].
- Schema-driven pre-flight validation of `spec.forProvider` (reject unknown fields / type mismatches / enum violations before any BTP call) is a natural extension once the schema is cached, but is deliberately not in this PR.
- No `UpdateInstance` payload-capture integration test — the filtering logic is unit-tested via `FilterToUpdateSchema`, and the e2e drift-probe exercises the update path end-to-end, but the provisioning fake's update path isn't wired to capture the exact outgoing payload.

## Testing

`go build ./...`, `go test ./...`, `go vet`, `gofmt`, `make lint` — all clean. Long e2e (`TestKymaEnvironment`) passes on this branch, including the new circuit-breaker drift-probe assess step.

Refs #682 · schema analysis in #682 (comment)


internal/clients/kymaenvironment/kymaenvironment.go
```diff
@@ -97,10 +107,26 @@ func (c KymaEnvironments) UpdateInstance(ctx context.Context, cr v1alpha1.KymaEn
...
 		return err
 	}
+	parameters = AddKymaDefaultParameters(parameters, GetKymaEnvironmentName(cr), string(cr.UID))
+
+	// Drop create-only fields (region, networking, modules, ...) the update
+	// API does not accept; sending them makes BTP reject or silently ignore
+	// the update (issue #682). Fail-closed on schema fetch errors so we never
+	// send an unfiltered payload; an unusable schema passes params through
+	// (see FilterToUpdateSchema).
+	schema, err := c.schemaFetcher.GetUpdateSchema(
```
maximilianbraun, line 120: Should be cached

n-losse:
@justin-luong are you still working on this one or should we reassign?

justin-luong:
> @justin-luong are you still working on this one or should we reassign?

Didn't see the comments to this PR, I'll keep working on it.

PR #830, opened 2026-07-15
feat(recovery): heal own lost-ID Create()s via semantic lookup gated by external-create-pending

### LINKED ISSUE #767: [BUG] No conflict errors shown in async `ServiceInstance` creations
**Describe the bug**
The service instance resource doesnt set the external name correctly, we expect to see a UUID after some time that reprsents the true id of an external resource from its API, but instead we see metadata.spec.name value of the resource or an empty field. while the correct annotation are set stating it is created!

Release candidate from main - docker pull ghcr.io/sap/crossplane-provider-btp/crossplane/provider-btp:v0.0.0-50c457265f27d5f6bab1fe063f7be4c9ba5a003e

**To Reproduce**
??

**Expected behavior**
After some time we expect the external name to have the correct external to be shown. I

**Screenshots**

Image

dattito:
Closes #767 

## What

Always-on heal that **recovers the provider's own lost-ID `Create()` attempts** — the failure modes where BTP successfully created a resource but the CR is left with only a fallback external-name (empty, or `== metadata.name`). Without the heal, such resources are orphaned: `Observe`/`Update` can never resolve them, and on **delete** the finalizer is stripped while the BTP resource is left behind (see #427).

## Package: `internal/recovery` (renamed from `adoption`)

The external-name ADR uses "adoption" for the brownfield / import case this feature explicitly **refuses**. Reusing the same word for the recovery-from-lost-ID case invited confusion, so the package (and all identifiers / event reasons) is now `recovery`.

## What separates recovery from import — the `external-create-pending` ownership check

Design follows [@jtucci's comment on #813](https://github.com/SAP/crossplane-provider-btp/discussions/813#discussioncomment-17566668): scope this to case 1 (recovery of our own lost-ID Create) and keep case 2 (import / brownfield / DR) explicit per the [external-name ADR](https://github.com/SAP/crossplane-provider-btp/blob/main/docs/contribution-notes/external-name-handling.md).

Each semantic-lookup match returns the BTP-reported `created_at`. Adoption only proceeds when the resource was plausibly born inside the window where our own `Create()` attempt could have produced it. The reference point is the `crossplane.io/external-create-pending` annotation that crossplane-runtime writes right before every Create (Create is aborted if the write fails, so a non-zero pending timestamp is a reliable "we attempted Create" signal; a missing one means we did not).

**Bounded window** — `[pending - 60s, pending + 5min]`:
- Lower bound: BTP Creates return in seconds; anything born meaningfully before we attempted Create can't be ours. The ~60s slack absorbs NTP drift between K8s and BTP.
- Upper bound: BTP Creates return in seconds. Anything born much later can't be the result of our attempt — probably a UI-created same-name resource that appeared later.

**No fallback to the CR's `creationTimestamp`**. If the pending annotation is missing, this controller never attempted a Create for this CR, so the heal short-circuits **before** running the expensive semantic lookup and refuses to adopt. This closes the "same-name resource that shows up any time after the CR" hole.

Matches outside the window cause a `Warning:RecoveryRefusedBrownfield` event and leave `crossplane.io/external-name` untouched, routing the user to the ADR-documented import flow. Users who follow the documented import flow (`managementPolicies: [Observe]`) never touch this feature — under `Observe` the reconciler never attempts a Create, so `external-create-pending` is never stamped and the heal short-circuits.

## Which failure modes are actually reachable

Verified against crossplane-runtime v2.2.0 (`reconciler.go`): if `external-create-pending` is set and `external-create-succeeded` is not, and the CR's external-name is non-deterministic (all BTP identifiers are random GUIDs), the reconciler **refuses to run `Observe`** and freezes the CR with a "cannot determine creation result" error. The heal lives in `Observe`, so a pod crash *during* Create is not reachable via this path.

The reachable states are narrower:

- **SM / CM two-phase create**. The operator writes the instance UUID as external-name after phase-1 and only upgrades to the compound `"<sID>/<bID>"` after phase-2. If phase-2's Create is interrupted, the CR has a single-UUID external-name (which is **not** a fallback per `IsFallbackExternalName`, so the heal doesn't fire — the two-phase loop resumes on its own). But if the operator was torn down and recreated between the two phases, or an early failure prevented phase-1's UUID from ever being persisted, the CR can end up with a fallback external-name while phase-1 already exists in BTP.
- **Hand-written SI / SB paths** that persist status/spec between BTP Create and the external-name update, and crash in between (rare, but seen in the wild — BTP resource exists, CR has no ID).

In every reachable case `external-create-pending` was written before the Create that succeeded, so the ownership check has its reference.

## How

- The heal runs in `Observe()` only when the underlying client reports the resource as non-existent **and** the external-name is a true fallback (`""` or `metadata.name`). Trigger is strictly `recovery.IsFallbackExternalName`; a non-compound single-UUID external-name (SM/CM phase-1 output) is deliberately **not** treated as heal-eligible, or two-phase Create would deadlock.
- Before running the semantic lookup, `recovery.HasCreateBeenAttempted(cr)` short-circuits when there is no `external-create-pending` annotation. Otherwise the lookup runs and the result is gated on `recovery.IsOwnedByCR`.
- On a **unique** match that passes `IsOwnedByCR`, the controller patches `crossplane.io/external-name`, clears stale async-failure conditions (SI only), emits `Normal:ExternalNameRecovered`, and returns `recovery.ErrRequeueAfterRecovery` to force a requeue — the only signal that is safe on the delete leg (no finalizer strip, no stale-name delete) and rebuilds the external client against the recovered name.
- SI/SB/SM/CM get SM credentials from a **subaccount-admin binding minted via the accounts-service** (`EnsureSemanticLookuper`, reuse-or-mint + cleanup). The per-resource `serviceManagerSecret` bindings are platform-scoped and do **not** list instances created by the btp terraform provider.
- Cleanup uses `context.WithoutCancel` so a reconcile timeout does not silently orphan the temporary admin binding minted for the lookup; delete errors are logged.
- Instance lookup keys on `(plan, instanceName)` and binding lookup on the managed binding name, so the subaccount-admin plan's access instance and any transient admin binding are never adopted.
- ServiceBinding lookup handles rotation naming (`<name>-<random>`) by using `status.AtProvider.Name` when set, falling back to `spec.forProvider.Name` otherwise.
- `>1` match refuses to adopt (`Warning:RecoveryAmbiguous`); ownership-check failures emit `Warning:RecoveryRefusedBrownfield`; lookup-call failures emit `Warning:RecoveryLookupFailed` and fall through (retry next reconcile).
- SI heal ordering: **clear async-failure conditions FIRST**, then persist the recovered external-name. Under the reversed order a partial failure (status write fails after external-name write succeeds) permanently stranded the CR with a real name and stale conditions. Both partial-failure orderings now self-recover on the next reconcile.

## Safety

- **Ownership check gates every recovery**. The pending-annotation-only reference makes the window bounded on both sides; without evidence that we attempted Create, no recovery. A same-name resource that shows up later cannot be pulled in.
- User-typed (non-fallback) external-names are never overwritten.
- Refuses to guess on ambiguous matches.
- Trigger scoped to `IsFallbackExternalName` — single-UUID external-names (natural SM/CM phase-1 output) are left alone so the operator's two-phase Create can complete phase-2.

## Testing

- `internal/recovery/recovery_test.go`:
  - `TestIsOwnedByCR_PendingWindow` — bounded window semantics: at attempt, +3s / +4m / +5m (in), +6m (out), −30s / −60s (in), −61s (out).
  - `TestIsOwnedByCR_RequiresPendingAnnotation` — critical safety: without `external-create-pending`, all btpCreatedAt values must be refused. Guards the removed-fallback behavior.
  - `TestHasCreateBeenAttempted`.
  - `TestIsFallbackExternalName` — including the single-UUID (SM/CM phase-1 output) regression case.
- Per-kind heal test files for **all five kinds** — subaccount, serviceinstance, servicebinding, servicemanager, cloudmanagement. Each covers: happy-path recover+requeue, brownfield refusal (btp outside window), **no-create-pending short-circuit** (guards the safety property), no match, lookup error, plus kind-specific edge cases (SB rotation name / empty `serviceInstanceID`; SM/CM single-UUID phase-1 output).
- Full suite (~1500 tests) green; `golangci-lint run ./...` = 0 issues; gofmt clean on touched files.
- Live-tested on the dev landscape (heuristic-curie MCP, `agentic-runtime-prov-canary` global account): fresh CloudManagement provisioned to Ready via a provider built from this commit; `external-name` transitions cleanly through the two-phase compound key; `external-create-pending` stamped as expected. Standing chain (Subaccount + Entitlement + ServiceManager) intact.

## Notes for reviewers

- Shared helpers: `internal/recovery` (`IsFallbackExternalName`, `HasCreateBeenAttempted`, `IsOwnedByCR`, event reasons, sentinel `ErrRequeueAfterRecovery`) and `SemanticLookuper` in `internal/clients/servicemanager`.
- Removed: `--adopt-orphaned-external-names` flag, `BTP_ADOPT_ORPHANED_EXTERNAL_NAMES` env var, `CrossplaneOptions.AdoptOrphanedExternalNames`, and all `adoptOrphans` plumbing across the five controllers. The ownership check makes the feature behavior-neutral on healthy systems, so no opt-in gate is needed.
- Removed: unused `di.NewSemanticLookuperFn` (leftover from an earlier iteration; the used path is `EnsureSemanticLookuper` via the accounts-service).
- Interface changes: `SemanticLookuper.{LookupServiceInstance,LookupServiceBinding,LookupInstanceAndBinding}` and `AccountsApiAccessor.SubaccountGuidBySubdomain` now return the BTP-reported `created_at` alongside the ID. `recovery.IsOwnedByCR` takes the CR object (`metav1.Object`) rather than a bare timestamp so it can read the pending annotation.

Refs: #813 (design discussion), #427 (generic external-name handling).


jtucci:
Hey @dattito have you taken a look at this discussion https://github.com/SAP/crossplane-provider-btp/discussions/813 ? 
I've been doing a lot of work and investigation around this topic over the last few months and I'm not sure a provider wide flag is the best approach since the external-name ADR is built around per-resource intent, and a global flag plus spec-matching adoption means a fresh MR can silently adopt an unrelated existing resource with no ownership check. Would be good to align on the concept in #813 first and happy to discuss there :)

internal/adoption/adoption.go
```diff
@@ -0,0 +1,175 @@
...
+// The heal path runs during Observe(): when the underlying client reports
+// the resource as non-existent and the external-name is a fallback, it
+// performs a semantic lookup against the BTP APIs and, on a unique match
+// that also passes the ownership check (IsOwnedByCR), adopts the resource
+// by patching crossplane.io/external-name with the real identifier.
+//
+// The ownership check is what keeps this a strict bug-fix rather than a
+// general import mechanism: adoption only proceeds when the BTP resource
+// was plausibly created by our own Create() attempt. See IsOwnedByCR.
+package adoption
```
jtucci: More of a nit, but the word `adoption` to me reads more like the import scenario vs the recovery / heal scenario.

internal/adoption/adoption.go
```diff
@@ -0,0 +1,175 @@
...
+// fall back to the CR's creationTimestamp, which is always a valid lower
+// bound because Create() is only ever issued after the CR exists.
+//
+// The two branches have different upper-bound semantics; see IsOwnedByCR.
+func ReferenceTime(cr metav1.Object) (t time.Time, fromPendingAnnotation bool) {
+	if p := meta.GetExternalCreatePending(cr); !p.IsZero() {
+		return p, true
+	}
+	return cr.GetCreationTimestamp().Time, false
+}
```
jtucci: Im not sure the CreationTimestamp fallback is needed, and it might actually open up the window of adopting a resource that wasn't created by this controller, since there is no upper bound on it (if the CR never got as far as Create, say a subaccount ref won't resolve, and a day later someone creates the same-named resource in the UI, wouldn't this branch then pull it in?).
The runtime writes external-create-pending before every Create and aborts if that write fails, so I tihnk a missing annotation means we never created anything. I don't think the delayed-Create case needs it either: once the Create runs, the annotation is there and the bounded branch takes over. And the ServiceManager torn-down-and-recreated case isn't served by this path anyway since the old instance predates the new CR, so the brownfield check refuses it.
Maybe I'm missing something though?

internal/controller/account/servicebinding/servicebinding.go
```diff
@@ -296,6 +323,80 @@ func (e *external) Delete(ctx context.Context, mg resource.Managed) (managed.Ext
...
+// next reconcile (after the parent heals) succeeds.
+//
+// Return contract matches the ServiceInstance heal: ErrRequeueAfterAdopt on a
+// successful adoption, nil when there is nothing to adopt or the lookup failed,
+// a real error only when persisting fails.
+func (e *external) healExternalName(ctx context.Context, cr *v1alpha1.ServiceBinding) error {
+	if e.newAdminLookuperFn == nil {
+		return nil
+	}
+	serviceInstanceID := internal.Val(cr.Spec.ForProvider.ServiceInstanceID)
```
jtucci, line 337: Should this get an adopt test file like the other kinds? serviceinstance, cloudmanagement and subaccount all have one so it might be good to cover the same cases here (adopt + requeue, brownfield refusal, no match, lookup error), plus the SB-specific bits, empty serviceInstanceID and the rotation fallback.

internal/controller/account/servicemanager/servicemanager.go
```diff
@@ -140,9 +159,118 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
...
+// successful adoption, nil when there is nothing to adopt or the lookup failed,
+// a real error only when persisting fails.
+func (c *external) healExternalName(ctx context.Context, cr *apisv1beta1.ServiceManager) error {
+	if c.newAdminLookuperFn == nil {
+		return nil
+	}
+	if cr.Status.AtProvider.DataSourceLookup == nil {
+		return nil
+	}
+	planID := cr.Status.AtProvider.DataSourceLookup.ServiceManagerPlanID
```
jtucci, line 187: Same here, should this mirror cloudmanagement_adopt_test.go?

internal/controller/account/serviceinstance/serviceinstance.go
```diff
@@ -277,6 +317,92 @@ func (c *external) Delete(ctx context.Context, mg resource.Managed) (managed.Ext
...
+	meta.SetExternalName(cr, guid)
+	if uErr := e.kube.Update(ctx, cr); uErr != nil {
+		return errors.Wrap(uErr, "cannot persist adopted external-name")
+	}
+	clearAsyncFailureConditions(cr)
+	if uErr := e.kube.Status().Update(ctx, cr); uErr != nil {
+		return errors.Wrap(uErr, "cannot clear async-failure conditions after adoption")
+	}
+
+	log.FromContext(ctx).Info("adopted existing BTP service instance by external-name", "guid", guid, "name", name)
```
jtucci: What happens if the Update at 372 succeeds but this status write fails? Next reconcile the conflict condition is still set, external-name is now a real UUID so IsFallbackExternalName is false, heal skipped so the branch returns the conflict error forever. Only a spec bump escapes. Maybe swap the order (clear condition first, then persist the name) then both partial failures recover on their own?

internal/clients/servicemanager/service_instance_proxy_client.go
```diff
@@ -89,6 +89,59 @@ func (t ServiceManagerInstanceProxyClient) describeAdminBinding(ctx context.Cont
...
+		binding, err = t.createAdminBinding(ctx, subaccountGuid)
+		if err != nil {
+			return nil, noop, err
+		}
+		cleanup = func() {
+			_ = t.deleteAdminBinding(ctx, subaccountGuid)
+		}
+	}
+
+	cl, err := NewServiceManagerClient(ctx, binding)
```
jtucci, line 147: since all callers run this via defer, and the reconcile ctx has a 1-minute timeout, a slow lookup that blows the budget cancels the ctx. This delete fails instantly, and the temp binding leaks silently and permanently (the reuse path never deletes it). Maybe context.WithoutCancel + log the error instead of _ =?

internal/adoption/adoption.go
```diff
@@ -0,0 +1,175 @@
+// Package adoption contains shared helpers for the "orphaned external-name
+// adoption" heal path used by the hand-written BTP resource controllers
+// (Subaccount, ServiceInstance, ServiceBinding, ServiceManager,
+// CloudManagement).
+//
+// Background: under certain async-create failure modes (pod restart
+// mid-create, two reconcilers racing a create, or a ServiceManager being
+// torn down and recreated) a managed resource can end up with only a
+// fallback external-name (empty, or equal to metadata.name — Crossplane's
+// default initializer value) even though BTP successfully created the
```
jtucci: I don't think the pod crash mid-create case is actually reachable by this heal. A crash between the create-pending and create-succeeded annotations makes the runtime freeze the resource ("cannot determine creation result") before Observe runs, and the heal lives in Observe. Might be worth changing the motivation text to whats actually reachable.

internal/di/creatorfns.go
```diff
@@ -16,3 +16,16 @@ func NewPlanIdResolverFn(ctx context.Context, secretData map[string][]byte) (ser
...
+
+// NewSemanticLookuperFn builds a servicemanager.SemanticLookuper from a BTP
+// service operator (service manager binding) secret's data. It is used by the
+// orphaned-external-name adoption heal path. The returned *ServiceManagerClient
+// is scoped to the subaccount the binding belongs to, so semantic lookups are
+// implicitly constrained to that subaccount.
+func NewSemanticLookuperFn(ctx context.Context, secretData map[string][]byte) (servicemanager.SemanticLookuper, error) {
+	binding, err := servicemanager.NewCredsFromOperatorSecret(secretData)
+	if err != nil {
+		return nil, err
```
jtucci: Is this being called anywhere? maybe leftover from a previous iteration?

dattito:
thanks for the great input @jtucci. All good concerns, I implemented/checked them.

internal/di/creatorfns.go
dattito: nope, leftover

schlingemann (dismissed):

jn-av (changes requested):
The implementation looks good, but could you please shorten some of the comments? That would drastically increase readability.

On another note, I would lean towards adding this as a feature flag. However, if you prefer it as a mainline feature, that is also fine. In that case, we should either update the ADR or document where we are deviating from it. That would be a valuable addition if we decide to go with the mainline approach.

dattito:
Hey @jn-av , thanks for the input. I adjusted the comments to be less noisy. Also @jtucci , @schlingemann and I discussed the topic and decided to _not_ guard it behind a feature flag, as this is only a bug fix that fixes _ already expected_ behaviour and does not introduce new one  (please see the [comment on #813 for this](https://github.com/SAP/crossplane-provider-btp/discussions/813#discussioncomment-17566668), we made sure it's only the bugfix (1), not the importing feature (2) by introducing the "bounded window").

jn-av (approved):
Thanks for the changes way better.

PR #834, opened 2026-07-16
Feat/issue 550 cloudmanagement external name adr

### LINKED ISSUE #550: [TASK] `External-Name`(CloudManagement): Ensure ADR compliance
**Part of Epic: #427** 
## Context

## Technical Tasks Checklist

**Acceptance Criteria:**

- [ ] **Migration Logic**:

**Acceptance Criteria:**
- [ ] **Observe() Tests**:

- [ ] **Create() Tests**:

- [ ] **Update() Tests**:

- [ ] **Delete() Tests**:

**Acceptance Criteria:**
- [ ] **Import Flow Test**:

### 5. Documentation

n-losse:
Closes #550

## What

Makes `CloudManagement` compliant with the external-name ADR (`docs/contribution-notes/external-name-handling.md`).

CloudManagement is a compound resource — it manages a `SubaccountServiceInstance` + `SubaccountServiceBinding` under the hood, with external-name format `<instanceID>/<bindingID>`.

### `zz_setup.go` — suppress default external-name initializer
Switch from `DefaultSetup` to `DefaultSetupWithoutDefaultInitializer` so crossplane-runtime no longer writes `metadata.name` into the annotation. An empty annotation now correctly signals "resource does not exist yet".

### `Observe()` — ADR Step 1 intentionally omitted
The ADR says to return `ResourceExists: false` immediately for an empty external-name. We deliberately skip that early return here so that `IsFallbackExternalName` (which covers `""`) can feed into the recovery path when `ObserveResources` returns `!ResourceExists`. Returning early would make the `""` case unreachable by recovery.

### `Observe()` — format validation (ADR Step 2)
Add `validateExternalName` helper that accepts:
- A bare UUID: instance created in phase 1 of `CreateResources`, binding not yet written (this is a real, reachable intermediate state — `CreateResources` is a two-phase create)
- A `<uuid>/<uuid>` compound key: both phases complete

Validation is skipped for fallback external-names (empty or `== metadata.name`) so recovery gets a chance to run first. Any other value returns an error so the user is alerted rather than a silent no-op observe.

### `cloudmanagement_types.go` — doc comment
Add the standard `// External-Name Configuration:` block above the `CloudManagement` struct for `make docs.generate-external-name`.

### Unit tests
Add `EmptyExternalName` and `InvalidExternalNameFormat` test cases to `TestObserve`. The `EmptyExternalName` case exercises the full observe path (including `setStatus`) rather than an early return. Update existing cases to use valid UUIDs as external-name (the new validation correctly rejects the prior `"test"` placeholder).

### Upgrade test
Add `Test_CloudManagement_External_Name` verifying the `ensureCompatibility()` migration from bare `metadata.name` to `<instanceUUID>/<bindingUUID>` when upgrading from `v1.11.0` (last release using `DefaultSetup`).

## What was already compliant (no changes needed)

- `Create()` — sets external-name on success, does not set on error
- `Update()` — immutable fields protected by `kubebuilder:validation:XValidation`
- `Delete()` — Upjet-backed connectors treat 404 as non-error
- `ensureCompatibility()` — migration logic already correct


internal/controller/account/cloudmanagement/cloudmanagement.go
```diff
@@ -199,6 +199,22 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
...
+	// ADR(external-name) Observe Step 2: validate external-name format, but
+	// only when it is a real BTP identifier. A fallback external-name (==
+	// metadata.name) is handled by the recovery path below and must not be
+	// rejected here before recovery gets a chance to run.
+	extName := meta.GetExternalName(cr)
+	if !recovery.IsFallbackExternalName(cr.Name, extName) {
+		if err := validateExternalName(extName); err != nil {
+			return managed.ExternalObservation{}, err
+		}
+	}
```
jtucci, line 215: SM wraps this same check in !meta.WasDeleted so that malformed annotations cant strand finalziers on delete. Was skipping that here intentional?

internal/controller/account/cloudmanagement/cloudmanagement.go
```diff
@@ -467,3 +483,16 @@ func unmarshalContext(src *string) *map[string]string {
...
+func validateExternalName(s string) error {
+	if internal.IsValidUUID(s) {
+		return nil
+	}
+	parts := strings.SplitN(s, "/", 2)
+	if len(parts) == 2 && internal.IsValidUUID(parts[0]) && internal.IsValidUUID(parts[1]) {
+		return nil
+	}
+	return fmt.Errorf("invalid external-name %q: must be a UUID or <uuid>/<uuid>", s)
+}
```
jtucci: Since the CM key is the same SM instance/binding format, have you considered using servicemanager.ValidateExternalName here ? Right now, both of these functions accept different keys for the same format.

jtucci (approved):
I think the recovery path looks right to me, dropping step 1 and letting fallback names through is the only way healExternalName could run and it mirrors what SM already does.

internal/controller/account/cloudmanagement/cloudmanagement.go
n-losse: ah, that SM change is newer than my branch - good find! and no, i'll add that check

internal/controller/account/cloudmanagement/cloudmanagement.go
n-losse: good find, using that function now

xdj2e9 (approved):
LGTM

PR #835, opened 2026-07-17
fix(servicemanager,cloudmanagement): don't orphan the service instance when deleting with a missing binding

dattito:
## What

Fixes a delete-path orphan in **both** the **ServiceManager** and **CloudManagement** controllers: the Kubernetes CR is finalized and removed while its service instance is left stranded in BTP. Matches the observed incident — *CR gone from K8s, service binding gone, service instance still in BTP*.

Both controllers are structurally parallel: they provision a **service instance + binding** (two-phase create) and store them as a compound external-name `"<serviceInstanceID>/<serviceBindingID>"`, so both had the identical bug and get the identical fix.

## Root cause

The cis (`CloudManagement`) and servicemanager (`ServiceManager`) tf clients' `ObserveResources` reported `ResourceExists:false` whenever the **binding** was missing, so the managed reconciler would drive phase-2 (`createBinding`). Correct while creating — but it was applied unconditionally, **including while the CR is being deleted**.

The crossplane managed reconciler only calls `Delete()` while the observed resource still exists (`reconciler.go`: `if observation.ResourceExists && policy.ShouldDelete()`). Under a deletion timestamp, `ResourceExists:false` makes it skip `Delete()`, unpublish, strip the finalizer and remove the CR. So:

1. `DeleteResources()` deletes the **binding first**, then the instance.
2. If the instance delete errors/lags (async terraform destroy), the reconcile requeues.
3. The next `Observe` sees the binding already gone → reports `ResourceExists:false`.
4. The reconciler treats the resource as already deleted, strips the finalizer, and the CR disappears **while the service instance remains in BTP**.

Intermittent in the wild because most deletes remove binding + instance in the same reconcile; the orphan only occurs when the instance delete doesn't complete in the pass that removed the binding.

## Fix

Make `ObserveResources` delete-aware in both tf clients: when the CR has a deletion timestamp, key `ResourceExists` off the **instance alone**. BTP does not allow deleting a service instance while a binding references it, so the binding is necessarily gone once the instance is gone — the instance is the authoritative "is anything left in BTP" signal during deletion. This keeps `Delete()` firing until the instance is actually gone; only then does `Observe` report false and let the finalizer drop. The non-deleting two-phase-create gating is unchanged.

```go
if meta.WasDeleted(cr) {
    status := ResourcesStatus{
        ExternalObservation: managed.ExternalObservation{
            ResourceExists:   siObs.ResourceExists, // instance-only
            ResourceUpToDate: true,
        },
    }
    if siObs.ResourceExists {
        status.Instance /* or InstanceID */ = ...
    }
    return status, nil
}
```

## Mid-delete status reporting

With the delete-aware `Observe` returning `ResourceExists:true` while the instance still exists mid-delete, the controllers' `setStatus` would previously key off that bool alone and mark the CR `Available` / `Bound` between the `Delete()` call (which set `Deleting()`) and the reconcile that finally observes the instance gone. Purely cosmetic — finalize is gated on `ExternalObservation.ResourceExists`, not the condition — but misleading in `kubectl get` output and dashboards while a delete is in flight.

`setStatus` in both controllers now branches on `meta.WasDeleted(cr) && status.ResourceExists` → `Deleting()` / `Unbound`, and reports `Available` / `Bound` only when not deleting. Under 5 added lines per controller, no struct/API change.

## Testing

**Reproduced and verified end-to-end against real BTP on a kind cluster** (provider run out-of-cluster), for **both** CloudManagement and ServiceManager:

| | Before fix (buggy) | After fix |
|---|---|---|
| `Observe` during delete (binding gone, instance present) | `ResourceExists:false` | `ResourceExists:true` |
| `Delete()` external calls | **0** | **2** ("Successfully requested deletion") |
| CR finalized | ✅ immediately | ✅ after delete completes |
| service instance in BTP afterwards | **orphaned** ❌ | **deleted** ✅ (verified via `btp list services/instance`) |

Deterministic repro for each controller: get the resource to `Ready`, delete its binding out-of-band via the `btp` CLI, then delete the CR — forcing exactly the "binding gone, instance present, mid-deletion" edge state. Confirmed instance/binding presence at each step with `btp list services/instance|binding`; confirmed via BTP CLI that the instance is deleted (not just the CR).

**Unit regression tests** added:

- `cis_tf_client_test.go` and `service_manager_tf_client_test.go` — delete-path scenarios:
  - `DeletingInstanceExistsBindingGone` — instance exists, binding gone → `ResourceExists:true`.
  - `DeletingInstanceGoneFinalizes` — instance gone → `ResourceExists:false`.
  - (cis) `DeletingInstanceExistsKeepsDeleting` — instance exists, binding present → keeps deleting.

  The binding's fake `Observe` returns an error in the delete cases to prove the binding is **not** consulted on the delete path.

- `internal/controller/account/{servicemanager,cloudmanagement}/*_test.go` — new `DeletingInstanceStillExists` case in `TestObserve`:
  - Asserts the CR ends up with `Deleting()` + `Unbound` (not `Available()` + `Bound`) when a delete is in flight and the instance still exists.

All new tests **fail against the pre-fix code** (guard integrity verified) and pass against the fixed code.

`go build ./...`, `go test ./...`, and `golangci-lint run ./...` all clean.

## Notes for reviewers

- Distinct from #830 (semantic-lookup healing of orphaned external-names). That heals a CR that still exists with a fallback external-name; this fixes the delete path so the CR isn't finalized while the instance still exists. Different root cause, isolated on its own branch.
- Only the two tf clients + the two controllers changed:
  - `internal/clients/cis/cis_tf_client.go`
  - `internal/clients/servicemanager/service_manager_tf_client.go`
  - `internal/controller/account/cloudmanagement/cloudmanagement.go`
  - `internal/controller/account/servicemanager/servicemanager.go`
  - (plus their tests)

### DIFF
--- a/internal/clients/cis/cis_tf_client.go
+++ b/internal/clients/cis/cis_tf_client.go
@@ -186,6 +186,40 @@ func (tf *TfClient) ObserveResources(ctx context.Context, cr *apisv1beta1.CloudM
 	if err != nil {
 		return ResourcesStatus{}, err
 	}
+
+	// Deletion path: key ResourceExists off the INSTANCE alone.
+	//
+	// BTP invariant: a service instance cannot be deleted while a service
+	// binding still references it, so the binding is necessarily gone once the
+	// instance is gone. The instance's existence is therefore the
+	// authoritative "is there anything left in BTP" signal during deletion.
+	//
+	// Why this matters: the crossplane managed reconciler only calls Delete()
+	// while the observed resource still exists (reconciler.go: `if
+	// observation.ResourceExists && policy.ShouldDelete()`). If we report
+	// ResourceExists:false while the CR is being deleted, it skips Delete(),
+	// strips the finalizer and removes the CR — orphaning whatever is still in
+	// BTP. That is exactly the observed orphan incident: DeleteResources()
+	// deletes the binding FIRST and then the instance; if the instance delete
+	// errors or lags, the next Observe sees the binding gone, and the old
+	// two-phase-create gating below reported ResourceExists:false ("binding
+	// needs creation") which, under a deletion timestamp, finalized the CR and
+	// stranded the instance. Keying off the instance keeps Delete() firing
+	// until the instance is actually gone; only then do we report false and
+	// let the finalizer drop.
+	if meta.WasDeleted(cr) {
+		status := ResourcesStatus{
+			ExternalObservation: managed.ExternalObservation{
+				ResourceExists:   siObs.ResourceExists,
+				ResourceUpToDate: true,
+			},
+		}
+		if siObs.ResourceExists {
+			status.Instance = tf.sInstance.Status.AtProvider
+		}
+		return status, nil
+	}
+
 	if !siObs.ResourceExists {
 		return ResourcesStatus{
 			ExternalObservation: managed.ExternalObservation{ResourceExists: false},
--- a/internal/clients/cis/cis_tf_client_test.go
+++ b/internal/clients/cis/cis_tf_client_test.go
@@ -394,6 +394,94 @@ func TestObserveResources(t *testing.T) {
 				err: nil,
 			},
 		},
+		{
+			// REGRESSION GUARD (orphan bug): CR is being deleted, the instance
+			// still exists in BTP but the binding is already gone (e.g. the
+			// binding delete landed and the instance delete then errored/lagged,
+			// or the binding was removed out-of-band). The OLD code reported
+			// ResourceExists:false here ("binding needs creation"), which under a
+			// deletion timestamp made the managed reconciler skip Delete(), strip
+			// the finalizer and ORPHAN the instance in BTP. We must report
+			// ResourceExists:true so Delete() keeps firing until the instance is
+			// actually gone.
+			//
+			// The binding is intentionally NOT observed on the delete path (BTP
+			// invariant: an instance cannot be deleted while a binding references
+			// it, so instance-gone implies binding-gone). To prove the binding is
+			// never consulted here, its observeFn returns an error: if the code
+			// observed it, this test would surface that error.
+			name: "DeletingInstanceExistsBindingGone",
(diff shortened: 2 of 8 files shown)


schlingemann:
**Minor (cosmetic, non-blocking): CR reports `Available`/`Bound` mid-delete.**

On the delete path when the instance still exists, `ObserveResources` now correctly returns `ResourceExists: true`. But `setStatus` in both `servicemanager.go` and `cloudmanagement.go` branches only on that bool, so during deletion (binding already gone) the CR is marked `Available()` + `ServiceManagerBound`/`CisStatusBound`. `Delete()` does set `Deleting()` on entry, but the next `Observe` overwrites it back to `Available()`. This is purely cosmetic — finalize is gated on `ExternalObservation.ResourceExists`, not the condition — but it's misleading in CR status/dashboards while a delete is in flight.

Suggested fix (make `setStatus` delete-aware via the `cr` it already has):

```go
switch {
case meta.WasDeleted(cr) && status.ResourceExists:
    // Mid-deletion, external instance still present: report Deleting,
    // not Available/Bound. Status-only; finalize is gated on ResourceExists.
    cr.Status.SetConditions(xpv1.Deleting())
    cr.Status.AtProvider.Status = apisv1beta1.ServiceManagerUnbound // CisStatusUnbound for cloudmanagement
case status.ResourceExists:
    cr.Status.SetConditions(xpv1.Available())
    cr.Status.AtProvider.Status = apisv1beta1.ServiceManagerBound
default:
    cr.Status.SetConditions(xpv1.Unavailable())
    cr.Status.AtProvider.Status = apisv1beta1.ServiceManagerUnbound
}
```

~4 lines per controller, no struct/API change. Could be pinned with an assertion in the existing `DeletingInstanceExistsBindingGone` tests. Non-blocking — happy to leave as-is if you'd rather keep the diff minimal.

schlingemann (approved):

PR #836, opened 2026-07-17
fix: support custom IdP origin in CloudFoundryEnvironment initialOrgManagers

### LINKED ISSUE #505: [BUG] CloudFoundryEnvironment: initialOrgManagers hardcodes origin to `sap.ids`
v1.6.1

## Problem

When creating a `CloudFoundryEnvironment`, the `initialOrgManagers` field accepts only email addresses (`[]string`) without the ability to specify an Identity Provider origin per manager. The controller hardcodes the origin to `sap.ids` when adding Org Managers via the Cloud Foundry API:

```go
// internal/clients/cfenvironment/cfenvironment.go
const defaultOrigin = "sap.ids"

for _, managerEmail := range cr.Spec.ForProvider.Managers {
    cloudFoundryClient.addManager(ctx, managerEmail, defaultOrigin)
}
```

This means users from a custom Identity Provider (e.g. a custom IAS tenant) cannot be added as CF Org Managers during environment creation. The CF API call `CreateOrganizationRoleWithUsername` always receives `sap.ids` as the origin, regardless of the user's actual IdP.

Notably, the `User` type defined in `apis/environment/v1alpha1/cfenvironment_types.go` already includes an `Origin` field and is used in the observation/status (`CfEnvironmentObservation.Managers []User`), but the spec uses `[]string` instead of `[]User`.

**To Reproduce**
1. Create a `CloudFoundryEnvironment` CR with a user from a custom IdP in `spec.forProvider.initialOrgManagers`
2. The controller calls the CF API with `origin: sap.ids` instead of the user's actual IdP origin
3. The user is either not found or assigned under the wrong identity

**Expected behavior**
Users from custom Identity Providers should be assignable as CF Org Managers. The `initialOrgManagers` field should support specifying the IdP origin per user, for example by using the existing `User` type (with `username` and `origin` fields) instead of plain strings.

## Impact

In setups where a custom Identity Provider is used, no users from that IdP can be added as CF Org Managers during `CloudFoundryEnvironment` creation.

**Additional context**
The codebase already contains a `User` struct with an `Origin` field and a TODO comment suggesting this should be addressed:
```go
// todo: default origin to "sap.ids", replace this with scim lookup
```

dwlou:
Fixes #505

## Summary

- The `initialOrgManagers` field previously hardcoded `sap.ids` as the identity provider origin for all CF Org Managers, making it impossible to add users from custom IdPs during environment creation.
- Entries now support an optional `|origin` suffix (e.g. `user@company.com|custom.idp`). Entries without a separator continue to default to `sap.ids`, preserving full backward compatibility with existing CRs.

## Backward Compatibility

This is intentionally a non-breaking change. The field type remains `[]string` so existing CRs require no migration.

A proper typed change (replacing `[]string` with `[]User`) is tracked in #837 and should be introduced when promoting this resource to `v1beta1`.

## Test plan

- [ ] Existing `TestFilterOutUser` tests still pass with no changes to behavior
- [ ] New `TestParseManagerString` covers plain email (defaults to `sap.ids`), `email|custom.idp`, and explicit `email|sap.ids`
- [ ] `go build ./...` passes cleanly

### DIFF
--- a/apis/environment/v1alpha1/cfenvironment_types.go
+++ b/apis/environment/v1alpha1/cfenvironment_types.go
@@ -45,7 +45,8 @@ func (u *User) String() string {
 
 // CfEnvironmentParameters are the configurable fields of a CloudFoundryEnvironment.
 type CfEnvironmentParameters struct {
-	// A list of users (email) to assign as the Org Manager role.
+	// A list of users to assign as the Org Manager role.
+	// Each entry is either a plain email address (uses "sap.ids" as origin) or "email|origin" to specify a custom identity provider (e.g. "user@company.com|custom.idp").
 	// The technical user referenced in the ProviderConfig is automatically added as Org Manager and can be omitted from this list (see https://help.sap.com/docs/btp/sap-business-technology-platform/about-roles-in-cloud-foundry-environment).
 	// Cannot be updated after creation --> initial creation only
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="OrgManagers can't be updated once set"
--- a/internal/clients/cfenvironment/cfenvironment.go
+++ b/internal/clients/cfenvironment/cfenvironment.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strings"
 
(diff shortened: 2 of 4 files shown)


mpechkurov (approved):
LGTM

Issue #524, opened 2026-02-24
[BUG] Service instance parameters get lost


ikhandamirov:
**Describe the bug**

Service instance parameters get lost, when exporting a BTP service instance.

**To Reproduce**
1. Create a service instance using the respective crossplane-btp-provider e2e test (test/e2e/testdata/crs/serviceinstance)
2. Note the field
    ```yaml
    parameters:
      HTML5Runtime_enabled: false
    ```
3. Export the created service instance.
4. Note that the exported manifest contains
    ```yaml
    parameters: null
    ```
5. Use BTP CLI to access the instance information with either of the commands:
    - https://help.sap.com/docs/btp/btp-cli-command-reference/btp-list-services-instance
    - https://help.sap.com/docs/btp/btp-cli-command-reference/btp-get-services-instance
6. Note that the returned json does not contain any `parameters` 

**Expected behavior**

Both BTP CLI and the export tool should be able to retrieve service instance parameters.

lennardrother:
`btp list services/instance` does not return parameters. `btp get services/instance` only does so with the parameter `--show-parameters true`. This is only supported if the corresponding service offering has `instances_retrievable` set to true, which destination from test/e2e/testdata/crs/serviceinstance does not.

jn-av:
@ikhandamirov are you planning to work on it or should we make it open for everyone?

ikhandamirov:
@jn-av , this is a [documented](https://github.com/SAP/crossplane-provider-btp/blob/main/docs/end-user-guides/import-landscape/exporter-user-guide.md#specforproviderparameters-is-always-null) limitation. Cannot be solved, because BTP CLI, used by the exporter underneath, does not retrieve `parameters`.

I'm closing the issue as not planned.

Issue #447, opened 2025-12-16
[TASK] Tweak `ServiceInstance` parameters updating behaviour


sk31337:
**Context**

When using Crossplane Provider BTP to manage SAP BTP service instances, users need to understand whether changes to Kubernetes secrets referenced in `.spec.forProvider.parametersFrom` (or `.spec.forProvider.parameterSecretRefs`) automatically trigger reconciliation and update of the service instance.

The BTP Service Operator provides an explicit `watchParametersFromChanges` field (boolean) that allows users to control this behavior. When set to `true`, any change to referenced secret values triggers an automatic update of the service instance.

Since our implementation is based on Upjet, diff detection is outsourced to terraform and it should in theory work, but wasn't properly tested before. 

**Scope**
- [x] verify that changing secret values leads to updating remote instances
- [ ] add e2e tests that verifies that behaviour
- [ ] manually test behaviour with offerings for which the broker does not return data (@sk31337 might be able to assist to find different offerings here)
- [ ] document those insights on updating behaviour and documentation in CRD comments to reflect in CRD browser

**Additional context**

- **BTP Service Operator Reference**: [watchParametersFromChanges documentation](https://github.com/SAP/sap-btp-service-operator?tab=readme-ov-file#specifying-parameters)
- **Crossplane Provider BTP API**: [parameterSecretRefs documentation](https://doc.crds.dev/github.com/SAP/crossplane-provider-btp/account.btp.sap.crossplane.io/ServiceInstance/v1alpha1@v1.5.0#spec-forProvider-parameterSecretRefs)

**Priority**: Low - This affects user experience and operational reliability when managing service instances with dynamic configurations from k8s secrets but because of automatically reconcilation every ~1 minute time to live of this drift is short.

PR #839, opened 2026-07-27
[DO NOT MERGE] Workaround: detect ServiceInstance parameter drift via Service Manager

jn-av:
> ⚠️ **DO NOT MERGE.** This is a deliberate downstream **workaround** for a limitation in the bundled BTP Terraform provider, not a change intended for upstream. It is published as a draft so it can be reviewed and consumed from our fork. See "Why this should not be merged" below.

## Problem

For `ServiceInstance`, a `parameters`-only change on an existing instance is never detected as drift and never reaches BTP. Root cause is in the bundled `terraform-provider-btp`: its `Read()` discards the `parameters` value returned by the Service Manager, so upjet-based drift detection never covers them and `Observe()` reports `UpToDate`. The upstream fix was declined (see [terraform-provider-btp#1643](https://github.com/SAP/terraform-provider-btp/pull/1643)) for state-consistency reasons; tracked here as #524 and #447.

## Workaround (Crossplane-side only)

In `Observe()`, once the TF layer reports `UpToDate`, fetch the instance parameters **directly from the Service Manager** (reusing the existing subaccount-admin lookuper) and compare them against the desired parameters. On a difference, force `ResourceUpToDate:false` so `Update()` runs and sends the parameters. The TF provider binary is not touched.

The comparison is a deep, order-independent **subset** match (`desired ⊆ server`), verified empirically against live BTP instances:

| Case | Offering | Behaviour |
|---|---|---|
| add/change of a desired param | cloud-logging (`instances_retrievable:true`) | **drift detected → update applied** |
| offering returns no params | aicore (`instances_retrievable:false`, HTTP 400) | skip, no drift |
| server-added defaults / extra fields | objectstore | ignored, no phantom drift / loop |

**Deletion is intentionally not handled** — the Service Manager update API cannot reliably clear a parameter (empty `{}` is a no-op, omitting a key merges, explicit `null` crashes the broker), so forcing drift on removal would loop or fail. Removal is left to the recreate path (documented at `paramsSubsetMatch`).

## Tests

- Unit tests for the subset compare and the drift/skip logic (`params_compare_test.go`, `params_drift_test.go`).
- E2E `TestServiceInstance_ParameterDrift` (one-mds/sap-integration): ready → change `businessSystemId` → update reconciles → stays settled (no loop). **Passing** against a live subaccount.
- The existing creation-flow e2e is extended with a non-retrievable "no phantom drift" assertion (destination).

## Why this should not be merged

- It is a **compensating workaround** for a provider-binary limitation the upstream Terraform provider maintainers explicitly chose not to fix. If/when that is addressed upstream, this should be removed rather than carried.
- It adds a Service Manager round-trip per `Observe()` on up-to-date instances and cannot detect parameter *deletion* — acceptable for our use case but not a general-purpose guarantee.

Refs #524, #447, [terraform-provider-btp#1643](https://github.com/SAP/terraform-provider-btp/pull/1643)

### DIFF
--- a/internal/clients/servicemanager/recovery.go
+++ b/internal/clients/servicemanager/recovery.go
@@ -1,8 +1,11 @@
 package servicemanager
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
+	"io"
 	"strings"
 	"time"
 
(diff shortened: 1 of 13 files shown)


jn-av:
Closing in favor of a branch-based PR directly in this repo (I have write access here; the fork was unnecessary). Superseded by the new PR.

Issue #676, opened 2026-05-28
v2 Migration guide 


sdischer-sap:
## Context
Migration to v2 shouldn't case issues unless very specific legacy concepts are used. We still need to document very clearly what those concepts are and how to potentially migrate them.

## Acceptance criteria
- Published Migration Guide in our documentation (and release notes) to clearly documents incompatibilities containing
  - quick summary of https://docs.crossplane.io/latest/guides/upgrade-to-crossplane-v2/#external-secret-stores, link to details
  - removed globalaccount (see results in spike #371)
  - changed `ServiceManager` default `planName`: https://github.com/SAP/crossplane-provider-btp/issues/845
  - warn of consequences if not properly applied during the upgrade

Issue #371, opened 2025-10-16
Spike Remove a CRD - can user upgrade?


GenosseOtt:
**Describe the bug**
1. Install a Browser with a given CRD A
2. Use A and create an instance of it
3. Remove the CRD, 
4. run upgrade

### Highlight in UI 
- https://github.com/openmcp-project/backlog/issues/319

### End user guide (poilicy)
- Outline a Kyberbo policy that either blocjks or allows but audits (creates PR/send email) when deprecated resources are in place

### End user guide (FAQ)
- why can my resource not be updated? what does error message look like 

## Accetpance Criteria 
- Run scenario
- investigate how crossplane handles this case 
- note down end user impact in an FAQ (unless its super end user friendly!)
- Follow up on https://github.com/SAP/crossplane-provider-btp/issues/161 - what can u recommend to customer [bash script to "check and clean by riunnong this"] 
- Bring topic to dev round

sdischer-sap:
Not really relevant right now.

xdj2e9:
As I plan to work on #161 I looked into this. PR https://github.com/SAP/crossplane-provider-btp/pull/850 right now includes the upgrade test that was used for this.

My conclusions:

## 1. Run scenario

Ran the full remove-a-CRD path against `GlobalAccount`:

1. Install provider `v1.11.0` (last release, still serves the `globalaccounts` CRD) → CRD Established, provider Healthy.
2. Create a `GlobalAccount` CR → stored.
3. Upgrade to a build with the CRD removed → provider Healthy again.
4. Inspect the CR and CRD, then delete the CR.

Observed: CRD still served, CR still stored (orphaned) and retrievable, no finalizers, `kubectl delete` removes it cleanly.

## 2. How Crossplane handles it

1. **Upgrade succeeds** — new revision goes Healthy, no stall. Dropping a CRD from the package does not block the revision swap.
2. **CRD and CR retained, not cascade-deleted** — no data loss. Both stay, owned by the inactive prior revision; nothing reconciles them.
3. **Orphan deletes cleanly** — no finalizer, so `kubectl delete` does not hang.

**Note:** The `GlobalAccount` CR did not reach `Available` during the run — the `v1.11.0` provider could not decode the API response, so the deprecated resource has drifted from the live API. Independent of that, the run shows removing the CRD does not block the provider upgrade.

## 3. Notes for endusers (#161, #676)

[...]

Issue #675, opened 2026-05-28
Upgrade Tests for update to v2.0.0 release 


sdischer-sap:
## Context
As described in the related epic the v2.0.0 release should be backwards compatible apart from the specified exceptions there. Therefor we should have some upgrade tests that covers that. 

## Technical requirements
- remove `Globalaccount` from upgrade test resources
- make sure it runs? 
- make sure e2e long and upgrade tests are running in general

xdj2e9:
Closing:
- **Globalaccount test resources removed** — https://github.com/SAP/crossplane-provider-btp/pull/850 removed related unit tests and E2E tests references (nothing was existing in the upgrade tests)

- **Upgrade tests passed** — run [32003287845](https://github.com/SAP/crossplane-provider-btp/actions/runs/32003287845): v1.12.0 → local

- **e2e long passed** — run [31985899559](https://github.com/SAP/crossplane-provider-btp/actions/runs/31985899559)

PR #850, opened 2026-08-03
feat: remove deprecated globalaccount

### LINKED ISSUE #161: Remove deprecated `GlobalAccount`

xdj2e9:
Closes #161.

## Summary

- Dropped the `GlobalAccountRef`/`GlobalAccountSelector` reference fields on `Subaccount`. `Subaccount.spec.forProvider.globalAccountGuid` stays as a plain optional string; the global account is set via `ProviderConfig` instead.
- Removed the deprecated `GlobalAccount` managed resource:
  - deleted the `globalaccounts` CRD and controller
  - `make generate` updated managed/managedlist/deepcopy/resolvers and the `subaccounts` CRD without the GlobalAccount references

## User-visible changes

|---|---|---|
| `GlobalAccount` (`account.btp.sap.crossplane.io/v1alpha1`) | Removed | Was deprecated; use `ProviderConfig.spec.globalAccount`. Requires manual pre-upgrade cleanup - see migration note below. |
| `Subaccount.spec.forProvider` | `globalAccountRef` / `globalAccountSelector` removed | `globalAccountGuid` (plain string) unchanged. |

## Migration

See results in spike #371.

Migration guide will be added with #676


xdj2e9:
Note: upgrade test `Test_KymaEnvironment_External_Name` failed when running them locally. As I could reproduce the same issue on current main I will investigate the failing upgrade tests with the follow up issue https://github.com/SAP/crossplane-provider-btp/issues/675

internal/controller/template.go
```diff
@@ -29,7 +28,6 @@ import (
 // the supplied manager.
 func CustomSetup(mgr ctrl.Manager, o internalopts.CrossplaneOptions) error {
 	for _, setup := range []func(ctrl.Manager, internalopts.CrossplaneOptions) error{
-		globalaccount.Setup,
```
Kukkerem, line 32: Wouldn’t this make existing `GlobalAccount` resources unclearable? Successfully reconciled resources have the Crossplane finalizer, and after removing this controller, nothing can clear it. Maybe we could document the required pre-upgrade cleanup in this PR. WDYT?

SatabdiG:
This a mechanically sound removal 🔥 CRD, controller, generated code, examples, and testdata all cleaned up together — no half-removed state anywhere.

But we do need to sort out our cleanup story before this merges. +1 to @Kukkerem's point above: any GlobalAccount that reconciled successfully carries the Crossplane finalizer, and once this controller is gone there's nothing left to clear it. That means existing GAs on live clusters get stuck on delete indefinitely.

internal/controller/template.go
xdj2e9: Good catch. Thank you for pointing out.

The #371 spike missed this: its CR never reached `Available` (v1.11.0 can't
decode the API), so it never reconciled and never got the finalizer - which is
why that orphan deleted cleanly. I noted the caveat on #371:
https://github.com/SAP/crossplane-provider-btp/issues/371#issuecomment-5266418683

So the migration guide should contain something like:
```bash
# repoint refs to ProviderConfig.spec.globalAccount, then:
kubectl delete globalaccounts.account.btp.sap.crossplane.io --all
# if a CR hangs in Terminating (reconciled → has the finalizer):
kubectl patch globalaccount <name> --type=merge -p '{"metadata":{"finalizers":[]}}'
kubectl delete crd globalaccounts.account.btp.sap.crossplane.io
```

I would prefer to write the migration guide as part of #676 including this new knowledge about the behavior.

Would that be fine for you?

SatabdiG: I'm fine with the guide living in #676 rather than inline here, given the exact steps are now nailed down in this thread. 👍 from me

SatabdiG (approved):
🔥

Kukkerem (approved):
LGTM

PR #854, opened 2026-08-04
feat: add external-name ADR compliance for Entitlement

### LINKED ISSUE #558: [TASK] `External-Name`(Entitlement): Ensure ADR compliance
**Part of Epic: #427** 
## Context

## Technical Tasks Checklist

**Acceptance Criteria:**

- [ ] **Migration Logic**:

**Acceptance Criteria:**
- [ ] **Observe() Tests**:

- [ ] **Create() Tests**:

- [ ] **Update() Tests**:

- [ ] **Delete() Tests**:

**Acceptance Criteria:**
- [ ] **Import Flow Test**:

### 5. Documentation

Kukkerem:
Implements ADR-compliant external-name handling for `Entitlement` as part of Epic #427.

The external-name is the compound key `<subaccount-guid>/<service-name>/<service-plan-name>`, with an optional fourth `<service-plan-unique-identifier>` segment when the spec sets one. BTP exposes no GUID for an entitlement assignment, so the key has to be built from the identity the API itself uses to address the assignment.

The consequence worth reviewing: this key is **not unique per managed resource**. Several `Entitlement` CRs deliberately share one BTP assignment and each contributes part of the requested amount, so the annotation identifies the shared assignment, not the CR. A CR with an empty annotation that finds an existing assignment therefore does not adopt it silently — it requires a same-key sibling whose own annotation already proves the assignment is provider-managed. Without that proof Observe and Create fail with a terminal, user-actionable error instead of taking over quota this provider never created. Deletion inherits the same rule: a CR that never joined the aggregate refuses to finalize while the assignment still reserves something, rather than deleting quota it does not own.

`AutoAssigned` assignments are carved out on both sides: Create skips the BTP write and only stamps the annotation, and deletion finalizes without touching BTP, emitting an `AutoAssignedPreserved` event — BTP grants these implicitly and revoking them is not ours to do.

Because the annotation is now compared against `spec.forProvider` on every reconcile, the five identity fields (`subaccountGuid`/`subaccountRef`, `serviceName`, `servicePlanName`, `servicePlanUniqueIdentifier`) are immutable via `XValidation`. A parsed key that disagrees with the spec is rejected with a message naming the differing segments; an import typo cannot silently re-target a live assignment.

Migration is implicit. `zz_setup.go` switches to `DefaultSetupWithoutDefaultInitializer` so fresh CRs reach Observe with an empty annotation, and both an empty annotation and the pre-ADR default sentinel (`external-name == metadata.name`) resolve the key from spec; Observe stamps the compound form once the assignment is confirmed. Existing CRs need no user action.

Two smaller changes fall out of this. The client interface takes an `ExternalNameKey` instead of the whole CR, which makes the describe cache key explicit (`ExternalNameKey.CacheKey()`, qualifier excluded — the BTP request is not keyed by it). And Create re-reads through `DescribeInstanceFresh`, bypassing the TTL cache, before re-running the ownership guard: the Observe that decided to create may have been served from cache, and another actor could have created the assignment since.

Drift is reported as a `Drift` status condition plus a Warning event deduped against the previously persisted message, so an unchanged diff does not re-emit on every poll tick.

## Import example

External-name is the compound key. Unlike #671, `spec.forProvider` must still carry the same identity — the annotation is validated against it — and import needs `managementPolicies: ["*"]` because the subaccount context comes from the spec.

```yaml
kind: Entitlement
metadata:
  annotations:
    crossplane.io/external-name: 6aa64c2f-38c1-49a9-b2e8-cf9fea769b7f/postgresql-db/development
  name: entitlement-import
spec:
  managementPolicies: ["*"]
  forProvider:
    subaccountGuid: 6aa64c2f-38c1-49a9-b2e8-cf9fea769b7f
    serviceName: postgresql-db
    servicePlanName: development
    amount: 1
  providerConfigRef:
    name: account-provider-config
```

## Test plan

- [x] `go build ./...`, `go vet -tags e2e ./test/e2e/...`, `go vet -tags upgrade ./test/upgrade/...`
- [x] Unit tests for key parsing/building/mismatch, the client, and every Observe/Create/Update/Delete branch above
- [ ] E2E `TestEntitlementImportFlow` green in CI
- [ ] Upgrade `Test_Entitlement_External_Name` green in CI (legacy `metadata.name` annotation migrates to the 3-segment key, resource stays Ready)

Closes #558


dwlou:
what does this mean? 

`AutoAssigned assignments are carved out on both sides: Create skips the BTP write and only stamps the annotation, and deletion finalizes without touching BTP, emitting an AutoAssignedPreserved event — BTP grants these implicitly and revoking them is not ours to do.`

Kukkerem:
> what does this mean?
> 
> `AutoAssigned assignments are carved out on both sides: Create skips the BTP write and only stamps the annotation, and deletion finalizes without touching BTP, emitting an AutoAssignedPreserved event — BTP grants these implicitly and revoking them is not ours to do.`

Hey, it means, if I understood correctly, that some entitlements aren't ours to manage. BTP marks some plans as automatically assigned and available to a subaccount without any admin action (per the docs), and they are not removable. For those, we adopt the assignment but never write to BTP. On create there's nothing to create, so we just record the identity. On delete, there's nothing to revoke, so we let the resource finalize and emit an event saying the entitlement was left untouched.

internal/controller/account/entitlement/entitlement.go
```diff
@@ -9,6 +9,8 @@ import (
 	"github.com/pkg/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/crossplane/crossplane-runtime/v2/pkg/event"
```
SatabdiG, line 12: I noticed that this file imports github.com/pkg/errors, but internal/clients/entitlement/entitlement.go and key.go both use github.com/crossplane/crossplane-runtime/v2/pkg/errors (and key_test.go uses stdlib errors). So three errors packages across one feature. It's not currently breaking anything ofcourse — TestUpdate's wantSentinel checks confirm errors.Is survives the pkg/errors ↔ crossplane-runtime/pkg/errors wrap boundary today — but the adoption-guard and mismatch-error paths depend on that chain staying intact, and that's an implicit cross-package guarantee rather than an enforced one. 

Can we standardize on crossplane-runtime/v2/pkg/errors here to match the client layer or maybe we can add a comment about controller intentionally differs? What do you think?

internal/controller/account/entitlement/entitlement.go
```diff
@@ -409,6 +836,12 @@ func (c *external) findRelatedEntitlements(
 		if ent.Spec.ForProvider.ServicePlanName != ours.Spec.ForProvider.ServicePlanName {
 			continue
 		}
+		// A qualifier identifies a distinct plan: nil matches only nil.
+		entQualifier, ourQualifier := ent.Spec.ForProvider.ServicePlanUniqueIdentifier, ours.Spec.ForProvider.ServicePlanUniqueIdentifier
+		if (entQualifier == nil) != (ourQualifier == nil) ||
+			(entQualifier != nil && *entQualifier != *ourQualifier) {
+			continue
+		}
 		if ent.GetCondition(xpv1.Deleting().Type).Reason == xpv1.Deleting().Reason {
```
SatabdiG, line 847: Delete() calls `DescribeInstance/DeleteInstance` directly it fully trusts that Observe() already validated ownership earlier in the same reconcile. That's consistent with the managed-reconciler contract (Delete only follows an Observe that returned ResourceExists:true on the same object), and persistExternalName being skipped for deleting CRs relies on the same assumption elsewhere in this file.Maybe  we add a one-line comment stating that explicitly? It'll save whoever touches this next from re-deriving it by reading five other functions to understand the flow

internal/clients/entitlement/key_test.go
```diff
@@ -0,0 +1,316 @@
...
+	if withQual1.CacheKey() != withQual2.CacheKey() {
+		t.Fatalf("CacheKey() differs for keys that only differ by qualifier: %q vs %q", withQual1.CacheKey(), withQual2.CacheKey())
+	}
+	if withQual1.CacheKey() != withoutQual.CacheKey() {
+		t.Fatalf("CacheKey() differs between a present and an absent qualifier: %q vs %q", withQual1.CacheKey(), withoutQual.CacheKey())
+	}
+}
+
+func TestNewExternalNameKeyRejectsInvalidSpec(t *testing.T) {
+	tooLong := strings.Repeat("a", 510)
```
SatabdiG, line 243: This covers a slash embedded in ServiceName ("alert/notification") but not one embedded in ServicePlanUniqueIdentifier. Since the qualifier is the field conditionally appended to the segment loop rather than always present, it's the one most likely to get missed if that loop is ever refactored — do you think its worth adding a case for it here?

SatabdiG (commented):
🔥 

This is a genuinely hard problem — a compound key that's shared across multiple CRs, with no GUID from the provider to anchor identity, is one of the more complex instance of our external-name series so far — and I thought that the design holds together well!

I liked your the four-exception model in external-name-handling.md (aggregate adoption, legacy-sentinel sibling proof, own-sentinel self-migration, AutoAssigned self-adoption) to carve up what could easily have become one tangled conditional, and the test suite! 

Great work, loved the ownership-guard design!

internal/clients/entitlement/key_test.go
Kukkerem: It makes sense, thank you, I added it!

internal/controller/account/entitlement/entitlement.go
Kukkerem: Thank you, great insight, added it!

internal/controller/account/entitlement/entitlement.go
Kukkerem: `pkg/errors` is the convention in `internal/controller` 34 files to 1, so changing this would make it the outlier. Maybe a repo-wide sweep in a separate PR would be better? The sentinel chain is covered either way, 17 `wantSentinel` assertions do `errors.Is` across the boundary.

SatabdiG: good point! Doing a later repo wide sweep sounds like a good idea. It's a seperate ticket though shouldn't block this one.

SatabdiG (dismissed):
I think this is one of the more complex external-names.

This PR introduces five predicates (assignFailedNoQuota, assignmentStillReserved, deletingWithReservedQuota, deletingAutoAssigned, adoptionGuardApplies) that each answer a differently-scoped "is this reserved/absent" question. 

They're individually well-named and well-tested, but easy to confuse with each other at a glance since they overlap conceptually. Would it be worth adding a short table in the external-name.md; mapping predicate → what it answers → where it's used, plus a note on the two easiest to mix up? happy to open this as a follow-up PR instead if you'd rather keep this one scoped as-is.

Kukkerem:
> I think this is one of the more complex external-names.
> 
> This PR introduces five predicates (assignFailedNoQuota, assignmentStillReserved, deletingWithReservedQuota, deletingAutoAssigned, adoptionGuardApplies) that each answer a differently-scoped "is this reserved/absent" question.
> 
> They're individually well-named and well-tested, but easy to confuse with each other at a glance since they overlap conceptually. Would it be worth adding a short table in the external-name.md; mapping predicate → what it answers → where it's used, plus a note on the two easiest to mix up? happy to open this as a follow-up PR instead if you'd rather keep this one scoped as-is.

Good catch on the confusion. Table went into the ADR (`external-name-handling.md`). The `external-name.md` is generated below the `## Generated Data Below` marker, so it wouldn't survive a regen. It did surface a gap though: added notes there on the delete refusal and on `AutoAssigned` never being revoked, through the Go doc comment that feeds the generator.

SatabdiG (approved):
lgtm

PR #859, opened 2026-08-04
chore(ci): derive Go version from go.mod in all workflows

xdj2e9:
Replaces hardcoded `GO_VERSION` env vars with `go-version-file: go.mod` in all `actions/setup-go` steps across every workflow. `go.mod` is now the single source of truth for the Go version; future bumps via renovate propagate to CI automatically without a separate workflow change.

✅  Workflow logs confirm that resolved go version matches the one specified in `go.mod`

Triggered by failing workflows in this renovate PR: https://github.com/SAP/crossplane-provider-btp/pull/857

### DIFF
--- a/.github/workflows/check-go-licenses.yaml
+++ b/.github/workflows/check-go-licenses.yaml
@@ -3,10 +3,6 @@ name: Check Go Dependency Licenses
 on:
   workflow_call:
 
-env:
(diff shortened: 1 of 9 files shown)


Kukkerem (approved):
LGTM

PR #860, opened 2026-08-05
chore(deps): drop obsolete genproto replacement

xdj2e9:
The conflict is no longer existing. Downstream dependencies now all import from the split `.../api` module.

#838 can be closed after this PR is merged as the bump is then not required anymore.


SatabdiG (approved):
👍

PR #871, opened 2026-08-06
test(tfclient): cover the error branches of the terraform setup builders

### LINKED ISSUE #63: Testing for Connecting TF Resources
## Context
During the work on https://github.com/SAP/crossplane-provider-btp/pull/53 we noticed that the code for connecting a tf resource to a `ProviderConfig` isn't being tested right now.
This refers to the two methods called `TerraformSetupBuilder` and `TerraformSetupBuilderNoTracking` ([see here](https://github.com/SAP/crossplane-provider-btp/blob/main/internal/clients/tfclient/tfclient.go#L49)). While the first one is being used for any upjetted resource type the second one is being used specifically within ServiceManager and CloudManagement, where we have a native controller then internally delegates to a tf resource. 
It probably makes sense to also combine those methods while we are at it. You can refer to similar tests like the ones for the rolecollection resource [here](https://github.com/SAP/crossplane-provider-btp/blob/main/internal/controller/security/rolecollection/rolecollection_test.go#L319).

## Expected Outcome
- [ ] `TerraformSetupBuilder` and `TerraformSetupBuilderNoTracking` are covered by unittests

AliAltivate:
Addresses #63.

### What was already there

The issue says the code connecting a tf resource to a `ProviderConfig` "isn't being tested right now". That's no longer entirely true — `setup_test.go` already covers the happy path and the custom-IDP conditional for both builders. What it doesn't cover is any error branch. Measured before this change:

| | before |
| --- | --- |
| `TerraformSetupBuilder` | 72.4% |
| `TerraformSetupBuilderNoTracking` | 76.2% |

Every uncovered statement in both functions was an error branch.

### What this adds

Table-driven tests in the existing style, for both builders:

- managed resource is not a `LegacyManaged`
- `ProviderConfigReference` is nil (`TerraformSetupBuilder` only — `NoTracking` doesn't have this branch)
- `ResolveProviderConfig` fails
- the `ProviderConfigUsage` apply fails (`TerraformSetupBuilder` only)
- the credentials-secret lookup fails
- the secret has nil data
- the credentials are not valid JSON

Each case asserts the specific error rather than merely that something failed, so a refactor that returns the wrong error is caught.

| | before | after |
| --- | --- | --- |
| `TerraformSetupBuilder` | 72.4% | **96.6%** |
| `TerraformSetupBuilderNoTracking` | 76.2% | **100%** |

Package total goes 72.9% → 78.8%. No production code is touched.

### The one branch I did not cover, and why

`TerraformSetupBuilder` is at 96.6% rather than 100% because of the `errTrackRUsage` branch:

```go
if err = tracking.NewDefaultReferenceResolverTracker(client).Track(ctx, mg); err != nil {
    return ps, errors.Wrap(err, errTrackRUsage)
}
```

I could not reach it honestly with `fake.LegacyManaged`. In `resourcetracker.go`, `Track` returns early only if `createTracking` fails for a resolved reference — and `fake.LegacyManaged` has no reference-tagged fields, so that loop body never runs. The other candidate, making `List` fail, doesn't work either: `SetConditions` swallows that error into `NewInUseCondition(hasUsages, err)` and `Track` still returns `nil`. Reaching it would need a purpose-built managed type with reference tags plus a failing Get on the referenced object, which felt like more machinery than this issue warrants. Happy to add it if you'd like it covered.

### Notes for the reviewer

- Where the middle of an error chain comes from crossplane-runtime rather than this repo (`cannot apply ProviderConfigUsage` / `cannot get object` from the applicator, `cannot get credentials secret` from `CommonCredentialExtractor`), the assertion checks for the substrings that matter instead of pinning the full string. The previous commit here was a crossplane/upjet bump, and I didn't want to leave tests that snap on the next one and look like a real regression.
- The `TrackProviderConfigUsageError` case fails the mocked `Get` for `*v1alpha1.ProviderConfigUsage`, because the usage tracker goes through `Apply`, which reads before it writes. `MockCreate`/`MockUpdate` are deliberately left nil in that case — if the path went through them instead, the test would fail rather than silently pass.
- `MockList` is set on the tracking suite only because `SetConditions` calls `List` on the cases that get past the tracker; without it they nil-panic. It isn't masking an assertion, since that error is swallowed either way.
- Verified by mutation: dropping the `errors.Wrap(err, errGetServiceAccountCreds)` in `setup.go` makes the corresponding cases fail. Reverted; the diff is the test file only.
- `go build ./...` clean, `go vet` clean, `gofmt` clean, and the full suite is green (60 packages, 0 failures).

The issue also suggests combining the two builders while we're here. I left that alone deliberately — it changes production code and reads like a separate decision, and I'd rather land the coverage first. Glad to follow up if you want them merged.


xdj2e9 (approved):
LGTM

Agreed on leaving out the refactoring here.

PR #873, opened 2026-08-06
fix(accounts): surface API error details instead of only the general message

### LINKED ISSUE #92: [FEATURE] More detailed error feedback from accounts API
**Is your feature request related to a problem? Please describe.**
We found an unexpected case in the accounts API. If the `displayName` contains a `/` the POST request will fail and the provider will show an error event like this: `API Error: Request payload is invalid, Code 11000`. What exactly was the problem we could only find by debugging the issue in dev mode. The raw response from the API contained more details on what the issue exactly is.

**Describe the solution you'd like**
Reflect that more precise error in the responses as well. See snippet below for how the structure exactly is.
 

**Additional context**
Example raw response:
```
{
  "error" : {
    "code" : 11000,
    "message" : "Request payload is invalid",
    "target" : "/accounts/v1/subaccounts",
    "correlationID" : "bb8d3f4b-c9f8-4126-4813-8265e34d687a",
    "details" : [ {
      "code" : 11001,
      "message" : "displayName: Display name can't contain / "
    } ]
  }
}
```

We can't know for sure, but this `details` section in the response probably is new. Ideally we should use this occasion to raise a support ticket to update the documentation to reflect the proper schema: https://api.sap.com/api/APIAccountsService/path/createSubaccount
The `details` is contained there, but the overall structure is incorrect, which we internally modified in the generated client to compensate for that.

AliAltivate:
Fixes #92.

### The cause

The details were already being parsed — they just never reached the user.

`ApiExceptionResponseObjectError` has no `Details` field, but its generated `UnmarshalJSON` deletes only `code`, `correlationID`, `message` and `target` and collects every remaining key into `AdditionalProperties`. So for the response in the issue, `details` is sitting in there as a `[]interface{}` the whole time:

```
AdditionalProperties: map[string]interface{}{
  "details": []interface{}{
    map[string]interface{}{"code": float64(11001), "message": "displayName: Display name can't contain / "},
  },
}
```

`specifyAPIError` formatted only the top-level `message` and `code` and dropped the rest, which is why the reason was only visible by reading the raw response in dev mode.

### What this changes

`API Error: Request payload is invalid, Code 11000`

becomes

`API Error: Request payload is invalid, Code 11000; details: displayName: Display name can't contain /`

The conversion existed as four byte-identical copies — `internal/clients/directory`, `internal/clients/servicemanager` (as `specifyAccountsAPIError`), and the `globalaccount` and `subaccount` controllers. Since the issue asks for this on any accounts request rather than just the POST, I moved it to `internal/apierror.FromAccounts` and pointed all four at it, instead of applying the same fix four times.

### Backwards compatibility

This is the part the issue calls out, so it is where most of the tests went. When there are no usable details the message is byte-identical to before — same `fmt.Sprintf("API Error: %v, Code %v", ...)` with the same `internal.Val` derefs — so 409 conflicts and everything else convert exactly as they did. The `; details: ` suffix appears only when at least one detail message survives parsing.

Everything coming out of `AdditionalProperties` is untyped JSON, so each step is checked: `details` missing, not an array, elements not objects, `message` absent, null, non-string or blank all fall back to the general message rather than failing or leaking a Go-rendered value. Malformed entries are skipped individually, so one bad entry doesn't discard its usable siblings.

### Tests

18 cases in `internal/apierror/accounts_test.go`, 100% statement coverage of the new package. They drive the real `FromAccounts` entry point rather than an inner function, and build inputs by unmarshaling raw response JSON so the actual `AdditionalProperties` path is exercised instead of a hand-built map.

`GenericOpenAPIError` has only unexported fields and no constructor, so I extended the existing `internal/testutils` reflection helper with `NewAccountAPIErrorFromBody` (model + body from a raw response, which is what makes a `details` test possible at all) and `NewAccountAPIErrorBodyOnly` (body without a model, for the raw-body fallback). The existing `NewAccountAPIError` is untouched and its four current callers are unaffected.

Verified by mutation — each of these makes the suite fail, and I reverted each after checking:
- never appending details
- dropping the `strings.TrimSpace`
- changing `len(details) > 0` to `>= 0` so the suffix is always appended

### Verification

- `go build ./...`, `go vet`, `gofmt` clean
- full suite green: 61 packages, 0 failures
- `make lint` (golangci-lint v2.12.2): 0 issues
- `make check-diff` is unaffected — this touches no `apis/` types, CRDs or OpenAPI specs, only hand-written files under `internal/`

### What I could not verify, and two things to decide

**Not exercised against a live BTP account.** I have no global account to hit, so the parsing, the fallback and the 409 behaviour are covered by unit tests against the documented response shape from the issue. If the real API returns a `details` shape I haven't anticipated, that path is untested. Worth a maintainer sanity-check against a real tenant before merging.

**I kept only the detail `message`, not the detail `code`** (11001 in the example), on the grounds that the message already names the field and the reason. Easy to include if you'd rather have it for support tickets — say the word.

**The new package is optional.** If you'd rather not take `internal/apierror`, I'm happy to inline `FromAccounts` at the call sites instead; it just means the details logic exists in four places again. Same for `NewAccountAPIErrorBodyOnly` — it guards the raw-body fallback branch, which this PR doesn't change, so it's the easiest piece to drop if you want a smaller diff.

### Two accounts-API paths that still bypass the conversion

Both pre-existing, both deliberately left alone, both worth your call:

**`createBTPSubaccount`'s 409 branch** (`internal/controller/account/subaccount/subaccount.go:473`) wraps the raw error, so a subaccount name conflict surfaces as `...adopt the existing resource: 409 Conflict` rather than going through the conversion. `CreateDirectory` handles the same situation the other way — it wraps `specifyAPIError(err)` — so the two are inconsistent. Given requirement (d) you may well want this routed through `FromAccounts` too, but it changes a message that `subaccount_test.go:1306` currently pins, which makes it a behaviour decision rather than part of this fix. Say the word and I'll do it here or in a follow-up.

To be clear about what this means for the reported bug: the `/` in a display name is a 400/11000, which takes the `return specifyAPIError(err)` path, so the case in the issue is fixed. Only the 409 branch is unconverted.

**`btp/subaccount.go:GetBTPSubaccount`** calls the accounts API and returns the error raw. It has no callers anywhere in the repo, so it looks like dead code — happy to wire it in or delete it separately.

### DIFF
--- a/internal/apierror/accounts.go
+++ b/internal/apierror/accounts.go
@@ -1,0 +1,74 @@
+// Package apierror translates errors returned by the generated BTP OpenAPI
+// clients into messages that are useful in a managed resource's status and
+// events.
+package apierror
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/sap/crossplane-provider-btp/internal"
+	accountclient "github.com/sap/crossplane-provider-btp/internal/openapi_clients/btp-accounts-service-api-go/pkg"
+)
+
+// FromAccounts converts an error returned by the accounts service client into a
+// readable error, including the per-field reasons the API reports in "details".
+// Errors that did not come from that client are returned unchanged.
+func FromAccounts(err error) error {
+	genericErr, ok := err.(*accountclient.GenericOpenAPIError)
+	if !ok {
+		return err
+	}
+
+	if accountError, ok := genericErr.Model().(accountclient.ApiExceptionResponseObject); ok {
+		general := fmt.Sprintf("API Error: %v, Code %v", internal.Val(accountError.Error.Message), internal.Val(accountError.Error.Code))
+
+		// The API reports the actual cause of a rejected payload in a "details"
+		// array rather than in the top-level message. Without it the caller only
+		// sees "Request payload is invalid, Code 11000" and has to inspect the
+		// raw response to find out which field was wrong.
+		if details := detailMessages(accountError.Error.AdditionalProperties); len(details) > 0 {
+			return fmt.Errorf("%s; details: %s", general, strings.Join(details, "; "))
+		}
+		return errors.New(general)
+	}
+
+	if genericErr.Body() != nil {
+		return fmt.Errorf("API Error: %s", string(genericErr.Body()))
+	}
+
+	return err
+}
+
+// detailMessages pulls the messages out of the error's "details" array.
+//
+// The generated ApiExceptionResponseObjectError has no Details field, but its
+// UnmarshalJSON collects every key it does not model into AdditionalProperties,
+// so the details survive there as plain JSON values. Anything not matching the
+// expected shape is skipped rather than reported, which leaves the caller with
+// the same general message it would have produced before.
+func detailMessages(additionalProperties map[string]interface{}) []string {
+	entries, ok := additionalProperties["details"].([]interface{})
+	if !ok {
+		return nil
+	}
+
+	messages := make([]string, 0, len(entries))
+	for _, entry := range entries {
+		detail, ok := entry.(map[string]interface{})
+		if !ok {
+			continue
+		}
+		message, ok := detail["message"].(string)
+		if !ok {
+			continue
+		}
+		if message = strings.TrimSpace(message); message == "" {
+			continue
+		}
+		messages = append(messages, message)
+	}
+
+	return messages
+}
--- a/internal/apierror/accounts_test.go
+++ b/internal/apierror/accounts_test.go
@@ -1,0 +1,153 @@
+package apierror
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/sap/crossplane-provider-btp/internal/testutils"
+)
+
+func TestFromAccounts(t *testing.T) {
+	plain := errors.New("boom")
+
+	tests := map[string]struct {
+		in error
+		// want is the expected message. wantSame additionally asserts the
+		// original error is handed back untouched, rather than merely
+		// formatting to the same text.
+		want     string
+		wantSame bool
+	}{
+		"non-generic error passes through": {
+			in:       plain,
+			want:     "boom",
+			wantSame: true,
+		},
+		"nil passes through": {
+			in: nil,
+		},
+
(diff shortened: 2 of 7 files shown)


AliAltivate:
[in-session-review sha=0d21364d]

Dual review before opening. Codex was unavailable (usage limit, resets Aug 8), so the second pass was **GLM-5.2, not Codex**. Both reviewers ran unprimed against a refute-first rubric — report only P0/P1, and try to disprove each finding before reporting it.

**Reviewer 1 (Claude):** no P0/P1. Traced the no-details path to confirm byte-identity with the pre-change output, checked every type assertion out of `AdditionalProperties` for panic risk, confirmed the other `specifyAPIError`-style functions belong to *different* generated clients (xsuaa, saas, entitlement, provisioning) so excluding them is correct, and confirmed `%+v` is only ever applied to custom resources — so dropping `github.com/pkg/errors` in the subaccount copy loses nothing observable.

**Reviewer 2 (GLM-5.2):** no P0. One P1, which I accept as accurate but out of scope: `createBTPSubaccount`'s 409 branch (`subaccount.go:473`) wraps the raw error and so bypasses the conversion entirely, inconsistently with `CreateDirectory`, which wraps `specifyAPIError(err)`. I verified this directly rather than taking it on trust.

**Resolution:** not changed here, and called out in the PR description instead. Routing it through `FromAccounts` would change a message that `subaccount_test.go:1306` currently asserts, which makes it a behaviour decision for the maintainers rather than part of this fix. It also does not affect the case in #92 — a `/` in a display name returns 400/11000, which takes the converted path.

Both reviewers independently flagged `btp/subaccount.go:GetBTPSubaccount` as bypassing the conversion; I confirmed it has no callers, so it is pre-existing dead code and is noted in the description as a possible follow-up.

**Checks run locally** (I can't trigger this repo's Actions): `go build ./...`, `go vet` and `gofmt` clean; full suite 61 packages / 0 failures; `make lint` (golangci-lint v2.12.2) 0 issues; new package at 100% statement coverage; three mutations each verified to fail the suite before being reverted.

PR #874, opened 2026-08-06
fix(ci): pin e2e crossplane chart to 2.1.8 to unblock provider install

xdj2e9:
- **Pin the e2e Crossplane chart to 2.1.8** (from 2.3.4). Chart 2.2.0+
  (crossplane/crossplane#6981) resolves `spec.package` against a real registry before the package
  cache resulting in 401s responses by Docker hub failing all E2E tests. 2.1.8 is
  the last release before that breaking change
- **Constrain renovate to `allowedVersions=<2.2.0`** so it stops re-proposing the breaking bump.

We already hit this exact issue and fixed it the same way in #820 (revert to 2.1.7); #856 then bumped
to 2.3.4 and reintroduced it. The renovate pin prevents a third round.

See [v2.2.0 crossplane release notes](https://github.com/crossplane/crossplane/releases/tag/v2.2.0):
```
The on-disk structure of the package cache has changed. This breaks an undocumented behavior via which packages could be side-loaded into Crossplane, which was especially useful for testing. See https://github.com/crossplane/crossplane/pull/6981 for details on the change and https://github.com/crossplane/crossplane/issues/7147 for discussion of the test changes necessary to accommodate it.
```

### DIFF
--- a/.github/workflows/e2e_test.yaml
+++ b/.github/workflows/e2e_test.yaml
@@ -47,11 +47,14 @@ env:
   E2E_BUILD_REGISTRY: index.docker.io/e2e
   E2E_PROVIDER_IMAGE: index.docker.io/e2e/provider-btp-amd64
   CROSSPLANE_CLI_VERSION: v2.0.2
-  # renovate: datasource=helm depName=crossplane registryUrl=https://charts.crossplane.io/stable
-  CROSSPLANE_CHART_VERSION: '2.3.4'
-  # SHA256 of crossplane-2.3.4.tgz downloaded from charts.crossplane.io/stable
+  # renovate: datasource=helm depName=crossplane registryUrl=https://charts.crossplane.io/stable versioning=semver allowedVersions=<2.2.0
+  # Pinned to last 2.1.x: 2.2.0+ (crossplane/crossplane#6981) resolves spec.package
+  # against a real registry before the cache, breaking the local :latest install path.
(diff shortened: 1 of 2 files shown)


xdj2e9:
Created follow up issue: https://github.com/SAP/crossplane-provider-btp/issues/875

SatabdiG (dismissed):
lgtm

xdj2e9:
The fix is in the workflow file that will be used from the main branch so this branch never can pass the E2E checks. I verified locally that the adapted chart version fixes the provider install with these changes:
```
$ make test-acceptance testFilter=TestAccount
[...]
===========Test Summary===========
--- PASS: TestAccount (128.10s)
    --- PASS: TestAccount/BTP_Subaccount_Controller (128.10s)
        --- PASS: TestAccount/BTP_Subaccount_Controller/Check_Subaccount_Created (0.01s)
        --- PASS: TestAccount/BTP_Subaccount_Controller/Check_Subaccount_Updated (15.06s)
        --- PASS: TestAccount/BTP_Subaccount_Controller/Check_rejected_Updates (0.02s)
        --- PASS: TestAccount/BTP_Subaccount_Controller/Check_Subaccount_moved_to_directory (10.05s)
        --- PASS: TestAccount/BTP_Subaccount_Controller/Check_Subaccount_Deleted (45.06s)
PASS
✅ All tests passed
```

SatabdiG (approved):

PR #877, opened 2026-08-06
[Do-not-merge]fix: deliver async Terraform results to the real MR and report external health honestly

mirzakopic:
> **FYI**: this branch is being soaked as an RC build in a production-like environment before it should be considered for merge. Feedback welcome meanwhile.

## Motivation

Operating this provider at fleet scale, we root-caused a class of incidents where BTP resources failed externally while every observable layer of the provider reported health, and refused deletions retried blindly and invisibly forever. Two defects compound:

1. **Every async Terraform callback result is silently dropped.** The internal Terraform shadow resource is named `TF-<name>` (Terraform resource names must not start with a digit). The async callbacks in `internal/clients/tfclient/tf_async_callback.go` receive that shadow identity and call the save function with `name.String()` — which, for a cluster-scoped resource, renders `"/TF-<name>"`. The `Get` inside the save function 404s on **every** async completion, for create, update and destroy alike, logged only at `level=info` ("callback failed … not found"). In one measured 6-hour window: 7,397 dropped destroy results, 579 updates, 52 creates. Failed creates are therefore never recorded (`checkAsyncOperationFailure` can never fire → `Available()` is set over a dead instance — this is #280's mechanism), and failed destroys leave no condition, no event, no backoff — just a blind re-destroy every ~45–80s, forever.
2. **Readiness ignores the external observation.** `Observe`'s `UpToDate` branch sets `xpv1.Available()` without consulting `status.atProvider.state` / `ready` / `usable`; instances with `state=failed, ready=false, usable=false` still report `Ready=True`. Same class on Entitlements: `entityState=PROCESSING_FAILED` (e.g. a quota decrease rejected because instances still consume it) is reported `Synced=True` with the platform's `stateMessage` copied nowhere.

## Changes

| Commit(s) | Fix |
|---|---|
| `159e48c`, `ba73f12` | **Async callbacks resolve the real MR**: the save path receives a proper `types.NamespacedName` with the `TF-` shadow prefix stripped (names without the prefix pass through untouched, so other consumers are unaffected). Resolution failure is now `level=error` + a best-effort Warning event instead of `info`. The status write is retried on conflict instead of one-shot. Condition messages are head-truncated at 16KiB (Terraform CLI error output can be tens of KiB; head-preserving because the create-conflict detection greps the message head). Contract tests assert the round-trip for cluster-scoped MRs. |
| `e39894b`, `d9b54bc`, `fa06f50` | **Readiness reflects external health** (closes #280): a ServiceInstance whose observation reports `state=failed` — or `ready=false`/`usable=false` while not deleting — gets `Ready=False/ExternalResourceFailed` with the observed detail instead of `Available()`. Entitlements in `PROCESSING_FAILED` are no longer `Available` and carry the platform `stateMessage` in the condition. The Terraform shadow deliberately keeps its `Available` mark once the external ID exists, so upjet still runs `Plan` and drift detection / spec repair keep working for unhealthy instances. |
| `08f2067`, `1623bca` | **Nil-response guards**: `GetEnvironmentInstanceByID` dereferenced `resp.StatusCode` on transport errors where `resp == nil` — an unrecoverable panic loop per affected KymaEnvironment (~1,160 recovered panics/6h observed). Fixed there plus two identical patterns found by sweep (`directory.go`, `subscription.go`). |
| `7295306` | **ResourceUsage: NotFound on our own status write is terminal**: an object deleted mid-reconcile no longer requeues forever (3,546 error lines/6h for already-deleted objects observed). |
| `b292b14` | **Usage tracking tolerates a missing source during deletion**: when the referencing resource is itself being deleted, a NotFound tracked source no longer fails `connect()` closed — the resource stayed permanently un-reconcilable and un-deletable otherwise. Fail-closed behavior is unchanged outside deletion. |
| `91a0a4b`, `e8a9d67`, `82dbf76` | **Visible delete-failure state + external-state metric**: a failed async destroy now surfaces a dedicated `ExternalDeletion` condition (`DeleteFailed`, carrying since-when and the last error) — deliberately not the `Ready` type, which the managed reconciler overwrites with `Deleting()` on every pass — plus an edge-triggered Warning event. New gauge `btp_managed_resource_external_state{kind,state}` (bounded cardinality: kind × state only) makes failed populations scrapeable. Collector Lists are bounded (30s) and missing CRDs are log-and-skip so trimmed installs don't crash-loop. |
| `aa9f817`, `dee2d16` | **Actionable subaccount deletion errors**: a BTP `Code 70011` rejection ("You can't delete subaccounts with active service instances") now enumerates the blocking service instances via the subaccount's Service Manager admin binding; when enumeration itself fails, the error says exactly why (no binding available / list call failed with detail) instead of asserting a guessed cause. |

## Behavioral changes reviewers should weigh

- **Entitlement `PROCESSING_FAILED` flips from `Available` to `Ready=False`.** An existing code comment argued for `Available` to avoid flapping downstream orchestration; we believe the truthful signal wins (the masked variant is how failed states went unnoticed), but compositions gating on Entitlement readiness will now see the failure. One existing test assertion updated accordingly.
- **With callback results finally landing, `checkAsyncOperationFailure` starts actually firing** for failed creates, driving the not-up-to-date/Update path that previously never ran. Destroy failures intentionally do not route through it (that would call `Update()` on a deleting resource).
- `SaveConditionsFn`'s signature changed (`string` → `types.NamespacedName`); the package is `internal/`, so no external consumers can be affected.

## Testing

- `go build ./...`, `go vet ./...`, `go test ./... -count=1` green (62 packages).
- New unit/contract tests: callback name round-trip (cluster-scoped, prefixed and unprefixed), conflict-retried save, readiness mapping fixtures (`state=failed` with `ready/usable` both true and both false — both cohorts were observed in production), shadow-keeps-Available (drift detection preserved), nil-response transport errors, mid-reconcile deletion of a ResourceUsage, tracking tolerance only-under-deletion, delete-failure condition edge-triggering (incl. stale pre-deletion failures not misreported), 70011 enumeration success/failure shapes.
- Not yet run here: `make test-e2e-long` / acceptance (kind-based); the RC soak covers live behavior.


jn-av:
Rebased onto current main by merging origin/main into the branch. This picks up the self-healing fixes from #922 (CloudManagement) and #924 (ServiceManager) which are now included. Build is green.

PR #886, opened 2026-08-10
chore: add safe-to-delete label to directories

xdj2e9:
Adds `safe-to-delete: ["yes"]` to every test `Directory` fixture so CI-created directories carry the same cleanup opt-in label the subaccount fixtures already use. Without it, leaked test directories have no reliable delete indicator


SatabdiG (approved):
lgtm

PR #887, opened 2026-08-10
chore(ci): use BTP CLI instead of raw HTTP requests in e2e cleanup script

### LINKED ISSUE #252: [TASK] E2E - cleanup.go switch to BTP CLI approach
The cleanup script for the e2e tests uses a mix of btp cli calls and http requests.
It would be clean to change the rest of the http requests to btp cli calls:
- get subaccounts
- delete subaccount
- get directory
- delete directory
- get uaa auth would not be necessary anymore if switching to btp cli for subaccount/directory operations

## Advantages
- CLI is maintained for us
- one approach for everything

**Acceptance criteria**
- all http requests should be done with btp cli calls

AliAltivate:
Fixes #252.

**What this PR does**

Switches the remaining raw HTTP requests in `.github/workflows/cleanup.go` to `btp` CLI calls, so the cleanup script uses one approach for everything:

| Before (HTTP against the accounts service) | After (btp CLI) |
|---|---|
| `GetUaaAuth` (CIS binding → UAA token) | removed — no longer needed |
| `GetSubaccounts` (`GET /accounts/v1/subaccounts`) | `btp --format json list accounts/subaccount` |
| `DeleteSubaccount` (`DELETE …?forceDelete=true`) | `btp delete accounts/subaccount <guid> --force-delete --confirm` |
| `GetDirectoriesOfBuild` (`GET /accounts/v1/directories/<guid>`) | `btp --format json get accounts/directory <guid>` |
| `DeleteDirectories` (`DELETE …?forceDelete=true`) | `btp delete accounts/directory <guid> --force-delete --confirm` |

Command syntax was verified against btp CLI client **v2.106.1** `--help`. The `list accounts/subaccount` JSON shape (`{"value": [...]}`) is the same one `cmd/exporter/btpcli/subaccount.go` in this repo already parses, and the command lists subaccounts in directories too — same scope as the old global-account CIS call.

**Behavior preserved**

- The "pending deletion" filter in `GetSubaccounts`, the `BUILD_ID` label checks for subaccounts and directories, the directory double-check (hierarchy custom property → per-directory label), the 30-attempt/45s retry loop, and all existing log lines.
- `CIS_CENTRAL_BINDING` is still parsed: `btpLogin` uses `cisBinding.Uaa.Identityzoneid` as `--subdomain`.
- CLI deletes exit non-zero on failure and the output is included in the error, matching `deleteTrustConfigurations` (the old HTTP deletes returned nil even on non-2xx).

**Two disclosed decisions**

1. **Drive-by fix:** old `cleanup()` called `CleanUpSubaccounts(...)` before checking the `GetSubaccounts` error — on error `subaccounts` is nil and the range over `subaccounts.Value` panics. The error is now checked first.
2. **Left untouched:** `GetUaaAuthForTrustConfiguration` and `BtpSecuritySecret` appear to be pre-existing dead code (no callers found). They are not part of the runtime cleanup flow, so I left them for a maintainer to decide on — happy to remove them in a follow-up if wanted. (They are also what keeps the `net/http`, `net/url`, `encoding/base64` imports in use.)

**Verification**

- `gofmt`, `go vet` and `go build` clean on the file.
- Smoke run with no env fails gracefully at the first env parse, before any CLI/network call.
- The e2e cleanup itself needs a real BTP landscape, so it can only be exercised by CI. Note that `e2e_test.yaml` deliberately runs `cleanup.go` from the **base branch**, so this change first takes effect for the scheduled/merge workflows after it lands.

Dual-reviewed before posting (Claude + Codex gpt-5.6-sol at xhigh, unprimed): no P0/P1 findings.


AliAltivate:
[in-session-review sha=198bc16]

Dual review per our contribution pipeline (both reviewers unprimed):

- **Claude (Opus)** — semantic diff review against the old HTTP flow: filtering semantics (pending-deletion filter, BUILD_ID label checks, directory double-check), retry-loop structure, error handling, CLI syntax cross-checked against btp CLI v2.106.1 `--help`. No P0/P1.
- **Codex (gpt-5.6-sol, xhigh)** — self-contained diff review: "No P0/P1 findings. The CLI syntax and JSON shapes match the supplied v2.106.1 facts. Existing build-ID and pending-deletion filtering remain intact, the 30-attempt/45-second retry loop is preserved, and failed CLI operations remain retryable. The reordered GetSubaccounts error check correctly removes the prior nil-dereference risk."

Local verification: gofmt/go vet/go build clean; no-env smoke run fails gracefully at the first env parse. E2E itself requires a BTP landscape and runs from the base branch in `e2e_test.yaml`, so the new code first takes effect after merge.

xdj2e9 (approved):
LGTM

Thank you for the contribution.

PR #889, opened 2026-08-11
feat: backport terraform-provider-btp v1.25.0 to release/v1_10_x

### LINKED ISSUE #888: [BUG]In-place updates to ServiceInstance parameters are silently dropped on v1.11.0 - regression
**Describe the bug**
In-place updates to a ServiceInstance are not applied. After the instance is created, changing any of its parameters (e.g. storage or memory) has no effect on the actual BTP service instance , the reconcile fails and the instance stays on its original configuration. Updates of this kind worked in earlier provider versions, so this is a regression.

The failure appears to stem from the pinned terraform-provider-btp (1.23.1) btp_subaccount_service_instance resource: service_offering_name is Computed with a RequiresReplace plan modifier and is back-filled into state on refresh. On the next plan it recomputes and forces a destroy-and-recreate, which upjet blocks via its prevent_destroy lifecycle guard. The apply then errors and the update never reaches BTP.

Note: this appears to already be fixed upstream in terraform-provider-btp v1.25.0 (service_offering_name now uses UseStateForUnknown + RequiresReplaceIfConfigured), 

DO NOT INCLUDE INTERNAL INFORMATION, THIS IS A PUBLIC REPO. 

Tested with Provider Version

v1.11.0 (pins terraform-provider-btp 1.23.1). Not reproducible on v1.10.0 .

**To Reproduce**
Install crossplane-provider-btp v1.11.0.
Create a ServiceInstance refferencing any offering/plan so the plan ID is resolved and wait until it is Ready.
change a value under spec.forProvider.parameters (e.g. increase storage or memory).
Observe that the change is not applied:the underlying BTP service instance remains unchanged.

**Expected behavior**
Updating parameters on an existing ServiceInstance should update the corresponding BTP service instance in place (e.g. resize storage / memory vcpu), without attempting to destroy and recreate it.

**Screenshots**

**Additional context**

jtucci:
Fixes #888 by bumping terraform-provider-btp 1.23.1 → 1.25.0 (same bump as on main). No API/CRD changes.


xdj2e9:
Reopening PR to trigger fresh workflow runs.

xdj2e9 (approved):
Thank you for backporting!

PR #891, opened 2026-08-12
fix(ci): opt in to fork PR checkout for e2e

xdj2e9:
`actions/checkout` v7 refuses to check out fork code under `pull_request_target`, so e2e never runs on external PRs. Example: PR #889 (fork) - unit, license, CLA all pass; only e2e fails, at the `prepare-matrix` checkout step.

## Change

Add `allow-unsafe-pr-checkout: true` to the four checkout steps that use `inputs.checkout-ref` (prepare-matrix, fetch-chart, build, e2e-test) which are the ones that potentially pull forked code. v7 blocks these by default.

## Why this is safe

`prepare-matrix` runs in `pr-e2e-approval` (`required_reviewers`). A maintainer must approve before the job starts, and the fork checkout runs inside it - so fork code touches secrets only after a human reviewed the diff.

### DIFF
--- a/.github/workflows/e2e_test.yaml
+++ b/.github/workflows/e2e_test.yaml
@@ -68,6 +68,9 @@ jobs:
         with:
           #use inputs.checkout-ref to be able to call it from a pull_request_target workflow, since there it needs to be github.event.pull_request.merge_commit_sha and not the default
           ref: ${{ inputs.checkout-ref }}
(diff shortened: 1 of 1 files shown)


n-losse (approved):

n-losse:
doc link https://docs.github.com/en/actions/reference/security/securely-using-pull_request_target

xdj2e9:
> doc link https://docs.github.com/en/actions/reference/security/securely-using-pull_request_target

Thank you for adding the source of truth!

Issue #892, opened 2026-08-12
Preserve user-written duration literals in resource specs


Kukkerem:
## Problem

Duration fields in a spec are stored as `metav1.Duration`, which rewrites `720h` into `720h0m0s`
when the provider writes the resource. The provider also becomes the field manager of those fields
while doing so, so re-applying the unchanged manifest conflicts with the provider on values the user
never touched. Reported in #841.

Affected fields:

- `ServiceBinding.spec.rotation.frequency` / `.ttl`
- `KymaEnvironmentBinding.spec.forProvider.rotationInterval` / `.ttl` - their defaults `1h` and
  `1h15m` are themselves shorthand, so copying the defaults into a manifest reproduces it

The canonical form (`720h0m0s`) is documented as a workaround, but the API still rewrites what users
write.

## Goal

A duration value survives a provider write unchanged, so any valid Go duration string stays
apply-stable. Rotation timing and the CRD schema type (`string`) do not change.

## Approach

Use a provider-local duration type that remembers the literal it was decoded from and writes that
literal back, instead of `metav1.Duration`.

Objects created before the fix already store the rewritten value, so those need one
`--force-conflicts` apply or the canonical value. The fix prevents new occurrences.

## Docs to update when this lands

#893 documented the canonical form as the workaround. When the rewrite stops, that guidance goes with it:

- `apis/account/v1alpha1/servicebinding_types.go`: drop the canonical-form lines from the
  `RotationParameters.Frequency` / `.TTL` godoc, then regenerate
  `package/crds/account.btp.sap.crossplane.io_servicebindings.yaml`.
- `examples/sample/servicebinding.yaml`: drop the comment about shorthand values conflicting on re-apply.
- `docs/contribution-notes/service-binding-rotation.md`: drop the paragraph about re-serialization and field
  ownership.
- `docs/end-user-guides/services/consume-service.mdx`: shorten the rotation-conflict FAQ to a migration note.
  Bindings created by an older provider version still hold `720h0m0s` in their stored spec, with the provider
  recorded as the field manager of `.spec.rotation.frequency` / `.ttl`, so the first apply after the upgrade
  still needs the canonical value or `--force-conflicts`.

Repro, since no test in `test/e2e` or `test/upgrade` uses server-side apply today: apply the #841 manifest
twice with `kubectl apply --server-side` and expect no conflict, with `.spec.rotation.frequency` still `720h`
in the stored object.

Issue #841, opened 2026-07-28
[BUG] rotation fields conflict with themselves


gczuczy:
Migrated from internal bugreport:

### Priority

/priority medium

### When did it (start to) happen?

2026-07-28

### Which Provider is affected?

/provider BTP

### Which provider version do you use? 

1.9.0

### Steps to Reproduce

1. Add a `ServiceBinding` with the `.spec.rotation`'s 'frequency` and `ttl` fields set to the hour, like "720h", exactly
2. Re-apply the same manifest, observe the error messsage
3. Check the manifest, notice the values has been mutated, suffixed with a `0m0s`
4. Alter your original manifest, adding the zero minute and seconds
5. Notice it is now not conflicting with itself

The original error message on the conflict:
```
Error: UPGRADE FAILED: conflict occurred while applying object /x account.btp.sap.crossplane.io/v1alpha1, Kind=ServiceBinding: Apply failed with 2 conflicts: conflicts with "crossplane-provider-btp" using account.btp.sap.crossplane.io/v1alpha1:
- .spec.rotation.frequency
- .spec.rotation.ttl
```

### Manifests

```yaml
apiVersion: account.btp.sap.crossplane.io/v1alpha1
kind: ServiceBinding
metadata:
  name: "x"
spec:
  forProvider:
    name: x
    serviceInstanceRef:
      name: "x"
    subaccountRef:
      name: "x"
  writeConnectionSecretToRef:
    name: "x"
    namespace: x
  rotation:
    frequency: "720h"
    ttl: "1080h"
  providerConfigRef:
    name: "x"
```

### Additional Information

Idea is that upon resource creation the intervals are re-formatted with including everything up to the seconds. When re-applying the same manifest, with the same values, and error is returned, because it doesn't match the re-serialized value of itself.

Once appending the suffixes for zero minutes and seconds to the same values, the error is gone. So, this must be some missed case in the validation/mutating webhook logic.

Kukkerem:
Hey @gczuczy, confirmed, thanks for the precise repro! One correction: no webhook is involved. `metav1.Duration` re-encodes through Go's `time.Duration.String()` on the provider's first write, so your value gets rewritten and the provider takes ownership of the field, and that's what your later apply conflicts with. For now, use the canonical form. I'm working on the fix.

Kukkerem:
[...]

PR #893, opened 2026-08-12
docs: document canonical duration form for rotation fields

Kukkerem:
## Summary

`spec.rotation.frequency` and `spec.rotation.ttl` are `metav1.Duration`, which marshals through
`time.Duration.String()`. The provider re-serializes the whole object when it persists the external
name after `Create()` and when the finalizer is added, so `720h` is stored as `720h0m0s` and the
provider becomes the field manager of both fields. Every later server-side apply of the unchanged
manifest then conflicts with itself (#841). Nothing documented that, and the shipped examples used
the shorthand form, so users were led into it.

Documentation only. The provider-side fix is tracked in #892.

## Changes

- `RotationParameters.Frequency`/`.TTL` godoc: state the canonical form (`720h0m0s`, not `720h`); CRD regenerated.
- FAQ in the *Consume services* guide: the conflict error, its cause, and the form that avoids it.
- `docs/contribution-notes/service-binding-rotation.md`: explain the re-serialization and field ownership; spell out the durations in both example specs.
- `examples/sample/servicebinding.yaml`: canonical durations in the commented rotation block.


docs/end-user-guides/services/consume-service.mdx
```diff
@@ -57,4 +57,25 @@ The credentials are stored in a `Secret` you specify via `spec.writeConnectionSe
 
 ## ⁉ FAQs
 
-No FAQs yet. Got a question? Reach out to us and help us build this section.
+### Why does re-applying my `ServiceBinding` fail with a conflict on `spec.rotation`?
```
xdj2e9, line 60: Should we include the removal of this in the follow-up ticket #892?

xdj2e9 (approved):
LGTM. About the FAQ I feel it is not necessary anymore when we implemented supported for the short forms. WDYT?

docs/end-user-guides/services/consume-service.mdx
Kukkerem: Good idea, added it to the issue! I'd only shorten this FAQ rather than remove it, since bindings created before the fix still hold `720h0m0s`, and say a word about the migration.

PR #897, opened 2026-08-13
test: assert Subaccount and Directory external-name upgrade contracts

### LINKED ISSUE #504: [TASK] Create external-name upgrade tests for Subaccount and Directory
Custom upgrade tests should be put in place for the `Subaccount` and `Directory` resources to ensure external-name ADR compliance. 

## Acceptance Criteria
- [ ] `Test_Subaccount_External_Name` is adapted to reflect the recent external-name format change of `Subaccount` and `Directory`

Kukkerem:
## Summary

Both tests asserted only that the post-upgrade external-name parses as a UUID.
That condition holds whether or not a migration ran, so it did not constrain the
per-resource contract. Replaced with explicit pre- and post-upgrade assertions.

- **`Test_Subaccount_External_Name` (from v1.3.0):** v1.3.0 does not set the
  external-name, so crossplane-runtime's `NameAsExternalName` initializer leaves
  it at `metadata.name`. `migrateExternalName` resolves the account by
  `spec.forProvider.subdomain` + `region` and rewrites the annotation to the
  account GUID. The test now asserts the pre-upgrade sentinel, and that the
  migrated value is a UUID equal to the pre-upgrade
  `status.atProvider.subaccountGuid`.
- **`Test_Directory_External_Name` (from v1.5.0):** `CreateDirectory` already
  sets the annotation from `directory.Guid`, and the controller has no migration
  branch; a non-empty non-UUID external-name returns `errInvalidExternalName`.
  The test now asserts a GUID pre-upgrade and an unchanged value post-upgrade,
  both matching `status.atProvider.guid`. The doc comment is corrected to drop
  the migration path, which is not implemented for this resource.

Test-only change. No controller, client, fixture, Makefile, or workflow file is
touched, and both `FromVersion` tags remain the last release before the
respective format change.

## Test plan

- [x] `go test -c -tags=upgrade ./test/upgrade` and `go vet -tags=upgrade ./test/upgrade`

Closes #504

### DIFF
--- a/test/upgrade/directory_external_name_upgrade_test.go
+++ b/test/upgrade/directory_external_name_upgrade_test.go
@@ -20,15 +20,11 @@ var (
 	}
 )
 
-// Test_Directory_External_Name verifies that the Directory external-name is correctly
-// preserved and/or migrated during provider upgrades.
+// Test_Directory_External_Name verifies that upgrading from v1.5.0 preserves the directory GUID
+// external-name and keeps it aligned with status.atProvider.guid.
 //
-// The Directory controller has backwards compatibility logic that migrates
-// external-names from old formats (metadata.name prior to ADR compliance) to GUID format.
-// This test ensures that:
-// 1. The external-name exists before upgrade
-// 2. After upgrade, the external-name is in GUID format (UUID)
-// 3. The resource remains healthy after upgrade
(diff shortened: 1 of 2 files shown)


xdj2e9 (approved):
LGTM

PR #911, opened 2026-08-17
fix(kyma): set landscapeLabel on KymaEnvironment create

### LINKED ISSUE #910: [BUG] KymaEnvironment create fails with "Missing required parameter: landscape (landscapeLabel)"
**Describe the bug**

Creating a `KymaEnvironment` fails against the BTP provisioning API when the region has more than
one landscape:

```
create failed: while creating instance: Could not create KymaEnvironment:
{"error":{"code":11005,"message":"Missing required parameter: landscape (landscapeLabel)",
"target":"/provisioning/v1/environments"}}
```

Root cause: `CreateKymaEnvironment` (`btp/kyma_environment.go`) never sets `landscapeLabel` on the
create payload, so the field is omitted from the request. `landscapeLabel` is a top-level field on
the provisioning create payload, documented as "only required if the region has multiple
landscapes" - single-landscape regions default it server-side, multi-landscape regions reject the
create (code 11005). The `KymaEnvironment` CRD has no field for it either, so users can't supply
it. `CloudFoundryEnvironment` already handles this via `forProvider.landscape`.

v1.12.0 (also reproduces on v1.8.0)

**To Reproduce**

1. Create a `KymaEnvironment` in a subaccount whose region has multiple landscapes.
2. Wait for the provider to attempt creation.
3. See the create fail with `Missing required parameter: landscape (landscapeLabel)` (code 11005).

**Expected behavior**

The provider lets the user specify the target landscape and passes it to the provisioning API, so
the environment is created on multi-landscape regions.

**Additional context**

Suggested fix - mirror the CloudFoundry pattern:
- Add optional `landscapeLabel` to `KymaEnvironmentParameters`.
- Set it on the create payload (`payload.LandscapeLabel`).
- Create-only: the update (PATCH) payload has no landscape field - landscape is fixed at create.

Users on single-landscape regions leave it unset (unchanged behaviour); users on multi-landscape
regions set it to their landscape.

xdj2e9:
Closes #910.

## Summary

- **Add optional `forProvider.landscapeLabel` to `KymaEnvironment`**: `KymaEnvironmentParameters` had no landscape field, so users could not target a landscape on multi-landscape regions.
- **Set `landscapeLabel` on the create payload**: the create path never set it, so it was dropped from the request (`omitempty`), causing `Missing required parameter: landscape (landscapeLabel)` (code 11005) on multi-landscape regions. Mirrors `CloudFoundryEnvironment`'s `forProvider.landscape`.
- **Create-only**: the provisioning PATCH payload has no landscape field as it is fixed at create.
- **Not breaking for existing resources**: the field is optional and unset passes a nil pointer that serializes as absent, so existing `KymaEnvironment`s send the exact same create request as before. Behaviour changes only when a user sets it.
- **Unit test via httptest, not e2e**: e2e can't cover this (we don't run against a multi-landscape subaccount). It is tested against an `httptest` server and asserts `landscapeLabel` is part of the actual payload when set and not part of it when unset. Method used before in `internal/clients/entitlement/entitlement_test.go`.

## User-visible changes

| --- | --- | --- |
| `KymaEnvironment.spec.forProvider.landscapeLabel` | new optional string | Unset = unchanged behaviour (single-landscape regions default it server-side). Set to the target landscape on multi-landscape regions. Immutable after create in practice: the API ignores it on update. |

### DIFF
--- a/apis/environment/v1alpha1/kymaenvironment_types.go
+++ b/apis/environment/v1alpha1/kymaenvironment_types.go
@@ -27,6 +27,13 @@ type KymaEnvironmentParameters struct {
 	// +kubebuilder:validation:Optional
 	Name *string `json:"name"`
 
+	// LandscapeLabel is the name of the landscape within the logged-in region on which the
+	// environment instance is created. Only required when the region has more than one landscape;
+	// single-landscape regions default it server-side. Set at create only - landscape is fixed
+	// once the environment exists.
+	// +kubebuilder:validation:Optional
+	LandscapeLabel *string `json:"landscapeLabel,omitempty"`
+
 	// Provisioning parameters for the instance.
 	//
 	// The Parameters field is NOT secret or secured in any way and should
--- a/apis/environment/v1alpha1/kymaenvironment_types.go
+++ b/apis/environment/v1alpha1/kymaenvironment_types.go
@@ -32,6 +32,8 @@ type KymaEnvironmentParameters struct {
 	// single-landscape regions default it server-side. Set at create only - landscape is fixed
 	// once the environment exists.
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="landscapeLabel is immutable after creation"
 	LandscapeLabel *string `json:"landscapeLabel,omitempty"`
 
 	// Provisioning parameters for the instance.
--- a/apis/environment/v1alpha1/zz_generated.deepcopy.go
+++ b/apis/environment/v1alpha1/zz_generated.deepcopy.go
@@ -586,6 +586,11 @@ func (in *KymaEnvironmentParameters) DeepCopyInto(out *KymaEnvironmentParameters
 		*out = new(string)
 		**out = **in
 	}
+	if in.LandscapeLabel != nil {
+		in, out := &in.LandscapeLabel, &out.LandscapeLabel
+		*out = new(string)
+		**out = **in
+	}
 	in.Parameters.DeepCopyInto(&out.Parameters)
 }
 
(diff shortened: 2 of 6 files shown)


apis/environment/v1alpha1/kymaenvironment_types.go
```diff
@@ -27,6 +27,13 @@ type KymaEnvironmentParameters struct {
 	// +kubebuilder:validation:Optional
 	Name *string `json:"name"`
 
+	// LandscapeLabel is the name of the landscape within the logged-in region on which the
+	// environment instance is created. Only required when the region has more than one landscape;
+	// single-landscape regions default it server-side. Set at create only - landscape is fixed
+	// once the environment exists.
+	// +kubebuilder:validation:Optional
+	LandscapeLabel *string `json:"landscapeLabel,omitempty"`
```
mpechkurov, line 37: The field is documented as create-only, but without an XValidation rule Kubernetes will accept changes to it on existing resources. The reconciler will silently ignore the new value (update API has no landscape field), leaving spec and actual state diverged with no error surfaced.

Suggested fix:
```go
// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="landscapeLabel is immutable after creation"
```

mpechkurov, line 37: `*string` allows `landscapeLabel: ""` — an empty string pointer is not nil, so it serializes as `"landscapeLabel": ""` and BTP rejects it with error 11005, the exact error this PR fixes. A user who sets the field but leaves it blank gets the same confusing failure as before.

Suggested fix: add `+kubebuilder:validation:MinLength=1`, consistent with `CloudFoundryEnvironment.landscape`.

xdj2e9:
@mpechkurov 
Thank you for the feedback. I agree that these are very sensible safeguards. I added them. Could you recheck?

mpechkurov (approved):

PR #914, opened 2026-08-18
fix(kyma): add landscapeLabel support for multi-landscape regions

### LINKED ISSUE #910: [BUG] KymaEnvironment create fails with "Missing required parameter: landscape (landscapeLabel)"

dwlou:
Fixes #910

## Summary
- Add optional `landscapeLabel` field to `KymaEnvironmentParameters` (`forProvider`)
- Pass it through to the provisioning API create payload (`CreateEnvironmentInstanceRequestPayload.LandscapeLabel`)
- Single-landscape regions are unaffected — leaving the field unset omits it from the payload (same behaviour as before)

> **Note:** `zz_generated.deepcopy.go` and CRD manifests still need to be regenerated (`make generate`) before merging.

## Test plan
- [ ] Create a `KymaEnvironment` without `landscapeLabel` on a single-landscape region — should work as before
- [ ] Create a `KymaEnvironment` with `landscapeLabel` set on a multi-landscape region — should no longer fail with error code 11005

### DIFF
--- a/apis/environment/v1alpha1/kymaenvironment_types.go
+++ b/apis/environment/v1alpha1/kymaenvironment_types.go
@@ -35,6 +35,11 @@ type KymaEnvironmentParameters struct {
 	// in a Secret and use the ParametersFrom field.
 	// +kubebuilder:pruning:PreserveUnknownFields
 	Parameters runtime.RawExtension `json:"parameters,omitempty"`
+
+	// The name of the landscape within the logged-in region on which the environment instance is created.
+	// Only required if the region has multiple landscapes.
+	// +kubebuilder:validation:Optional
+	LandscapeLabel *string `json:"landscapeLabel,omitempty"`
 }
 
 // KymaEnvironmentObservation are the observable fields of a KymaEnvironment.
--- a/btp/kyma_environment.go
+++ b/btp/kyma_environment.go
@@ -40,7 +40,7 @@ func (c *Client) GetKymaEnvironment(
 	return c.GetEnvironmentByNameAndType(ctx, instanceName, environmentType)
 }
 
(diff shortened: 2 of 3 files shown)


dwlou:
duplicate #911

PR #915, opened 2026-08-18
chore(config): enable and group dependency updates for v1 release branch

xdj2e9:
Enable renovate dependency updates on the `release/v1_10_x` branch. Restricted to patch/digest updates (no major/minor), with ungrouped deps folded into one "other dependencies" PR.

Dry-run:
- `main` was unchanged (still 10 existing PRs)
- `release/v1_10_x` branch gets 4 grouped PRs, all patch/digest

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -5,6 +5,7 @@
     ":dependencyDashboard",
     "schedule:weekly"
   ],
+  "baseBranches": ["main", "release/v1_10_x"],
   "ignorePresets": ["group:goOpenapi"],
   "postUpdateOptions": [
     "gomodTidy"
@@ -13,6 +14,11 @@
   "prConcurrentLimit": 10,
   "prHourlyLimit": 0,
   "packageRules": [
+    {
+      "matchBaseBranches": ["release/v1_10_x"],
+      "matchUpdateTypes": ["major", "minor"],
+      "enabled": false
+    },
     {
       "groupName": "crossplane and upjet",
       "matchPackageNames": [
@@ -40,6 +46,10 @@
     {
       "groupName": "github actions",
       "matchManagers": ["github-actions"]
+    },
+    {
+      "matchBaseBranches": ["release/v1_10_x"],
+      "groupName": "other dependencies"
     }
   ],
   "customManagers": [
--- a/renovate.json
+++ b/renovate.json
@@ -5,7 +5,7 @@
     ":dependencyDashboard",
     "schedule:weekly"
   ],
-  "baseBranches": ["main", "release/v1_10_x"],
+  "baseBranchPatterns": ["main", "release/v1_10_x"],
   "ignorePresets": ["group:goOpenapi"],
   "postUpdateOptions": [
     "gomodTidy"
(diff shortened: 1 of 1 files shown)


n-losse (dismissed):

renovate.json
```diff
@@ -40,6 +46,10 @@
     {
       "groupName": "github actions",
       "matchManagers": ["github-actions"]
+    },
+    {
+      "matchBaseBranches": ["release/v1_10_x"],
```
Kukkerem: Isn't this rule in the wrong place? Renovate merges them in order, with later ones winning. This one has no package match, so on `release/v1_10_x` it matches everything and overwrites `groupName` on all five groups above it. Shouldn't we move it under the first rule?

renovate.json
```diff
@@ -5,6 +5,7 @@
     ":dependencyDashboard",
     "schedule:weekly"
   ],
+  "baseBranches": ["main", "release/v1_10_x"],
```
Kukkerem: Wouldn't `baseBranchPatterns` be the current name?

renovate.json
```diff
@@ -13,6 +14,11 @@
   "prConcurrentLimit": 10,
   "prHourlyLimit": 0,
   "packageRules": [
+    {
+      "matchBaseBranches": ["release/v1_10_x"],
```
Kukkerem, line 18: Wouldn't an allow-list be safer here? Denying only `major`/`minor` still lets `replacement` through, and `config:recommended` pulls in `replacements:all`.
```
    { "matchBaseBranches": ["release/v1_10_x"], "enabled": false },
    { "matchBaseBranches": ["release/v1_10_x"], "matchUpdateTypes": ["patch", "digest"], "enabled": true },
```

renovate.json
```diff
@@ -13,6 +14,11 @@
   "prConcurrentLimit": 10,
```
Kukkerem: Is this enough? We have 7 open PRs currently.

renovate.json
xdj2e9: No overwrite. At least that is what my dry-run with renovate:41 showed. Group rules win over a broader/later rule regardless of position. In the dry-run crossplane, kubernetes, github-actions each keep their own release branch; the catch-all only groups "single" dependencies.

Still I restructured anyway. WDYT?

renovate.json
xdj2e9: Adapted

renovate.json
xdj2e9: Confirmed, config:recommended extends replacements:all. I tried the unfiltered enabled:false and then re-enable approach, but ended up with 0 PRs in the dry-run. enabled:false with explicit type list worked. Dry-run showed v1 release gets patch/digest only.

renovate.json
xdj2e9: I agree. With the current configuration we may easily exceed 10. My suggestion: raise it to 20 for now and also adapt grouping for main in a follow up PR (which then could reduce this again). WDYT?

Kukkerem: 20 sounds good!

renovate.json
Kukkerem: It looks right, thanks! The docs say the opposite, but either way, the new order looks good.

Kukkerem (approved):
LGTM :fire:

PR #916, opened 2026-08-18
fix: preserve duration literals in resource specs

### LINKED ISSUE #892: Preserve user-written duration literals in resource specs

Kukkerem:
## Summary

`ServiceBinding.spec.rotation.frequency`/`.ttl` and `KymaEnvironmentBinding`'s
`rotationInterval`/`ttl` were `metav1.Duration`, which marshals through
`time.Duration.String()`. Every provider write - persisting the external name after
`Create()`, adding the finalizer - rewrote `720h` to `720h0m0s` and made the provider
field manager of a value the user never wrote, so re-applying the unchanged manifest
conflicted (#841).

`v1alpha1.Duration` remembers the literal it decoded and writes that back. Rotation
timing and the CRD schema type (`string`) are unchanged.

Objects created by an older provider still hold `720h0m0s` with the provider recorded
as field manager, so the first apply after upgrading still needs the canonical value
or `--force-conflicts`.

## Changes

- `apis/v1alpha1/duration.go`: embeds `time.Duration` and keeps the decoded literal.
  `MarshalJSON`/`ToUnstructured`/`String` return it, `ToUnstructured` being the path
  structured-merge-diff takes. The literal is only used while the embedded value still
  equals what was decoded, so assigning `.Duration` in Go falls back to canonical form
  rather than serializing a stale literal. `Equal` compares elapsed time only, which
  go-cmp needs to walk the unexported fields. Deepcopy is generated.
- `RotationParameters.Frequency`/`.TTL` and
  `KymaEnvironmentBindingParameters.RotationInterval`/`.BindingTTl` switch to it.
  Deepcopy regenerated; both CRDs keep `type: string`.
- Round-trip tests for the type and for `ServiceBinding`; existing tests moved off
  `metav1.Duration`.
- Drop the canonical-form guidance added by #893: `RotationParameters` godoc (CRD
  regenerated), the sample manifest, the rotation contribution notes, and the
  *Consume services* FAQ, which becomes a migration note for bindings that still hold
  a rewritten value.

## Test plan

- [x] `make reviewable`
- [ ] apply the #841 manifest twice with `kubectl apply --server-side`: no conflict,
      `.spec.rotation.frequency` still `720h`

Closes #892


apis/v1alpha1/duration.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// ToUnstructured preserves the original literal during structured-to-unstructured conversion.
+func (d Duration) ToUnstructured() interface{} {
+	return d.String()
+}
+
+// Equal compares elapsed time and ignores the preserved literal.
+// go-cmp uses this method because Duration contains unexported state.
+func (d Duration) Equal(o Duration) bool { return d.Duration == o.Duration }
```
SatabdiG, line 53: **Nit**: Equal() is correctly wired up for go-cmp (nice touch documenting why it's needed given the unexported fields). One thing worth a quick check before merge: is go-cmp/cmp.Diff the only place RotationParameters/KymaEnvironmentBindingParameters get compared? If anything elsewhere uses reflect.DeepEqual directly (or testify's assert.Equal), it'll see text/parsed as live struct fields and could report a false diff that never goes through Equal().

SatabdiG (approved):
lgtm 🔥 solid pr

apis/v1alpha1/duration.go
Kukkerem: Checked: no `reflect.DeepEqual` in the repo reaches these types (all 20 callsites are labels, scalar pointers and GUIDs), and the testify asserts in `keyrotator_test.go` are on scalars only. The only comparison that walks them is `cmp.Diff` on whole CRs in `servicebinding_test.go`, which honours `Equal`. Removing `Equal` makes go-cmp panic on the unexported field, so it can't silently false-diff.

PR #922, opened 2026-08-18
fix(cloudmanagement): self-adopt a truncated compound external-name

jn-av:
## Problem

A `CloudManagement`'s `external-name` encodes two IDs as `instanceID/bindingID`. In this failure mode it is truncated down to the **bare instance UUID** (the binding-ID segment is lost) while the binding still exists and is ready in BTP. `Observe` then reports `ResourceExists=false`, so the CR loops forever on `API Error Creating Resource Service Binding (Subaccount): Conflict` — the binding already exists in BTP, so the re-create 409s. Downstream, the connection secret is never written, so consumers of the CloudManagement credentials cannot proceed.

The prevention fix (`Create()` no longer truncates the external-name) is already merged, but the truncated state can still be reached — so this adds a **recovery** path that heals it instead of requiring a manual `kubectl annotate`.

## Why the existing recovery path did not fire

`healExternalName` already does the right thing on a match (`LookupInstanceAndBinding` → rewrite the full `instanceID/bindingID`). Two gates blocked it for this state:

1. **`recovery.IsFallbackExternalName`** treats only `"" || == metadata.name` as a fallback; a bare-UUID external-name (the legitimate phase-1 intermediate) is deliberately not a fallback, so `healExternalName` was never called.
2. **`recovery.IsOwnedByCR`** (time window vs `external-create-pending`) cannot pass here: in the Conflict retry loop `create-pending` is rewritten to "now" on every retry while the instance `created_at` stays fixed, so the window flips to brownfield within ~1 minute of instance birth.

## Fix

Two small helpers + wiring; `healExternalName` itself is unchanged.

- **`recovery.IsTruncatedCompoundExternalName`** — recognizes the bare-UUID compound state (non-empty, `!= metadata.name`, no `/`). The Observe guard now fires on it **in addition to** the fallback case, and only when `ResourceExists == false` (so a healthy phase-1→phase-2 create is untouched).
- **`recovery.IsOwnedByExternalNameInstanceID`** — ownership proof for the truncated path: the instance UUID our own phase-1 `Create` already wrote into `external-name` equals the instance the spec-scoped semantic lookup (plan + name, unique) finds in BTP. Used **instead of** the time window on this path. The fallback path keeps `IsOwnedByCR` unchanged.
- The instance-ID proof additionally requires that the lookup found a **real binding** (`sbID != ""`). This keeps a healthy phase-1 (instance in BTP, binding not yet created → `sbID == ""`) from healing to the same bare UUID and spuriously requeuing every reconcile.

## Tests

- `recovery_test.go`: `IsTruncatedCompoundExternalName` (bare-UUID→true; compound/empty/==name→false); `IsOwnedByExternalNameInstanceID` (match/empty/differ).
- `cloudmanagement_recovery_test.go`: truncated external-name + broken time window + instance-ID match → heals to compound name, requeues; truncated + mismatched instance ID → refused brownfield; bare-UUID with `ResourceExists=true` → recovery not triggered; healthy phase-1 (bare UUID, no binding in BTP) → does not heal. Existing fallback-path tests unchanged.
- `go build ./...`, `go vet`, and `go test ./internal/recovery/ ./internal/controller/account/... ./internal/clients/servicemanager/...` green.
- Verified end-to-end against a real BTP subaccount on a kind cluster: truncating a healthy CloudManagement's external-name to the bare instance UUID (with instance+binding IDs also cleared from status to exclude the migration path) heals back to the compound name purely via the BTP lookup.

## Out of scope

The halted `ExternalCreateIncomplete` variant (`create-pending` newer than `create-failed`, no `create-succeeded`) is frozen by crossplane-runtime **before** `Observe` runs (`Requeue:false`, "cannot determine creation result"), so an Observe-based heal cannot reach it.

### DIFF
--- a/internal/controller/account/cloudmanagement/cloudmanagement.go
+++ b/internal/controller/account/cloudmanagement/cloudmanagement.go
@@ -206,13 +206,13 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 		return managed.ExternalObservation{}, errors.Wrap(statusErr, errSetStatus)
 	}
 
-	// Recovery: BTP has a resource matching this managed CR, but our
-	// external-name is still a fallback — semantic lookup + ownership check
-	// (see internal/recovery). Fires only on TRUE fallback: a single-UUID
-	// external-name is the natural output of phase-1 Create and must NOT
-	// re-trigger recovery (would trap CM in an infinite loop before phase-2).
+	// Recovery: BTP has a matching resource but our external-name is a fallback
+	// or a truncated compound (bare instance UUID). Fires only when
+	// ResourceExists is false, so a healthy phase-1→phase-2 create is untouched.
+	extName := meta.GetExternalName(cr)
 	if err == nil && !resStatus.ResourceExists &&
-		recovery.IsFallbackExternalName(cr.Name, meta.GetExternalName(cr)) {
+		(recovery.IsFallbackExternalName(cr.Name, extName) ||
+			recovery.IsTruncatedCompoundExternalName(cr.Name, extName)) {
 		if healErr := c.healExternalName(ctx, cr); healErr != nil {
 			return managed.ExternalObservation{}, healErr
 		}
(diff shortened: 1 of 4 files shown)


internal/controller/account/cloudmanagement/cloudmanagement.go
```diff
@@ -269,9 +269,26 @@ func (c *external) healExternalName(ctx context.Context, cr *apisv1beta1.CloudMa
...
+		return nil
+	}
+
+	// Ownership proof, either of:
+	//  - the time window vs external-create-pending (IsOwnedByCR); or
+	//  - the truncated-compound match: external-name is truncated (existingBID
+	//    == ""), the lookup found a real binding (sbID != ""), and the instance
+	//    UUID we hold equals the found instance.
+	ownedByTime := recovery.IsOwnedByCR(cr, instanceCreatedAt)
+	ownedByTruncatedMatch := existingBID == "" && sbID != "" &&
```
Kukkerem: Could the ownership check branch on `truncated`? `existingBID == ""` is always true here because the caller admits only slash-free fallback or truncated names, so the instance-ID proof also applies to fallback names. The current `||` also lets the time window accept a truncated name whose instance ID does not match `siID`. Fallback should use `IsOwnedByCR`. The truncated path should require a binding and a matching instance ID.

Kukkerem (commented):
Great work! Only one question :fire:

internal/controller/account/cloudmanagement/cloudmanagement.go
jn-av: Good catch. Fixed: ownership now branches on `truncated` — truncated names use instance-ID match only, fallback names use the time window only. Also dropped the now-unused `existingBID` from `splitExternalName`. Same fix applied to the SM branch.

mpechkurov (approved):

PR #924, opened 2026-08-19
fix(servicemanager): self-adopt a truncated compound external-name

jn-av:
## Problem

A `ServiceManager`'s `external-name` encodes two IDs as `instanceID/bindingID`. It can be truncated down to the **bare instance UUID** (binding-ID segment lost) while the binding still exists and is ready in BTP. `Observe` then reports `ResourceExists=false`, so the CR loops forever on `API Error Creating Resource Service Binding (Subaccount): Conflict` — the binding already exists in BTP, so the re-create 409s.

This is the same failure mode and the same fix as the CloudManagement recovery path — ServiceManager is the other two-phase resource with a compound external-name.

## Why the existing recovery path did not fire

`healExternalName` already does the right thing on a match (`LookupInstanceAndBinding` → rewrite the full `instanceID/bindingID`). Two gates blocked it for this state:

1. **`recovery.IsFallbackExternalName`** treats only `"" || == metadata.name` as a fallback; a bare-UUID external-name (the legitimate phase-1 intermediate) is deliberately not a fallback, so `healExternalName` was never called.
2. **`recovery.IsOwnedByCR`** (time window vs `external-create-pending`) cannot pass in the Conflict retry loop: `create-pending` is rewritten to "now" on every retry while the instance `created_at` stays fixed, so the window flips to brownfield within ~1 minute.

## Fix

- **`recovery.IsTruncatedCompoundExternalName`** — recognizes the bare-UUID compound state. The Observe guard now fires on it **in addition to** the fallback case, and only when `ResourceExists == false` (so a healthy phase-1→phase-2 create is untouched).
- **`recovery.IsOwnedByExternalNameInstanceID`** — ownership proof for the truncated path: the instance UUID our own phase-1 `Create` already wrote into `external-name` equals the instance the spec-scoped semantic lookup (plan + name, unique) finds in BTP. Used **instead of** the time window on this path. The fallback path keeps `IsOwnedByCR` unchanged.
- The truncated path additionally does nothing when the lookup found no binding (`sbID == ""`): a healthy phase-1 (instance created, binding not yet) is in the time window, and without this guard it would rewrite external-name to the same bare UUID and requeue, starving phase-2.

`healExternalName` itself is unchanged.

## Tests

- `recovery_test.go`: `IsTruncatedCompoundExternalName`, `IsOwnedByExternalNameInstanceID`.
- `servicemanager_recovery_test.go`: truncated external-name + broken time window + instance-ID match → heals to compound name, requeues; truncated + mismatched instance ID → refused brownfield; healthy phase-1 (fresh instance, no binding) → does not heal; bare-UUID with `ResourceExists=true` → recovery not triggered. Existing fallback-path tests unchanged.
- `go build ./...`, `go vet`, and `go test ./internal/recovery/ ./internal/controller/account/servicemanager/ ./internal/clients/servicemanager/` green.

## Note on shared helpers

The two `recovery` helpers are also added by the CloudManagement fix in a separate PR. They are identical; whichever lands first brings them to `main`, and the other rebases onto it with a trivial (identical) resolution.

### DIFF
--- a/internal/controller/account/servicemanager/servicemanager.go
+++ b/internal/controller/account/servicemanager/servicemanager.go
@@ -220,7 +220,7 @@ func (c *external) healExternalName(ctx context.Context, cr *apisv1beta1.Service
 		return nil
 	}
 
-	externalNameInstanceID, existingBID, _ := strings.Cut(meta.GetExternalName(cr), "/")
+	externalNameInstanceID, _, _ := strings.Cut(meta.GetExternalName(cr), "/")
 
 	// Truncated external-name (bare instance UUID) with no binding found in BTP
 	// is a healthy phase-1: the instance UUID is already correct and there is no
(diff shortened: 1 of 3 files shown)


enrico-kaack-comp (approved):
LGTM

PR #925, opened 2026-08-19
feat(servicemanager): change default planName to subaccount-admin

### LINKED ISSUE #845: Change default ServiceManager planName from service-operator-access to subaccount-admin
## Context

The `ServiceManager` resource defaults to `planName: service-operator-access`. This plan is intended **exclusively** for the SAP [BTP Service Operator ](https://github.com/SAP/sap-btp-service-operator/blob/main/README.md#installation-and-setup) and is *opinionated* for this specific usage context, which can leads to unexpected problem. (see discussion #844). 

## Change

- Update the default `planName` on the `ServiceManager` CRD from `service-operator-access` to `subaccount-admin`.
- Create a migration guide for existing customers who are already productive with `service-operator-access`

dwlou:
Fixes #845

## Summary
- Change `DefaultPlanName` constant from `service-operator-access` to `subaccount-admin` in `ServiceManagerParameters`
- Update CRD manifest (`package/crds/account.btp.sap.crossplane.io_servicemanagers.yaml`) via `make generate`
- Add migration guide at `docs/end-user-guides/account/servicemanager-planname-migration.md`

**Existing resources are unaffected** — `planName` is immutable once set (CEL rule `self == oldSelf`). Only new resources without an explicit `planName` will get the new default.

## Test plan
- [ ] Create a new `ServiceManager` without `planName` — verify it defaults to `subaccount-admin`
- [ ] Create a new `ServiceManager` with `planName: service-operator-access` explicitly — verify it still works
- [ ] Verify existing `ServiceManager` resources are not affected on upgrade

```release-note
feat(servicemanager): change default planName from service-operator-access to subaccount-admin — existing resources are unaffected; users relying on the old default who need service-operator-access must now set it explicitly
```


apis/account/v1beta1/servicemanager_types.go
```diff
@@ -72,7 +72,7 @@ type ServiceManagerParameters struct {
 	// Planname for service manager instance
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:Enum=subaccount-admin;service-operator-access;container;subaccount-audit
-	// +kubebuilder:default:=service-operator-access
+	// +kubebuilder:default:=subaccount-admin
```
SatabdiG, line 75: Since`DefaultPlanName` is live in the reconcile path. Maybe we should have a unit test in `ServicePlanName()`  (table test: empty PlanName → subaccount-admin; explicit PlanName → passthrough)

SatabdiG (commented):
🔥 just a small comment from my side

SatabdiG (approved):
🔥

PR #926, opened 2026-08-19
chore(renovate): reduce dependency-update PR volume

xdj2e9:
## Why

Renovate opens too many small PRs. Every batch also runs the e2e cluster, so both cost and manual-merge work scale with PR count. Three sources drive the volume: standalone go-module bumps each get their own PR, updates land weekly, and prior batches often needed hand-reverting of day-old deps. This PR addresses all three - fewer batches, loose modules folded into one PR, and a release-age hold - while keeping majors isolated so breaking bumps stay reviewable.

## Changes

- **Schedule:** replace `schedule:weekly` with `"schedule": ["* * 1,15 * *"]` (1st + 15th).
- **Grouping:** add an `"other dependencies"` group (`matchManagers: ["gomod"]`, non-major types) that collapses the standalone modules (`go-cfclient/v3`, `logr`, `kubelogin`, `golang.org/x/*`, ...) into one PR; majors like `jwt/v4 -> v5` stay split.
- **Base preset:** swap `config:recommended` for `config:best-practices` (adds `docker:pinDigests`, `helpers:pinGitHubActionDigests`, `:configMigration`; npm/lockfile presets are no-ops here).
- **Digest batching:** add a `"docker images"` group so the new Docker digest bumps batch; action digest pins already batch in the existing `"github actions"` group.
- **Release-age hold:** add `minimumReleaseAge: "7 days"` + `internalChecksFilter: "strict"` to hold fresh releases.

## Notes

- Verified with `renovate-config-validator --strict` and a local `--dry-run=full`.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,10 +1,12 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
-    "config:recommended",
+    "config:best-practices",
     ":dependencyDashboard"
   ],
   "schedule": ["* * 1,15 * *"],
+  "minimumReleaseAge": "7 days",
+  "internalChecksFilter": "strict",
   "baseBranchPatterns": ["main", "release/v1_10_x"],
   "ignorePresets": ["group:goOpenapi"],
   "postUpdateOptions": [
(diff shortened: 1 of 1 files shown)


SatabdiG (approved):
lgtm

PR #928, opened 2026-08-19
fix(subaccountapicredential): set key as sensitive

dwlou:
## Summary

- Mark the `key` attribute as sensitive in `SubaccountApiCredential` so upjet routes it to the connection secret instead of the observed status
- Update generated types, connection-details mapping, deep-copy code, and CRD schema accordingly
- `status.atProvider.key` was not referenced anywhere in the codebase, so no consumer code is affected

## Test plan

- [ ] Verify `attribute.key` is present in the connection secret for certificate-based credentials
- [ ] Verify `status.atProvider.key` is empty after reconciliation


mpechkurov (approved):

xdj2e9 (approved):
LGTM.

One thought on the test plan, not blocking: both items are manual checks. Should we open a follow-up issue to add e2e coverage for the certificate scenario - assert `status.atProvider.key` is absent after sync and `attribute.key` is present in the connection secret?

dwlou:
> LGTM.
> 
> One thought on the test plan, not blocking: both items are manual checks. Should we open a follow-up issue to add e2e coverage for the certificate scenario - assert `status.atProvider.key` is absent after sync and `attribute.key` is present in the connection secret?

good point .. i think manual check is sufficient. `status.asProvider` is written with every `observe`.

PR #933, opened 2026-08-20
ci: remove obsolete workflow files

n-losse:
both not running anyway, and not needed


jn-av (approved):
good catch, we can get rid of it. Thanks!

PR #935, opened 2026-08-24
ci: prepare for shared workflows 

n-losse:
https://github.tools.sap/cloud-orchestration/provider-stream-items/issues/139

In order to have shared workflow files for all the open source crossplane providers in this org, start moving logic into callable workflows. For now, these stay here to simplify testing. They can be moved to a dedicated repo later if needed, or be called by the other providers from here. 

Convention of using *.lib.yaml for call-only workflows copied from https://github.com/openmcp-project/build/blob/main/.github/workflows

This also tries to make naming more consistent.

# Note

Since this changes the job names and nesting level, this will require the branch protection rules to be updated. 

Curreny rules are 

image

Rule names are job names only, workflow names do not matter.


.github/workflows/check-unit-tests.lib.yaml
```diff
@@ -1,31 +1,29 @@
...
-    - name: Build
+    - name: go build
       run: go build -v ./...
 
-    - name: Unit-Test Init
-      run: make test.init
-    - name: Unit-Test Run
-      run: make test.run
+    - name: go test
+      run: go test -v ./...
```
xdj2e9, line 29: The `make test.run` call also included coverage and test log artifact creation. Is it intentional to drop this?

.github/workflows/check-reviewable-diff.lib.yaml
```diff
@@ -23,9 +30,10 @@ jobs:
     - name: Install Go tools
       run: go install tool
 
-    - name: Make Reviewable
+    - name: make reviewable
```
xdj2e9, line 33: Although it is already like this in current main: `make reviewable` also runs the unit tests meaning we run them always twice. Maybe that would be worth to have a look at while doing this rework?

xdj2e9 (commented):
Nice restructureing. Looks good already in general to me. Some points need clarification for me.

.github/workflows/check-reviewable-diff.lib.yaml
n-losse: yes, i did not drop the doubling for now, i just had a comment about this in check-unit-tests.lib.yaml 

i figured the "bare" unit tests in the other one give quick feedback, and this one is then more thorough. that is also why i changed the bare unit tests to just go test, so there's less doubling

but of course we could also drop that, and just keep `reviewable` , open to that

.github/workflows/check-unit-tests.lib.yaml
n-losse: see the other comment

xdj2e9 (approved):

PR #936, opened 2026-08-24
fix/improve: service binding leak after pod restart and create idempotency

enrico-kaack-comp:
## Observations

1) Retired service bindings could silently leak in Service Manager. The deletion path relied on Terraform's destroy. But a destroy run against an empty or cold Terraform state exits successfully without calling delete.
 For example after a pod restart, where the workspace never observed the binding's GUID while it was current. 
The provider treated that success from tf as "deleted," pruned its retiredBinding list, and the binding lived on in BTP with no record that anything was wrong.

2) Create was not idempotent across retries for rotated bindings. Each Create attempt used a new random name. If an attempt called Service Manager, created the binding, and then lost its result before persisting the external-name (crash or requeue in that window), the next reconcile generated a new name and created a second binding _ leaking the first.

Basically any kind of interruption between 2 and 4 resulted in a binding that the provider never tracked or deleted:
```
1) Create() call
2) new random name
3) call btp
4) write response in CR
``` 

3) Deletion failures were invisible. When a retired key could not be deleted, it was quietly kept in the list for the next attempt with no signal that it was overdue or repeatedly failing.

## How this PR addresses them

1) Three-phase, verified deletion. Deletion now runs as seed _ destroy _ verify:
  - Seed: connect without marking for deletion so Terraform state is populated with the binding's id.
  - Destroy: reconnect with the deletion mark set, so prevent_destroy is off and the destroy runs against real state, initialised tf state.
  - Verify: re-observe the binding and return an error if it still exists, so a silent no-op destroy can no longer be mistaken for a successful delete.

The deletion mark is now passed explicitly through the client factory rather than being inherited from the CR's own deletion timestam.

2) Idempotent Create. For rotated bindings, the binding name is committed to an annotation before any external call, and reused on retry instead of regenerated. Create then does a lookup-before-create:
  if an owned binding already exists under that name (a prior attempt that succeeded but lost its result), it is adopted and its GUID recorded instead of creating a duplicate. The pending-name marker is
  cleared atomically with persisting the external-name. Non-rotated bindings continue to use their stable spec name with no extra bookkeeping.
this basically solves this distributed transaction problem Between external system create and writing into CR status.

3) Observable deletion failures. Retired keys now track DeletionAttempts and LastDeletionError. A non-zero attempt count means the binding is overdue and repeatedly failing to delete, giving operators something concrete to alert on.

4) The recovery lookup logic was also refactored into a shared lookupOwnedBinding helper used by both the heal path and Create, so ownership checks behave consistently.

## Cold Workspace Delete Problem Reference
https://github.com/crossplane/upjet/blob/main/pkg/terraform/files.go#L251
```
// EnsureTFState writes the Terraform state that should exist in the filesystem
// to start any Terraform operation.
func (fp *FileProducer) EnsureTFState(_ context.Context, tfID string) error { //nolint:gocyclo // easier to follow as a unit
	// ...turn errors.Wrap(err, errCheckIfStateEmpty)
	}
	// We don't fill up the TF state during deletion because Terraform's removal
	// of them from the TF state file signals that the deletion was successful.
	// This is especially useful for resources whose deletion are scheduled for
	// a long period of time, where if we fill the ID, the queries would actually
	// succeed, i.e. GCP KMS KeyRing.
	if !empty || meta.WasDeleted(fp.Resource) {
		return nil
	}
```
If the state file is empty here and meta.WasDeleted would succeed, it would not write the Terraform state.
When upjet runs the destroy cli command, it pipes the error through to crossplane, where error == nil means deletion was successful.

Terraform cli Destroy command https://github.com/crossplane/upjet/blob/main/pkg/terraform/workspace.go#L248
Piping: https://github.com/crossplane/upjet/blob/main/pkg/controller/external.go#L476


enrico-kaack-comp:
IMPORTANT: run long e2e before merging since this contain rotation e2e tests
Run successful: https://github.com/SAP/crossplane-provider-btp/actions/runs/32734998120

internal/controller/account/servicebinding/servicebinding.go
```diff
@@ -391,21 +512,60 @@ func (e *external) emit(cr resource.Managed, ev event.Event) {
...
+	if _, err := e.clientFactory.CreateClient(ctx, cr, targetName, targetExternalName, false); err != nil {
+		return errors.Wrap(err, errSeedBinding)
+	}
 
-	// Create a client for the specific binding to delete
-	client, err := e.clientFactory.CreateClient(ctx, cr, targetName, targetExternalName)
+	// Phase 2: connect with the deletion mark set so prevent_destroy is off, then destroy.
+	client, err := e.clientFactory.CreateClient(ctx, cr, targetName, targetExternalName, true)
 	if err != nil {
 		return err
```
jn-av: since we wrapped the other errors also with more context it would be also nice here so it's easier to debug and understand the error

apis/account/v1alpha1/servicebinding_types.go
```diff
@@ -124,6 +124,20 @@ type RetiredSBResource struct {
...
 	// May change if the rotation settings change
 	DeletionDate *metav1.Time `json:"deletionDate"`
+
+	// DeletionAttempts counts how many times deletion of this retired binding
+	// has been attempted and did not verifiably remove it from Service Manager.
+	// It stays at 0 while deletion is not yet due; a non-zero value means the
+	// binding is overdue and repeatedly failing to delete, so it is safe to
+	// alert on.
+	// +kubebuilder:validation:Optional
+	DeletionAttempts int32 `json:"deletionAttempts,omitempty"`
```
jn-av, line 134: I like that. Makes it easy to find this security relevant issue

jn-av (commented):
Tested the workspace lost during rotation case on my machine, it worked. Implementation wise it also looks good to me, just one nice to have error wrapper. Thanks!

xdj2e9 (commented):
Thanks a lot for this focussed fix of our custom rotation logic. Looks good to me. Feel free to merge when @jn-av comments are addressed.

xdj2e9 (approved):

SatabdiG (approved):

PR #937, opened 2026-08-24
docs: add v1-to-v2 migration guide

### LINKED ISSUE #676: v2 Migration guide

xdj2e9:
Closes #676. Refs #845.

## Summary

Adds `docs/end-user-guides/migration/v1-to-v2.mdx` migration guide for upgrading from v1.x to v2.0.0. v2.0.0 lifts the provider to the Crossplane v2 runtime and removes legacy concepts, so users need pre-upgrade steps and check commands for each breaking change. The guide documents only the btp deltas and links upstream for general Crossplane v1→v2 core migration.

### Crossplane core compatibility
v2.0.0 runs on both Crossplane core v1 and v2 although the provider updated to crossplane runtime v2.

### External Secret Store (ESS) removal
`publishConnectionDetailsTo` stops publishing the moment v2.0.0 installs, silently, on any Crossplane core version. Replacement: `writeConnectionSecretToRef` or External Secrets Operator.

### `GlobalAccount` CRD removal
Describes all steps necessary to handle the removed `GlobalAccount` CRD (and all its implementation):
- Set `ProviderConfig.spec.globalAccount` (should be set anyway according to setup guide)
- drop `globalAccountRef`/`globalAccountSelector` (only cosmetic, was not read in v1.13.0 anyway)
- delete CRs pre-upgrade, the MRD (crossplane v2 only), the orphaned CRD post-upgrade (CRs before CRD, MRD before CRD (v2 only))

### New ServiceManager planName default
Default `spec.forProvider.planName` changes from `service-operator-access` to `subaccount-admin` (#845 / PR #925, on `main`). Existing resources are unaffected (create-time default, immutable); only a fresh create or GitOps recreate that omits the field adopts the new default. Added advice to pin `service-operator-access` if required.

## To do after merge

- [x] Merge PR to bump BTP submodule pointer in https://github.com/SAP/crossplane-provider-docs


docs/end-user-guides/migration/v1-to-v2.mdx
```diff
@@ -0,0 +1,112 @@
+---
+sidebar_position: 0
+---
+# Migrate from v1.x to v2.0.0
+
+v2.0.0 lifts the provider to the Crossplane v2 runtime and removes two legacy concepts. Managed resources stay cluster-scoped, so most manifests are unchanged.
```
enrico-kaack-comp: Link to which resources are changed: like see below

enrico-kaack-comp (dismissed):
Really good, well explained what to do!

docs/end-user-guides/migration/v1-to-v2.mdx
xdj2e9: Added

docs/end-user-guides/migration/v1-to-v2.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+**No.** v2.0.0 runs on both Crossplane core **v1 and v2** control planes. The provider's runtime bump to v2 is independent of the Crossplane core version - you can adopt v2.0.0 without upgrading core, and upgrade core on its own schedule.
+
+## Before you upgrade
+
+Do these while the **v1.x provider is still running**.
+
+### 1. Move off External Secret Stores (ESS)
+
+v2.0.0 removes the ESS code path. If any resource publishes connection details via `publishConnectionDetailsTo`, publishing stops the moment you install v2.0.0 - on any core version, with no error. Check whether you use it:
```
sk31337:
```suggestion
[v2.0.0 removes](https://docs.crossplane.io/latest/guides/upgrade-to-crossplane-v2/#removed-features) the ESS code path. If any resource publishes connection details via `publishConnectionDetailsTo`, publishing stops the moment you install v2.0.0 - on any core version, with no error. Check whether you use it:
```

docs/end-user-guides/migration/v1-to-v2.mdx
```diff
@@ -0,0 +1,112 @@
...
+kubectl get storeconfigs.secrets.crossplane.io -A
+kubectl get storeconfigs.btp.sap.crossplane.io -A
+kubectl get managed -A -o yaml | grep -i publishConnectionDetailsTo
+```
+
+The provider also removes its own `StoreConfig` CRD (`storeconfigs.btp.sap.crossplane.io`). Any `StoreConfig` you created for the provider is orphaned after the upgrade - see step 3 and "After you upgrade" for cleanup.
+
+Any hits: switch those resources to a native connection secret via `spec.writeConnectionSecretToRef`, or use the [External Secrets Operator](https://external-secrets.io/), before upgrading. Upstream detail: [Crossplane upgrade-to-v2, External Secret Stores](https://docs.crossplane.io/latest/guides/upgrade-to-crossplane-v2/#external-secret-stores).
+
+### 2. Set the global account in `ProviderConfig`, drop `GlobalAccount` refs
```
sk31337: (if possible and sensible) @xdj2e9 may you provide a warning/error message, that will be displayed if this change is not done?

This might help consumers to find this migration guide much faster, when searching for the warning/error message.

docs/end-user-guides/migration/v1-to-v2.mdx
```diff
@@ -0,0 +1,112 @@
...
+While the v1.x controller still runs, so it clears finalizers cleanly, delete the `GlobalAccount` CRs and any provider `StoreConfig`:
+
+```sh
+kubectl delete globalaccounts.account.btp.sap.crossplane.io --all
+kubectl delete storeconfigs.btp.sap.crossplane.io --all
+```
+
+### 4. Pin `planName` on ServiceManager if you rely on the default
+
+v2.0.0 changes the default `ServiceManager` `spec.forProvider.planName` from `service-operator-access` to `subaccount-admin`. `service-operator-access` is for the [SAP BTP Service Operator](https://github.com/SAP/sap-btp-service-operator/blob/main/README.md); `subaccount-admin` is the general-purpose plan for direct API access.
```
sk31337, line 141: (if possible and sensible) @xdj2e9 may you provide a warning/error message, that will be displayed if this change is not done?

This might help consumers to find this migration guide much faster, when searching for the warning/error message.

sk31337 (changes requested):

docs/end-user-guides/migration/v1-to-v2.mdx
xdj2e9: v2.0.0 is referencing our upcoming btp provider version, so linking crossplane v2 would be misleading

docs/end-user-guides/migration/v1-to-v2.mdx
xdj2e9: Nice idea. Unfortunately I'm not aware of any hook in the provider that could detect the "change is not done" situation and then display the warning/error message. So the only option would be to always write it which would feel too noisy to me. WDYT?

docs/end-user-guides/migration/v1-to-v2.mdx
xdj2e9: The `planName` is set as default by k8s based on the CRD and not by the provider. So the provider cannot distinguish if it is an explitly set `planName` or a defaulted one from the CRD. So in the end we end up in the same situation as https://github.com/SAP/crossplane-provider-btp/pull/937#discussion_r3851609189.

enrico-kaack-comp (dismissed):

docs/end-user-guides/migration/v1-to-v2.mdx
sk31337: Than pls add "btp provider...." so its not misleading :)

docs/end-user-guides/migration/v1-to-v2.mdx
sk31337: What would be the errors/warnings, when upgrading to btp provider 2.0.0 but not doing the migrations?

docs/end-user-guides/migration/v1-to-v2.mdx
sk31337: Resolve this conversation

docs/end-user-guides/migration/v1-to-v2.mdx
xdj2e9: Do you think it is misleading? This is the visual context when reading the guide:
image

My gut feeling is that replacing `v2.0.0` with `BTP Provider v2.0.0` in the guide would add too much repitition

docs/end-user-guides/migration/v1-to-v2.mdx
xdj2e9: Actually nothing breaks when these things are not done:
- The globalAccount property in the provider config is required already if upjetted resources are used. So most user already must have it (as it is also part of the [configuration guide](https://sap.github.io/crossplane-provider-docs/docs/crossplane-provider-btp/docs/end-user-guides/setup/configure-provider-btp#create-a-providerconfig)). Users who don't have it yet are also not broken (because the resources they use don't need it)
- The global account refs (selector / ref) are not read by any code path and would keep sitting there

The `If you skip these steps` section gives an overview of the consequences when not doing the migration.

enrico-kaack-comp (approved):

docs/end-user-guides/migration/v1-to-v2.mdx
sk31337: Thanks for the detailed explanation. Pls resolve the conversation :)

docs/end-user-guides/migration/v1-to-v2.mdx
sk31337: LGTM :) Thanks for the explanation.

sk31337 (approved):

Issue #282, opened 2025-08-22
[FEATURE] Support upcoming `Destinations` TF Resource


GenosseOtt:
**Context**
With an  upcoming Terraform version, we will have an easy way to also support`Destionations` via our Provider 

This issue depends on: https://github.com/SAP/terraform-provider-btp/issues/931

## Acceptance Criteria
- [ ] Support for Destinations service in Crossplane
- [ ] Adjust end user material

enrico-kaack-comp:
Requires 
- [x] #473

stefan-reichenbach:
@sdischer-sap Any chance to get this delivered soon?

mpechkurov:
Check-in 8.07 : 
- Update is done only resource work is pending. 
   - need to be clarified with @dwlou if we can to add another resource since we agreed not work with TF in the long term.

mpechkurov:
We agree in the team to use implementation with upjet.

mpechkurov:
check-in 19.08.2026 : 
- no blockers 
- we can check if we can finish this in next check in meeting

mpechkurov:
**Context:** This was discovered during the implementation of #903 (SubaccountDestination upjet-based resource).

## Root Cause: Why `SubaccountDestination` Is Blocked Until No-Fork

### Short version

The BTP Terraform provider has a bug ([SAP/terraform-provider-btp#1532](https://github.com/SAP/terraform-provider-btp/issues/1532)). When it reads a destination that doesn't exist yet, it crashes instead of returning "not found." The current architecture can't recover from that crash. The no-fork architecture — already written but not yet merged — handles it correctly.

### What happens today

When Crossplane wants to create a destination, it first runs `terraform refresh` to check if it exists. For a new resource, that should return "not found" and trigger Create.

Instead, the BTP Terraform provider crashes with:
> `Missing Resource Identity After Read`

This happens because the provider correctly says "this resource is gone" on a 404 — but forgets to also set the required identity fingerprint. The Terraform framework treats this as fatal and exits with a non-zero code.

Our architecture runs Terraform as a **subprocess** and treats any non-zero exit as a fatal error — so the reconciler retries forever and **never reaches Create**.

We tried patching the tfstate file with the identity data before terraform runs — but the validation fires inside the provider binary, not in the file on disk. There is no external workaround.

### Why no-fork fixes it

[...]

PR #939, opened 2026-08-25
feat: native Go implementation of SubaccountDestination

### LINKED ISSUE #903: Implement SubaccountDestination Crossplane resource
## Context
Part of epic #282 — Support Destinations TF Resource.

Implement the `SubaccountDestination` Crossplane managed resource as a **native Go controller** (not upjet-based). The Destination Service has its own REST API, so a native implementation avoids the overhead of Terraform provider wrapping and gives full control over credential handling and reconciliation logic.

## Approach

Instead of registering `btp_subaccount_destination` in the upjet config, this is implemented as a hand-written managed resource controller:

- **CRD types** defined manually in `apis/account/v1alpha1/subaccountdestination_types.go`
- **Destination Service client** in `internal/clients/account/destination/` — thin HTTP client with OAuth2 client credentials auth
- **Controller** in `internal/controller/account/subaccountdestination/` — implements `Connect`, `Observe`, `Create`, `Update`, `Delete`
- **ProviderConfig** extended with `destinationCredentials` field pointing at a ServiceBinding connection secret
- **External name** format: `<subaccount-id>/<destination-name>`

## Acceptance Criteria
- [x] CRD types with `subaccountRef` cross-resource reference
- [x] Native Go controller registered in provider setup
- [x] Destination Service client supporting GET / POST / PUT / DELETE with ETag-based optimistic concurrency
- [x] `destinationCredentials` in ProviderConfig — supports ServiceBinding `secretKey: credentials` format with normalization for different BTP landscape credential structures
- [x] Build and lint pass
- [x] E2e test for full CRUD flow (self-contained: provisions subaccount, ServiceManager, ServiceInstance, ServiceBinding, then creates/updates/deletes a destination)

mpechkurov:
## Summary

- Implements \`SubaccountDestination\` as a native Go Crossplane managed resource backed by the SAP Destination Service REST API, replacing the blocked upjet-based approach (see #903, EPIC #282)
- The upjet implementation was blocked by a bug in the BTP Terraform provider ([SAP/terraform-provider-btp#1532](https://github.com/SAP/terraform-provider-btp/issues/1532)) where \`handleReadErrors\` doesn't set \`resp.Identity\` on 404 responses
- Adds a generated OpenAPI client for the Destination Service API
- Destination Service OAuth2 credentials are supplied per-resource via \`spec.forProvider.destinationServiceBindingSecretRef\` (pointing to a secret created by a \`ServiceBinding\` CR or manually) — no global \`ProviderConfig\` field needed
- Manually tested end-to-end: Create ✅ Update (drift detection + ETag) ✅ Delete ✅

## Changes

- \`internal/openapi_clients/btp-destination-service-api-go/\` — generated OpenAPI client from SAP API Hub spec
- \`apis/account/v1alpha1/subaccountdestination_types.go\` — hand-written CRD types with \`additionalProperties\` map, \`additionalConfigurationSecretRefs\`, \`destinationServiceBindingSecretRef\` (required), SubaccountRef cross-reference, external-name doc comment
- \`internal/clients/account/destination/credentials.go\` — \`LoadFromSecret\` supporting both flat-key format (produced by \`ServiceBinding\` without \`secretKey\`) and single JSON-key format (produced by \`ServiceBinding\` with \`secretKey\`)
- \`internal/clients/account/destination/destination.go\` — \`DestinationClientI\` HTTP client (Get/Create/Update/Delete, IsNotFound/IsConflict predicates)
- \`internal/controller/account/subaccountdestination/subaccountdestination.go\` — native controller with Observe/Create/Update/Delete, \`validateExternalName\`, \`buildPropertyBag\`, \`isUpToDate\`; tracks managed keys in \`status.atProvider.managedKeys\` after each write so server-injected fields are ignored during drift detection
- \`internal/controller/template.go\` — registered in \`CustomSetup\`
- \`test/e2e/subaccount_destination_test.go\` — E2E test (create/update/delete flow)

## Test plan

- [x] Unit tests pass: \`go test ./...\`
- [x] E2E test \`TestSubaccountDestination_CreationFlow\` runs against a real BTP subaccount
- [ ] Verify \`kubectl explain subaccountdestination.spec.forProvider\` shows all fields
- [ ] Manual test: create, patch URL (verify update), delete (verify 404 from Destination Service API)

## Notes

- \`serviceInstanceId\` field is reserved in types for forward compatibility but returns an error if set (service-instance-scoped destinations use a different API path — separate issue)
- If an \`additionalConfigurationSecretRefs\` secret is deleted, the CR surfaces \`Synced=False\` with a clear error message and recovers automatically once the secret is restored; secret rotation is picked up transparently on the next reconcile


apis/account/v1alpha1/subaccountdestination_types.go
```diff
@@ -0,0 +1,233 @@
...
+	// +optional
+	SubaccountRef *xpv1.Reference `json:"subaccountRef,omitempty"`
+
+	// Selector for a Subaccount to populate subaccountId.
+	// +optional
+	SubaccountSelector *xpv1.Selector `json:"subaccountSelector,omitempty"`
+
+	// Name of the destination. Immutable after creation.
+	// +optional
+	Name string `json:"name,omitempty"`
```
SatabdiG, line 33: Should we add a CEL validation rule (+kubebuilder:validation:XValidation)? I was just wondering if a user edits `spec.forProvider.name`, `buildPropertyBag` will include the new name in the PUT body while the destination lookup key (external-name) stays the old name — what would the api do in such as case? Should we put a immutability guard either way?

internal/clients/account/destination/destination.go
```diff
@@ -0,0 +1,232 @@
...
+	// UAA holds nested credentials written by some BTP landscapes.
+	UAA *struct {
+		ClientID     string `json:"clientid"`
+		ClientSecret string `json:"clientsecret"`
+		URL          string `json:"url"`
+	} `json:"uaa,omitempty"`
+}
+
+// ParseCredential unmarshals a Destination Service binding JSON blob.
+func ParseCredential(raw []byte) (DestinationCredential, error) {
```
SatabdiG, line 33: This function and `assembleDestinationCredJSON` (config.go) both independently handle tokenurl vs token_url, and url vs uri naming variants —  would a shared helper would reduce the risk of the two falling out of sync as more BTP landscape quirks surfac?

apis/account/v1alpha1/subaccountdestination_types.go
```diff
@@ -0,0 +1,233 @@
...
+	ETag *string `json:"etag,omitempty"`
+
+	// RawProperties is the full property bag as returned by the Destination Service API.
+	// Used internally to determine whether an update is needed.
+	// WARNING: For destinations using BasicAuthentication or OAuth, the API response may
+	// include sensitive fields (e.g. Password, ClientSecret). These are stored verbatim
+	// in this Kubernetes object. Restrict access to SubaccountDestination resources via
+	// RBAC to prevent unauthorized exposure of credentials.
+	// +optional
+	RawProperties map[string]string `json:"rawProperties,omitempty"`
```
SatabdiG: Since this is purely advisory,also given this is a +kubebuilder:subresource:status field visible to anyone who can get the CR's status; is it worth actually filtering known-sensitive keys out of RawProperties before storing, rather than relying entirely on RBAC discipline?

SatabdiG (commented):
🔥 loved the test coverage, 

Couldn't find any usage for the `OpenAPI client`(grepped for imports outside its own package, no hits) — maybe the other files need this?

internal/clients/account/destination/destination.go
mpechkurov: Good catch. The two functions handle different input formats (flat secret keys vs JSON blob unmarshaling) so a shared helper would mix concerns rather than reduce them. To address the sync risk you identified, added cross-reference
  comments in both places pointing to the other — if a new landscape alias surfaces, the comment makes it explicit to update both.

apis/account/v1alpha1/subaccountdestination_types.go
mpechkurov: Good point. Added active redaction of known sensitive keys (Password, ClientSecret, KeyStorePassword, TrustStorePassword) in redactSensitiveKeys before storing in RawProperties, rather than relying solely on RBAC. Updated the field
  comment to reflect this. A kubebuilder annotation wouldn't help here since status is written by the controller, not user input — so there's nothing for the admission webhook to intercept.

apis/account/v1alpha1/subaccountdestination_types.go
mpechkurov: Good catch — this is a real correctness issue. If name is mutated after creation, the PUT body would send the new name while the external-name (used for GET/DELETE lookup) still holds the old one, which would cause BTP to create a
  duplicate or error. Added +kubebuilder:validation:XValidation:rule="self == oldSelf" on the field — consistent with how other immutable fields are guarded in this repo (e.g. rolecollection, kymaenvironment). CRD regenerated.

internal/openapi_clients/btp-destination-service-api-go/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+# Remove generated files we don't need
+rm -rf pkg/test pkg/go.mod pkg/go.sum pkg/git_push.sh pkg/.travis.yml
+
+go mod tidy
+```
+
+### Apply patches
+
+Edit `swagger-patched.json` before regenerating if the raw spec needs fixes.
```
xdj2e9: The is not the flow we should use. Other generated clients use an`swagger-patch.json` alongside the original `swagger.json` according to RFC 6902 JSON Patch. Then an json patch tool like `npx json-patch-cli` can be used to yield the `swagger-patched.json` which is then used to generate the OpenAPI client.

internal/openapi_clients/btp-destination-service-api-go/pkg/client.go
```diff
@@ -0,0 +1,704 @@
...
+	TrustCertificateOnSubaccountLevelAPI TrustCertificateOnSubaccountLevelAPI
+}
+
+type service struct {
+	client *APIClient
+}
+
+// NewAPIClient creates a new API client. Requires a userAgent string describing your application.
+// optionally a custom http.Client to allow for advanced features such as caching.
+func NewAPIClient(cfg *Configuration) *APIClient {
```
xdj2e9, line 82: As mentioned by @SatabdiG in https://github.com/SAP/crossplane-provider-btp/pull/939#pullrequestreview-5040831723 this client is unused and the custom HTTP calls implemented in `internal/clients/account/destination/destination.go` are used instead.

If possible we should use the generated client so that we do not have to additionally maintain the REST client.

Is there something blocking you from using the generated client?

apis/account/v1alpha1/subaccountdestination_types.go
```diff
@@ -0,0 +1,233 @@
...
+	// +optional
+	SubaccountSelector *xpv1.Selector `json:"subaccountSelector,omitempty"`
+
+	// Name of the destination. Immutable after creation.
+	// +optional
+	Name string `json:"name,omitempty"`
+
+	// Type of the destination. e.g. HTTP, LDAP, RFC, MAIL.
+	// +optional
+	Type string `json:"type,omitempty"`
```
xdj2e9, line 38: The OpenAPI specification intentionally underspecifies what exact properties can be send (https://api.sap.com/api/SAP_CP_CF_Connectivity_Destination/path/post_v1_subaccountDestinations):
```
 {
  "required": ["Name", "Type"],
  "type": "object",
  "properties": {
    "Name":         { "type": "string" },
    "Type":         { "type": "string", "enum": ["HTTP","RFC","MAIL","LDAP"] },
    "PropertyName": { "type": "string" }
  }
}
```

I think we should handle it the same in our provider and only define `Name` and `Type` and then offer the general purpose `AdditionalProperties` map. That way we match the actual API and the source of truth stays in their spec: https://help.sap.com/docs/connectivity/sap-btp-connectivity-cf/destination-service-rest-api-23ccafbea18f4b65919a2799f2cd20e6-150?locale=en-US

Otherwise we always need to keep the properties we specify ourselve (here `URL`, `Authentication`, `ProxyType`, `Description`) in sync with their spec.

apis/account/v1alpha1/subaccountdestination_types.go
```diff
@@ -0,0 +1,233 @@
...
+	ProxyType *string `json:"proxyType,omitempty"`
+
+	// Description of the destination.
+	// +optional
+	Description *string `json:"description,omitempty"`
+
+	// ServiceInstanceID scopes this destination to a service instance.
+	// Not yet implemented — setting this field returns an error.
+	// +optional
+	ServiceInstanceID *string `json:"serviceInstanceId,omitempty"`
```
xdj2e9: If that field is not yet implemented and optional: why not omit it completly until we implement it?

apis/account/v1alpha1/subaccountdestination_types.go
xdj2e9: I see a risk here: if the destination service adds new secrets we will leak them until we adapted our code.

Can we just drop the `RawProperties` completly from storage? I think they are just used in `Observe` for the `upToDate` check. But at that place they are available as a var already. So no need to store them.

apis/account/v1alpha1/subaccountdestination_types.go
```diff
@@ -0,0 +1,233 @@
...
+	// +optional
+	SubaccountRef *xpv1.Reference `json:"subaccountRef,omitempty"`
+
+	// Selector for a Subaccount to populate subaccountId.
+	// +optional
+	SubaccountSelector *xpv1.Selector `json:"subaccountSelector,omitempty"`
+
+	// Type of the destination. e.g. HTTP, LDAP, RFC, MAIL.
+	// +kubebuilder:validation:Required
+	Type string `json:"type"`
```
xdj2e9, line 87: As the OpenAPI spec gives us a fixed range of possible values (`"Type": { "type": "string", "enum": ["HTTP","RFC","MAIL","LDAP"] }`), we should add an enum validation.

apis/account/v1alpha1/subaccountdestination_types.go
xdj2e9: Given the API spec it should also not be optional I think. The same for `Type`.

apis/account/v1alpha1/subaccountdestination_types.go
```diff
@@ -0,0 +1,233 @@
...
+	// +optional
+	ServiceInstanceID *string `json:"serviceInstanceId,omitempty"`
+
+	// AdditionalProperties merged into the destination property bag.
+	// +optional
+	AdditionalProperties map[string]string `json:"additionalProperties,omitempty"`
+
+	// AdditionalConfigurationSecretRef for sensitive destination properties.
+	// +optional
+	AdditionalConfigurationSecretRef *xpv1.SecretKeySelector `json:"additionalConfigurationSecretRef,omitempty"`
```
xdj2e9: Let's align this with how `ServiceInstance` and `ServiceBinding` are doing it so that a user can pull properties from more than one secret:
- make it a list: `AdditionalConfigurationSecretRef *xpv1.SecretKeySelector ->
   []xpv1.SecretKeySelector`

It may also be possible to reuse the existing `lookupSecrets`, `mergeJsonData` methods from `internal/clients/account/serviceinstance/serviceinstance.go` (they need to be moved to some common place though).

internal/openapi_clients/btp-destination-service-api-go/README.md
mpechkurov: Good catch. The swagger-patch.json was missing because the raw spec didn't require any modifications — swagger.json and swagger-patched.json were identical. We've now added an empty swagger-patch.json (valid RFC  6902, no ops) to align with the convention used by other clients, and updated the README to use the standard wording.

internal/openapi_clients/btp-destination-service-api-go/pkg/client.go
mpechkurov: Already addressed in response to @SatabdiG's comment.

apis/account/v1alpha1/subaccountdestination_types.go
mpechkurov: I'd like to keep the explicit fields. Even though the API spec is open-ended, fields like URL, Authentication and ProxyType are needed for almost every real destination. Having them typed makes the CRD
  self-documenting and easier to use — users don't have to guess what keys to put in additionalProperties. The map is still there for anything else.

xdj2e9: But the API is documented by the destination service team already. Like this we are:
- duplicating the documentation making it necessary to keep it up-to-date
- signaling to users that only these fields are supported although this limitation does not exists.

Therefore I would add a link in the CRD of `additionalProperties` to the source of truth API docs of the destination team which perfectly describes what can be used.

apis/account/v1alpha1/subaccountdestination_types.go
mpechkurov: Agreed — removed serviceInstanceId from the spec entirely. It can be added back once implemented.

apis/account/v1alpha1/subaccountdestination_types.go
mpechkurov: Agreed — removed RawProperties from status entirely. It was only used in Observe for the up-to-date check where props is already available as a local variable, so no functional change. The redactSensitiveKeys
  helper is gone too.

apis/account/v1alpha1/subaccountdestination_types.go
mpechkurov: Done — changed AdditionalConfigurationSecretRef *xpv1.SecretKeySelector to AdditionalConfigurationSecretRefs []xpv1.SecretKeySelector in both ForProvider and InitProvider. Moved the shared LookupSecrets helper to
  internal/utils.go so it can be reused across controllers. The buildPropertyBag function now delegates to it.

apis/account/v1alpha1/subaccountdestination_types.go
mpechkurov: Good point — added a link to the SAP Destination Service REST API docs in the additionalProperties description so users have a direct reference to the full list of supported properties.

apis/account/v1alpha1/subaccountdestination_types.go
mpechkurov: Done — added +kubebuilder:validation:Enum=HTTP;RFC;MAIL;LDAP to both ForProvider.Type and InitProvider.Type. The enum is now enforced at admission.

apis/account/v1alpha1/subaccountdestination_types.go
mpechkurov: I've dropped explicit typed fields

apis/account/v1alpha1/subaccountdestination_types.go
```diff
@@ -0,0 +1,187 @@
...
+	//
+	// Two secret formats are accepted:
+	//   - Flat keys (leave key empty): the secret has individual keys clientid,
+	//     clientsecret, tokenurl/token_url, uri/url — the format written by
+	//     SubaccountServiceBinding when secretKey is not set.
+	//   - Single JSON key (set key): the named key holds a JSON object with
+	//     clientid, clientsecret, tokenurl, uri — the format written by
+	//     SubaccountServiceBinding when secretKey is set (e.g. secretKey: credentials).
+	//
+	// +optional
```
xdj2e9: I don't see a case where this is optional. I think it should be required.

internal/controller/account/subaccountdestination/subaccountdestination.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+// isUpToDate returns true when desired and observed are in sync.
+// observed is the raw property map returned by the Destination Service GET.
+// Checks both directions:
+//   - every key in desired must exist in observed with the same value
+//   - every key in observed must exist in desired (ensures removed additionalProperties
+//     keys trigger an update that removes them from BTP via full-replace PUT)
+//
+// This is safe because the BTP Destination Service GET returns only user-set
```
xdj2e9: Is there evidence that this is true? There is a risk to end up in a reconcilation loop if the server returns server side "calculated" properties.

If we want to keep the two sided check, we should extend a E2E test to ensure that also the first "Observe" reconcilation after "Create" is still succeeding.

To be on the save side we could track our managed keys to know what to expect (and in case of key removal also remove it from the managed keys so we also catch the key removal case).

internal/controller/account/subaccountdestination/subaccountdestination.go
```diff
@@ -0,0 +1,279 @@
...
+		"Name": p.Name,
+		"Type": p.Type,
+	}
+	for k, v := range p.AdditionalProperties {
+		props[k] = v
+	}
+	if len(p.AdditionalConfigurationSecretRefs) > 0 && kube != nil {
+		extra, err := internal.LookupSecrets(ctx, kube, p.AdditionalConfigurationSecretRefs)
+		if err != nil {
+			return nil, errors.Wrap(err, errBuildProps)
```
xdj2e9, line 244: This is also called from `Observe`. Should that one throw an error if a secret is not available? I'm wondering if that would be the desired behavior e.g. when a secret is rotated.

xdj2e9 (commented):
Thanks a lot for the adaptions made. Only a few things.

apis/account/v1alpha1/subaccountdestination_types.go
mpechkurov: Done. Required.

internal/controller/account/subaccountdestination/subaccountdestination.go
mpechkurov: This is intentional. If the secret is deleted the reconcile will fail with an error: "while observing destination: cannot build destination property bag: secrets \"my-secret\" not found". The CR will show Synced=False and recover automatically once the secret is restored. If the secret is rotated, the controller picks up the new values on the next reconcile and updates BTP, no manual action needed.

internal/controller/account/subaccountdestination/subaccountdestination.go
mpechkurov: We couldn't confirm that GET never returns server-injected fields. I've updated the implementation to track the keys we write in status.atProvider.managedKeys. Observe now only compares against those keys, so server-injected fields are ignored and removals are still detected correctly.

xdj2e9 (approved):
🚀

SatabdiG (approved):
🔥

PR #946, opened 2026-08-28
chore(ci): bump btp cli to 2.116.2 to fix subaccount createdDate decode

xdj2e9:
Refs #887.

Bump `btp_cli_version` 2.90.2 → 2.116.2 across all CI workflows (five pins). The 2.90.2 CLI can't decode `list accounts/subaccount`: the API now returns `createdDate` as a string, the old CLI expects int64, and it aborts with `cannot unmarshal string into Go struct field Subaccount.value.createdDate of type int64`. This broke e2e cleanup after #887 changed it to use the CLI.

Also adapt `createdDate`/`modifiedDate` to use string instead of int64 in two structs holding the values.

The failure was invisible on the runs of #887 because the cleanup job checks out `github.base_ref` which still used the REST API. For the same reason this fix only takes effect once merged to main.


xdj2e9:
I need to force merge this, because E2E cleanup cannot get green.

PR #953, opened 2026-08-31
feat(no-fork): purely upjetted resources

### LINKED ISSUE #690: Introduce no-fork for upjetted resources
**Scope**
Adding it to **purely upjetted resources** (not using some kind of hybrid approach) should be fairly straight forward. This applies to the kinds:
- `SubaccountTrustConfiguration`
- `GlobalaccountTrustConfiguration`
- `DirectoryEntitlement`
- `SubaccountServiceBroker`
- `SubaccountApiCredential`

Refer to our previous implemention of https://github.com/SAP/crossplane-provider-btp/pull/311

**Acceptance Criteria** 
- [ ] Mentioned resources using the new no-fork controller implementation
- [ ] verify no drifts are detected during creation ([previous upjet issue](https://github.com/crossplane/upjet/issues/547))
- [ ] Provider upgrade tests cover that an upgrade from previous versions does not cause any regressions
- [ ] each controller should be handled in the separate PR 
- [ ] **Optional** measure performance for for and no-fork

xdj2e9:
Closes #690.

Moves the five purely upjetted resources (`SubaccountTrustConfiguration`, `GlobalaccountTrustConfiguration`, `DirectoryEntitlement`, `SubaccountServiceBroker`, `SubaccountApiCredential`) to no-fork upjet.

## New dependency

The framework connector needs the upstream provider as a live Go object, so `github.com/SAP/terraform-provider-btp` is now a direct `go.mod` dependency (pinned at v1.25.0 to match `TERRAFORM_PROVIDER_VERSION` in the Makefile). Adapted `renovate.config` to update all places in sync.

## Import fix

`SubaccountTrustConfiguration` import returned HTTP 400 under the framework connector. The TF provider's `Read` needs both `subaccount_id` and `origin` from state; on import, the framework connector reconstructs state from the external-name via `SetIdentifierArgumentFn`. The existing config populated `origin` but not `subaccount_id`. Fix: `SetIdentifierArgumentFn` now splits `<sa-id>/<origin>` into both fields. Create was unaffected.

## Cleanup

The `zz-inject-identity` post-generation tool had no remaining targets after the migration. Therefore the binary and its `//go:generate` hook were removed. The runtime `NewIdentityInjectingStore` wrap in `setup.go` stays as it is used by the hybrid resources.


internal/clients/tfclient/setup.go
```diff
@@ -55,6 +55,7 @@ func TerraformSetupBuilder(version, providerSource, providerVersion string) terr
 				Source:  providerSource,
 				Version: providerVersion,
 			},
+			FrameworkProvider: config.GetProvider().TerraformPluginFrameworkProvider,
```
mpechkurov: Looks like config.GetProvider() is called on every reconcile here. It parses the full schema each time. Could we capture it once outside the closure ? Same in TerraformSetupBuilderNoTracking.

xdj2e9: Done

mpechkurov (approved):
LGTM

PR #963, opened 2026-09-01
fix(serviceinstance): unblock parameter-only updates swallowed by upjet late-init

### LINKED ISSUE #962: [BUG] ServiceInstance parameters-only updates silently ignored since v1.11.0
A `ServiceInstance` (`account.btp.sap.crossplane.io/v1alpha1`) update that changes only `spec.forProvider.parameters` is silently swallowed on provider v1.11.0 and later. The MR reports `Ready=True` almost immediately, but the new parameters are never applied to the BTP service instance. On v1.10.0 the same update reconciles correctly.

## Reproduce

1. Create a `ServiceInstance` with a plan that allows in-place parameter updates (`instances_retrievable=true`, `plan_updateable=true`), e.g. one-mds/sap-integration, leaving `serviceOfferingName`/`serviceplanName` empty in the spec.
2. Wait for `Ready=True`.
3. Change only `spec.forProvider.parameters`.
4. Observe: the MR flips back to `Ready=True` within seconds with no Terraform plan/update. The service instance keeps the old parameters.

On v1.10.0 step 3 produces a real drift-detect + update. On v1.11.0 it "passes" almost immediately with no update applied.

## Root cause

Provider v1.11.0 bumped `TERRAFORM_PROVIDER_VERSION` from 1.15.1 to 1.23.1. That bump changes the chain of events on a parameters-only update:

1. In terraform-provider-btp 1.23.1, [`resource_subaccount_service_instance`](https://github.com/SAP/terraform-provider-btp/blob/v1.23.1/btp/provider/resource_subaccount_service_instance.go#L93-L109) makes `service_offering_name` and `serviceplan_name` `Optional + Computed`. In [1.15.1](https://github.com/SAP/terraform-provider-btp/blob/v1.15.1/btp/provider/resource_subaccount_service_instance.go#L81-L95) those two were not schema attributes at all, and `serviceplan_id` was `Required`. On 1.23.1 the `Read` path now populates them on every refresh via [`lookupPlanAndOfferingNames`](https://github.com/SAP/terraform-provider-btp/blob/v1.23.1/btp/provider/resource_subaccount_service_instance.go#L722), even when the user left them empty. (`serviceplan_id` also flipped `Required` -> `Optional + Computed`, and `parameters` gained `CustomType: jsontypes.NormalizedType{}`.)
2. The crossplane spec leaves `serviceOfferingName`/`serviceplanName` empty, so on each refresh the state carries non-zero values for fields the spec has as zero.
3. upjet's generic late-init (spec-field zero + state-field non-zero) assigns those state values back into the spec, so [`LateInitialize`](https://github.com/crossplane/upjet/blob/v2.4.1/pkg/controller/external.go#L292) returns `changed=true`, which sets upjet's `lateInitedParams=true`.
4. upjet's `Observe` [`case lateInitedParams`](https://github.com/crossplane/upjet/blob/v2.4.1/pkg/controller/external.go#L340) short-circuit returns `upToDate:true` before running `Plan`. A parameters-only change never reaches Plan/Update, so it is silently dropped.

On v1.10.0 (TF provider 1.15.1) those fields are absent from state so the `lateInitedParams` stays false, `Plan` runs and finally the update is applied.

Upjet's late initialization docs: https://github.com/crossplane/upjet/blob/main/docs/configuring-a-resource.md#late-initialization-configuration

## Proposed fix

Add the server-derived Computed fields to the `ServiceInstance`'s `LateInitializer.IgnoredFields` so late-init never writes them back into a spec the user intentionally left empty:

- `serviceOfferingName`
- `serviceplanName`
- `serviceplanId`
- `parameters`

Ignoring only `parameters` is insufficient - the offering/plan fields are the primary trigger.

## Affected versions

v1.11.0 and later. Not present on v1.10.0.

xdj2e9:
Closes #962.

A `ServiceInstance` update changing only `spec.forProvider.parameters` was silently swallowed on provider v1.11.0+. The MR flipped to `Ready=True` within seconds while the BTP instance kept the old parameters.

Root cause (full detail in #962): the v1.11.0 Terraform-provider bump 1.15.1 -> 1.23.1 made `service_offering_name`/`serviceplan_name`/`serviceplan_id` `Optional + Computed`, so upjet late-init rewrites the spec each reconcile, sets `lateInitedParams=true`, and [`Observe`](https://github.com/crossplane/upjet/blob/v2.4.1/pkg/controller/external.go#L340) short-circuits before `Plan`. A parameters-only change never reaches Update.

## Fix

Remove the server-derived fields from late-init via `IgnoredFields` in each resource's `AddResourceConfigurator` (`config/subaccount_service_instance/config.go`, `config/subaccount_service_binding/config.go`). After the fix `lateInitedParams` stays false for a parameters-only change and `Observe` falls through to `Plan`. `ServiceBinding` shares the hybrid controller and its `parameters` field has the same `Optional + Computed` shape.

## Scope

Limited to the fields observed to trigger the short-circuit (`service_offering_name`, `serviceplan_name`, `serviceplan_id`, `parameters`). Other `Optional+Computed` fields (`labels`, `shared`) are late-init candidates but have not been seen to misbehave; [upjet's guidance](https://github.com/crossplane/upjet/blob/main/docs/configuring-a-resource.md#late-initialization-configuration) is to add `IgnoredFields` reactively on observed behaviour, not for the whole class.

## Test

`test/e2e/serviceinstance_parameter_update_test.go` guards the regression. It runs on one-mds/sap-integration, a plan proven to support in-place parameter updates. It flips a parameter and asserts on the Service Manager backend rather than CR conditions. Asserting on `Ready=True` alone would be a false positive: reaching Ready fast without applying the change is exactly how the bug hides.

## Manual Service Instance parameters fetch

The test reads instance parameters with a raw authed GET rather than the generated `GetServiceInstanceParameters`. The Service Manager swagger types the parameters response as `additionalProperties: {type: string}` while the API returns arbitrary JSON (the spec's own example holds a number and a bool), so the generated `map[string]string` decode fails on any non-string parameter.


gergely-szabo-sap (approved):
good job! lgtm

PR #964, opened 2026-09-02
Make E2E Kind node images reproducible

gergely-szabo-sap:
### Summary

Ensure E2E tests consistently use the configured Kind node image version.

### Changes

- Export `KIND_NODE_IMAGE_TAG` from the `Makefile`.
- Configure E2E cluster creation to use `kindest/node:<tag>`.
- Add a default image tag for direct `go test` invocations outside Make.

### Why

Reproducible E2E testing requires the Kubernetes node image to be explicit and consistent across local and CI environments. This prevents tests from silently using different default images, reducing environment-specific failures and making test results easier to compare and debug.

### Scope

This change only affects Kind node image selection for E2E tests; it does not change provider behavior or production resources.


test/e2e/main_test.go
```diff
@@ -180,6 +189,16 @@ func SetupClusterWithCrossplane(namespace string) {
 	)
 }
 
+// resolveKindNodeImageTag returns the configured Kind node image tag. Keep a
+// default here for direct `go test -tags=e2e` invocations that do not go
+// through the Makefile.
```
n-losse, line 194: is that something we want to encourage? ideally i'd imagine all e2e tests using the Makefile

gergely-szabo-sap: I would not encourage it, but we cannot prevent it. This is just a precaution to make this work even in such edge cases.

n-losse: making it fail would be a way of preventing it, kind of. :D 

if we want to make it work, can we make that use the latest instead of some fixed version hidden in here?

gergely-szabo-sap: The question is: which is worse? Hiding the node version or making it irreproducible? I lean towards hiding the node version as better. But I can be convinced :)

n-losse: my argument would be that the proper, reproducible way is to use the Makefile, and if we're enabling the test run without the Makefile as a side path, then that using the latest image is less surprising to the user than it using some obscure hardcoded version

gergely-szabo-sap: I agree to disagree. In my view, an irreproducible test can be more surprising than one that uses a hardcoded version. An irreproducible test is more difficult to debug and fix. An irreproducible test can generate obscure issue reports: "It was working on my machine but not on yours". Irreproducible tests are the root of all evil, the NULL pointer of testing :)

gergely-szabo-sap: @n-losse, I added https://github.com/SAP/crossplane-provider-btp/pull/964/commits/8e766c73e98c0e07775ee11c15d85429e9a6ad94

This commit introduces a renovate rule to keep the hardcoded node versions in sync and up to date.

n-losse: i think my preferences in order would be 

- make runs without makefile fail 
- make them use latest
- make them use fixed version 

but I'll be able to live with it :D

n-losse (approved):

Issue #941, opened 2026-08-26
v1alpha1 ServiceManager without planName breaks on v1-to-v2 upgrade (regression from #925)


xdj2e9:
## Summary

A **v1alpha1** `ServiceManager` created without an explicit `planName` becomes permanently `Synced=False` after upgrading the provider from v1.x to v2, with the reconcile error `API error during service instance update - update_instance is not supported`. The resource cannot recover on its own. Only v1alpha1 is affected - a v1beta1 `ServiceManager` upgrades cleanly.

This is a regression introduced by #925 (`change default planName to subaccount-admin`), currently on `main` and not in any released version - so it blocks the v2.0.0 upgrade path but affects no shipped release.

## Affected resources

A v1alpha1 `ServiceManager` with no `planName`. The v1alpha1 CRD has no `planName` field at all. Shipped examples that author exactly this: `examples/kyma.yaml`, `examples/kyma-on-steroids.yaml`, `examples/sample/services.yaml` (first block).

A v1beta1 `ServiceManager` is **not** affected: the CRD defaults `planName` at admission and pins it immutable (`self == oldSelf`), so it persists across the upgrade and is never re-resolved.

## Root cause

Three factors, all required:

1. **v1alpha1 status drops `dataSourceLookup`.** The CRD conversion is `strategy: None` and the v1alpha1 observation has no `dataSourceLookup`. After the upgrade `status.atProvider.dataSourceLookup` is nil.
2. **`IsInitialized` re-resolves the plan when `dataSourceLookup` is nil.** With (1), the plan resolver re-runs instead of reusing the persisted plan ID.
3. **#925 flipped the default plan** `service-operator-access` -> `subaccount-admin`. On re-resolution a `planName`-less SM resolves the new default, not the plan its instance was created on.

The upgraded provider then derives a `serviceplan_id` diff and issues an in-place service-instance update. The service plan is immutable in BTP, so the update is rejected (`update_instance is not supported`), leaving the resource `Synced=False`.

Issue #811, opened 2026-07-06
Evaluate replacing hybrid ServiceInstance/ServiceBinding controllers with plain upjet (TF provider v1.22.0 adds offering/plan name support)


dwlou:
## Background

`ServiceInstance` and `ServiceBinding` are currently implemented as **hybrid controllers**: upjet-generated CRD types with hand-written native controllers. The native layer exists primarily to resolve `offeringName` + `planName` → `serviceplan_id` UUID via the Service Manager API, because the Terraform resource `btp_subaccount_service_instance` previously only accepted `serviceplan_id`.

## New information

`SAP/terraform-provider-btp` v1.22.0 adds `service_offering_name` and `plan_name` parameters directly to `btp_subaccount_service_instance` (and presumably `btp_subaccount_service_binding`).

## Evaluation needed

1. **Verify the schema change** — confirm `service_offering_name` and `plan_name` are present in `btp_subaccount_service_instance` in v1.22.0 (not a v2 variant or different resource)
2. **Assess CRD impact** — the upjet-generated `SubaccountServiceInstance` types would need to be regenerated from the v1.22.0 schema; verify the new fields appear and there are no breaking changes to existing fields
3. **User-facing API** — the current `ServiceInstance` CR exposes `offeringName` and `planName` in `spec.forProvider`; these would need to flow through to the regenerated TF resource spec instead of `serviceplanId`
4. **Removable pieces** if migration is feasible:
   - `internal/clients/servicemanager/api_plan_resolver.go` — plan ID resolver
   - `internal/controller/account/serviceinstance/initializer.go` — plan initializer
   - `spec.forProvider.serviceManagerSecret` / `serviceManagerSecretNamespace` fields — no longer needed
   - `status.atProvider.servicePlanId` — no longer needed
   - The `TfProxyConnector` / `ServiceInstanceMapper` hybrid wiring in `internal/clients/account/serviceinstance/`

## Expected outcome

If feasible: replace the hybrid controllers with plain upjet controllers (same pattern as `SubaccountTrustConfiguration`, `DirectoryEntitlement`, etc.)

sdischer-sap:
Replacing the "manual" lookup with the newly offered fields in the terraform provider is something we have planed. 
It actually allows us to entirely remove the `ServiceManager` from the picture and massively simplify the process. 
Lets discuss that in open GH discussion regarding this topic: https://github.com/SAP/crossplane-provider-btp/discussions/647

[...]

PR #972, opened 2026-09-09
fix: resolve plan by name using tf, drop own resolution using service manager API

xdj2e9:
Relates to #811. Supersedes the #941 workaround (#942).

## Why now

ServiceManager/ServiceInstance/ServiceBinding related E2E tests fail across all branches (including main). They time out when SM's `Connect` resolves the plan by minting a temporary admin binding via the accounts-service, which returns Code 20048 on a fresh subaccount. Connect fails before building the external client and writes no condition, so the SM stalls `Ready=Unknown`. Locally the connect retry succeeds after 4-6 attempts, so it seems to depend on the load.

Instead of making our custom service plan Id lookup work again, it was decided to utilize the now natively supported lookup using the TF resource.

## Implementation

`btp_subaccount_service_instance` takes `service_offering_name` + `serviceplan_name` directly since terraform-provider-btp v1.22.0 (see https://github.com/SAP/crossplane-provider-btp/discussions/647). `serviceInstanceCr` now forwards those instead of the self-resolved `serviceplan_id`.

## Removes the #941 regression at the root

#942 added two workarounds: a `resourcesUpToDate` guard reporting up-to-date on a plan-only diff, and a `setStatus` self-heal rewriting a stale plan ID. Both existed only because the SM re-resolved a plan and could land on the flipped #925 default after a v1alpha1 upgrade. No self-resolution means no wrong ID, so both are gone. `setStatus` still records the observed plan ID for `healExternalName`.

## Upgrade test

`Test_ServiceManager_PlanName_Upgrade` gated both API versions on Synced+Ready across v1.13.0 -> local, catching a plan flip BTP *rejects*. Added `assertPlanUnchanged`: stash each plan ID pre-upgrade, assert it is identical post-upgrade. The plan is immutable in BTP, so it must not change - this also catches a flip the provider *persists* without a rejected update, which the Synced gate alone misses.

### DIFF
--- a/internal/clients/cis/cis_tf_client.go
+++ b/internal/clients/cis/cis_tf_client.go
@@ -93,10 +93,13 @@ func (tfI *TfClientInitializer) serviceInstanceCr(cm *apisv1beta1.CloudManagemen
 				ManagementPolicies: []xpv1.ManagementAction{xpv1.ManagementActionAll},
 			},
 			ForProvider: apisv1alpha1.SubaccountServiceInstanceParameters{
-				Name:          getServiceInstanceName(cm),
-				ServiceplanID: &cm.Status.AtProvider.DataSourceLookup.CloudManagementPlanID,
-				SubaccountID:  internal.Ptr(cm.Spec.ForProvider.SubaccountGuid),
-				Parameters:    internal.Ptr(`{"grantType":"clientCredentials"}`),
+				Name: getServiceInstanceName(cm),
+				// Let the upstream instance resolve the plan from name;
+				// serviceplan_name and serviceplan_id are mutually exclusive.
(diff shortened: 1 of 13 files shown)


xdj2e9:
Closing as this path ended in a dead end: `ServiceInstance` offers the `DataCenter` field which is not existing in upstream's terraform `btp_subaccount_service_instance` resource [schema](https://registry.terraform.io/providers/SAP/btp/latest/docs/resources/subaccount_service_instance#schema). That's why we need to keep resolving of plan ids using an intermediate subaccount-admin `ServiceManager` instance via the AccountsAPI (at least for this path) as long as it is not possible to pass `DataCenter` into the TF resource.

PR #974, opened 2026-09-09
feat(account): native Go implementation of SubaccountDestinationCertificate

### LINKED ISSUE #904: Implement SubaccountDestinationCertificate Crossplane resource
## Context
Part of epic #282 — Support Destinations TF Resource.

Implement the `btp_subaccount_destination_certificate` TF resource as an upjet-based Crossplane managed resource. Required for mTLS-based destination authentication.

## Acceptance Criteria
- [ ] Register `btp_subaccount_destination_certificate` in upjet config with external-name and `subaccount_id` reference
- [ ] Run code generation (CRD types, controller scaffolding, CRD YAML manifest)
- [ ] Build and lint pass
- [ ] E2e test for basic CRUD flow

mpechkurov:
Closes #904.

## Summary

- Implements `SubaccountDestinationCertificate` as a native Go Crossplane managed resource backed by the SAP Destination Service REST API (part of EPIC #282)
- Follows the same pattern as `SubaccountDestination` (#939): OAuth2 credentials supplied per-resource via `spec.forProvider.destinationServiceBindingSecretRef`
- External-name format: `{subaccountId}/{certificateName}` — supports both create and import flows
- BTP requires a file extension in the certificate `name` field (e.g. `.pem`) to determine the format; `Content` is base64-encoded PEM text for `.pem` certificates
- API errors from BTP are enriched with the response body across all operations (Get/Create/Update/Delete), consistent with other clients in this repo
- Manually tested end-to-end: Create ✅ Delete ✅

## Changes

- `apis/account/v1alpha1/subaccountdestinationcertificate_types.go` — CRD types with `name`, `content`, `type` (optional), `subaccountRef`, `destinationServiceBindingSecretRef`
- `internal/clients/account/destination/certificate.go` — `CertificateClientI` interface + OAuth2 HTTP client (Get/Create/Update/Delete)
- `internal/controller/account/subaccountdestinationcertificate/subaccountdestinationcertificate.go` — native controller with Observe/Create/Update/Delete, `validateExternalName`, `buildCertificate`, `isUpToDate`
- `internal/controller/template.go` — registered in `CustomSetup`
- `package/crds/account.btp.sap.crossplane.io_subaccountdestinationcertificates.yaml` — generated CRD
- `test/e2e/subaccount_destination_certificate_test.go` — E2E test (create/delete flow)
- `test/e2e/testdata/crs/SubaccountDestinationCertificate/` — test CRs (subaccount, service manager, service instance, service binding, certificate)

## Test plan

- [x] Unit tests pass: `go test ./internal/controller/account/subaccountdestinationcertificate/...`
- [x] E2E test `TestSubaccountDestinationCertificate_CreationFlow` passes against a real BTP subaccount
- [x] `kubectl explain subaccountdestinationcertificate.spec.forProvider` shows all fields
- [x] Manual test: create certificate CR, verify `Ready=True` and `atProvider.name` set, delete and verify removal from BTP

## Notes

- The `type` field is optional — when omitted BTP infers the type from the file extension in `name`
- Certificate `name` must include a file extension recognized by BTP (`.pem`, `.p12`, `.pfx`) — the E2E test uses `.pem` with base64-encoded PEM content


internal/controller/account/subaccountdestinationcertificate/subaccountdestinationcertificate.go
```diff
@@ -0,0 +1,222 @@
...
+func buildCertificateWithName(cr *v1alpha1.SubaccountDestinationCertificate, name string) destclient.Certificate {
+	cert := destclient.NewCertificate(name, cr.Spec.ForProvider.Content)
+	if cr.Spec.ForProvider.Type != "" {
+		cert.Type = &cr.Spec.ForProvider.Type
+	}
+	return *cert
+}
+
+func isUpToDate(cr *v1alpha1.SubaccountDestinationCertificate, observed *destclient.Certificate) bool {
+	// BTP's Destination API does not return certificate content in GET responses,
```
xdj2e9: I tested directly against the API and it returns the content. Could you please recheck so that we can have real drift detection?

What I saw though is that GET does not return `Type`. So this cannot be compared. I'm thinking if `Type` should even be offered as the API: 1) seems to deduct it from content anyway and 2) it does not return it back in the GET call.

apis/account/v1alpha1/subaccountdestinationcertificate_types.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// Reference to a Subaccount to populate subaccountId.
+	// +optional
+	SubaccountRef *xpv1.Reference `json:"subaccountRef,omitempty"`
+
+	// Selector for a Subaccount to populate subaccountId.
+	// +optional
+	SubaccountSelector *xpv1.Selector `json:"subaccountSelector,omitempty"`
+
+	// Content is the base64-encoded certificate (PEM or DER).
```
xdj2e9: According to the APIs docs content is not restricted to PEM and DER. It can also contain: JKS, PFX, P12 (see https://help.sap.com/docs/connectivity/sap-btp-connectivity-cf/certificate-types?locale=en-US)

apis/account/v1alpha1/subaccountdestinationcertificate_types.go
```diff
@@ -0,0 +1,158 @@
...
+	// +kubebuilder:validation:Required
+	Content string `json:"content"`
```
xdj2e9: As content potentially contains secrets (in the JKS, PFX, P12 type cases) it should be provided using a secret ref.

test/e2e/subaccount_destination_certificate_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+
+	"github.com/sap/crossplane-provider-btp/apis"
+	"github.com/sap/crossplane-provider-btp/apis/account/v1alpha1"
+)
+
+// TestSubaccountDestinationCertificate_CreationFlow tests create, update (content change),
+// and delete of a SubaccountDestinationCertificate resource.
+func TestSubaccountDestinationCertificate_CreationFlow(t *testing.T) {
```
xdj2e9, line 24: As drift dection on `Content` should be possible this E2E test can be extended to ensure an updated `forProvider.Content` will trigger reconilation.

apis/account/v1alpha1/subaccountdestinationcertificate_types.go
```diff
@@ -0,0 +1,158 @@
...
+	SubaccountRef *xpv1.Reference `json:"subaccountRef,omitempty"`
+
+	// Selector for a Subaccount to populate subaccountId.
+	// +optional
+	SubaccountSelector *xpv1.Selector `json:"subaccountSelector,omitempty"`
+
+	// Name of the certificate. Immutable after creation.
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="name can't be updated once set"
+	Name string `json:"name,omitempty"`
```
SatabdiG, line 31: Could it be that the immutability guard on `Name` is on the wrong place?

`InitParameters` is the resolver-mirror struct — the field users actually set is `SubaccountDestinationCertificateParameters.Name` (`forProvider.name`, required), and that one has no `XValidation` rule. As written, this looks like it'd silently no-op rather than reject a `Name` change on the live resource. Given the doc comment above it says "Immutable after creation," might be worth double-checking this was intentional?

internal/controller/account/subaccountdestinationcertificate/subaccountdestinationcertificate.go
mpechkurov: Confirmed via live API — GET returns Content, fixed drift detection. Type is not returned by GET so removed it from the check to avoid an infinite update loop; it remains a write-only input field.

apis/account/v1alpha1/subaccountdestinationcertificate_types.go
mpechkurov: Good catch — we only tested PEM/DER and missed the full list. Updated the comment to cover PEM, CER, CRT, JKS, PFX, P12 based on the BTP docs. Also noted the FIPS restriction for key stores.

test/e2e/subaccount_destination_certificate_test.go
mpechkurov: Added an assess step that switches contentSecretRef.key to a second valid PEM cert in the same secret, triggering a real drift detection cycle and verifying the resource returns Available.

apis/account/v1alpha1/subaccountdestinationcertificate_types.go
mpechkurov: Replaced content with contentSecretRef (*xpv1.SecretKeySelector), resolved in Connect, following the existing PasswordSecretRef pattern.

apis/account/v1alpha1/subaccountdestinationcertificate_types.go
mpechkurov: Moved XValidation from initProvider.name to forProvider.name where it's actually enforced.

test/e2e/subaccount_destination_certificate_test.go
```diff
@@ -0,0 +1,102 @@
...
+					},
+					Key: "content-v2",
+				}
+
+				if err := cfg.Client().Resources().Update(ctx, updated); err != nil {
+					t.Fatalf("failed to update contentSecretRef key: %v", err)
+				}
+
+				// Confirm the resource reconciles and comes back Available after the update.
+				waitForResource(updated, cfg, t, wait.WithTimeout(5*time.Minute))
```
xdj2e9: This only asserts that the resource has `Available == true`. But updating the secret ref will not make it unavailable meaning the assertion immediately passes.

Ideally we ensure that the update reached the BTP backend: either we can do a manual GET call here to compare the cert content stored in the backend to the one we updated OR we write an hash of the content into the status which then can be used to compare the spec with the API response.

mpechkurov: Thanks. 
I've implemented hash content option. And update the test.

xdj2e9 (approved):

PR #975, opened 2026-09-09
fix(servicebinding): patch metadata on create to avoid rotation deadlock

schlingemann:
## What

In `ServiceBinding`'s `Create()`, the persist of `external-name` + removal of the force-rotation annotation was a bare `e.kube.Update(ctx, cr)`. This changes it to a non-optimistic `MergeFrom` patch.

## Why

The external binding is already created in BTP by the time we persist metadata. The `Update` carries the CR's `resourceVersion` and fails with a 409 whenever the CR was modified concurrently — which is common under rotation, since rotation mutates the CR frequently. When that write fails, `Create()` returns an error, so crossplane-runtime never stamps `external-create-succeeded`. `external-create-pending` stays newest and the resource deadlocks with:

> cannot determine creation result - remove the crossplane.io/external-create-pending annotation if it is safe to proceed

A `MergeFrom` patch:
- **Cannot 409** — it's a server-side blind merge with no optimistic lock, eliminating the conflict class rather than retrying through it.
- **Correctly removes the annotation** — diffing from a `base` snapshot serializes the deletion as an explicit JSON `null`; a marshalled `Update` would omit the key, which a merge reads as "leave untouched".
- **Only touches the two-field diff**, so it won't clobber fields a concurrent writer changed.

## Testing

- Updated `SuccessWithoutRotation` to mock `Patch` instead of `Update`.
- Added `SuccessRemovesForceRotationAnnotation`: inspects the patch bytes and asserts the force-rotation annotation is deleted via `":null"`, and that `external-name` is set. This guards against a revert to `Update` (which wouldn't emit the null).
- `go vet` clean; all `TestCreate` cases pass.

## Notes for reviewers

- Recovery on a stuck resource in the field: `kubectl annotate servicebinding <name> crossplane.io/external-create-pending-` (do **not** delete/recreate the external binding — it's healthy).

### DIFF
--- a/internal/controller/account/servicebinding/servicebinding.go
+++ b/internal/controller/account/servicebinding/servicebinding.go
@@ -243,11 +243,13 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
 		return managed.ExternalCreation{}, errors.Wrap(err, errCreateBinding)
 	}
 
+	// Persist via a MergeFrom patch rather than Update: the patch can't 409 on a
+	// stale resourceVersion, which would otherwise fail Create() and deadlock the
+	// resource with "cannot determine creation result".
+	base := cr.DeepCopy()
 	meta.SetExternalName(cr, externalName)
 	meta.RemoveAnnotations(cr, servicebindingclient.ForceRotationKey)
-
-	// Call the kube client to update the external-name and force-rotation annotations
-	if err := e.kube.Update(ctx, cr); err != nil {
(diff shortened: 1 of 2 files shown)


dattito (dismissed):

xdj2e9:
@schlingemann  @dattito 
Before merging this, could you please check if this is a different fix compared to what PR #936 is already fixing?

dattito:
Hey @xdj2e9, this PR still looks valid and up-to-date :)

xdj2e9 (approved):

Issue #691, opened 2026-06-04
Introduce no-fork controller for `SubaccountServiceInstance`


sdischer-sap:
## Context
`SubaccountServiceInstance` this is upjet resource ([old comment] is one of our hybrid resources )
It means that instead of exposing it as a resource for the user we have a native resource that wraps that terraform resource and delegates the usual CRUD method calls towards an in-memory controller instance. 

This approach requires some special handling since parts of the reconciliation logic need to be integrated and reacted on. Moving towards no-fork here might be a bit more complex compared to regular upjet resources.

The relevant resources are using our **"hybrid approach"**. This docs might help to get started with those:
https://github.com/SAP/crossplane-provider-btp/blob/main/docs/contribution-notes/resource-implementation-approaches.md

## POC
You can relate to this **WIP POC PR**: https://github.com/SAP/crossplane-provider-btp/pull/680.
It already contains a mostly working version of ServiceInstance. Requires more testing and potentially cleanup though. 

## Technical Steps
- migrate `SubaccountServiceinstance` to no-fork implementation
- adapt `ServiceInstance` to potential changes of this controller implementation
- make sure other resources that use it stay compatible  (`ServiceManager` `CloudManagement`)
- verify migration between previous existing resources towards this new no-fork based reconcilation stays compatible (using upgrade tests)

mpechkurov:
Link to the upjet resource docs -> https://github.com/SAP/crossplane-provider-btp/blob/main/docs/contribution-notes/resource-implementation-approaches.md

PR #1000, opened 2026-09-15
feat: no-fork `ServiceBinding`

### LINKED ISSUE #692: Introduce no-fork controller for `SubaccountServiceBinding`
## Context
This is practically the same as #691, but for `SubaccountServiceBinding`. 
There is some more complexity on top related to the rotation feature, which won't work out of the box with the changed controller implementation.

The relevant resources are using our "hybrid approach". This docs might help to get started with those:
https://github.com/SAP/crossplane-provider-btp/blob/docs/implementation_approaches/docs/contribution-notes/resource-implementation-approaches.md

## POC
You can relate to this **WIP POC PR**: https://github.com/SAP/crossplane-provider-btp/pull/680.
It already contains a mostly working version of ServiceBinding. Requires more testing and potentially cleanup though. 
**Also rotation is still broken in that implementation.**

## Technical Steps
- migrate `SubaccountServiceBinding` to no-fork implementation
- adapt `ServiceBinding` to potential changes of this controller implementation
- make sure other resources that use it stay compatible (ServiceManager CloudManagement)
- refactor rotation feature to be compatible with new controller implementation (refer to [documentation](https://github.com/SAP/crossplane-provider-btp/blob/main/docs/contribution-notes/service-binding-rotation.md))
- verify migration between previous existing resources towards this new no-fork based reconcilation stays compatible (using upgrade tests)
- run long e2e tests to verify rotation feature capabilities

xdj2e9:
Closes #692.

Switch `SubaccountServiceBinding` to no-fork. With #1002 (`ServiceInstance`, #691) already merged, `ServiceBinding` was the last resource on the fork runtime, so this PR also removes that runtime from the code, the image, and the build.

## No-fork ServiceBinding

External-name handling moves to the upjet helper `FrameworkResourceWithComputedIdentifier("id", "NOT_EMPTY_GUID")`, replacing the hand-rolled `GetIDFn` closure. It keeps `LateInitializer.IgnoredFields = ["parameters"]` from #962.

The pre-create Observe now lives inside `ServiceBindingClient.Create`, so the client that plans is the client that creates. `resolveCreateName` still does lookup-before-create, so SB is not exposed to the duplicate-create that made #1002 add `recoverExternalNameFromTfState`.

## Remove the fork path and runtime

No upjet fork based resources are left, so the fork connector, the identity injector, and gluing code is removed.

No TF binary build into the image anymore. So everything related is removed.


cluster/images/provider-btp/Dockerfile
```diff
@@ -1,65 +1,16 @@
 FROM alpine@sha256:28bd5fe8b56d1bd048e5babf5b10710ebe0bae67db86916198a6eec434943f8b
-RUN apk --no-cache add ca-certificates bash
+RUN apk --no-cache add ca-certificates
 
 ARG TARGETOS
 ARG TARGETARCH
+ARG COMMIT_HASH
+ARG VERSION
+ARG BASE_NAME
```
mpechkurov, line 9: ARG BASE_NAME is declared but never passed via --build-arg in the Makefile — so org.opencontainers.image.base.name will always be empty. Is this intentional?

mpechkurov:
Looks nice. 
It's amazing cleanup. 👍

mpechkurov (dismissed):

cluster/images/provider-btp/Dockerfile
xdj2e9: Good catch. I fixed it.

mpechkurov (approved):

PR #987, opened 2026-09-14
feat(servicebinding): complete external-name ADR compliance

xdj2e9:
Closes #440.

## Goal

Complete `ServiceBinding` external-name ADR compliance. This adds format validation, a clear message and a documented adoption path.

## Validation

`Observe` now validates the external-name before it reaches the upjet client. A non-UUID value is rejected. The message tells the user to set the ServiceBinding GUID to adopt an existing binding, or drop the annotation to create a new one. Previously a bad value flowed through as the TF `id` and showed as an unclear Terraform error. Empty and `external-name == cr.Name` are skipped as they are covered by the recovery and heal paths already existing.

## Documentation

Added a doc comment on the `ServiceBinding` struct which generates the entry in `external-name.md`. As the BTP cockpit shows only the binding name, users are asked to use btp cli instead.

## Tests

Unit tests on `Observe` cover the invalid-GUID rejection. They also cover pass-through for empty and valid-GUID external-names.

An E2E test covering the import scenario was added (`servicebinding_import_test.go`).

The added upgrade test (`servicebinding_external_name_upgrade_test.go`) confirms a binding from an older release keeps its valid-GUID external-name after upgrade.

PR #1002, opened 2026-09-15
feat(no-fork): migrate ServiceInstance to the plugin-framework client

### LINKED ISSUE #691: Introduce no-fork controller for `SubaccountServiceInstance`

Kukkerem:
Continues #953 with the first hybrid resource. `btp_subaccount_service_instance`
moves from the forked CLI client to the in-process Terraform Plugin Framework
client. `ServiceInstance` is a native CR delegating to an embedded upjet
controller, so the switch touches the shared `tfclient` wiring as well as the
resource config. `SubaccountServiceBinding` stays on the CLI client until #692.
The external-name format is unchanged (ServiceInstance GUID), so there is
nothing to migrate.

## Changes

The resource moves from `CLIReconciledExternalNameConfigs` to
`TerraformPluginFrameworkReconciledExternalNameConfigs`, and
`NewInternalTfConnector` now derives the client kind from include-list
membership rather than a call-site flag, so no call site can disagree with the
map. The hand-rolled `notEmptyGUID` `GetIDFn` gives way to upjet's
`FrameworkResourceWithComputedIdentifier("id", "NOT_EMPTY_GUID")`, which returns
a fresh `ExternalName`, so the customised fields are re-applied after the
assignment.

`QueryAsyncData` gated observation write-back on `TypeAsyncOperation`. The
framework async client marks a settled resource with `LastAsyncOperation/True`
and never sets that condition, so the old gate would have returned `nil` forever
and silently stopped external-name and status propagation.

`recoverExternalNameFromTfState` persists an identity upjet learns without the
async completion gate ever opening: a create that fails after BTP already created
the instance leaves the GUID in the partial Terraform state, which upjet stamps
onto the mapped resource while `QueryAsyncData` still reports nothing. Without
it the identity of a live instance is discarded on every reconcile and the
not-existing leg can create a duplicate. Only a fallback external-name is
overwritten, and only by a UUID.

`Observe` rejects a non-UUID external-name before it reaches upjet, with the same
guard as `ServiceBinding` (#987). It is skipped while the CR is deleting, because
an `Observe` error returns before `Delete` runs and a bad annotation would
otherwise strand the finalizer.

The BTP plugin-framework provider is built once via `sync.OnceValue`, lazily so
that `btp.SetDebug` from `main()` still reaches HTTP tracing. The `ServiceManager`
and `CloudManagement` setups build their connectors at controller setup instead
of per reconcile, so the framework client's `OperationTrackerStore` survives
across reconciles. `resource-implementation-approaches.md` documents the hybrid
no-fork path and corrects the claim that the async callback carries
`LastAsyncOperation` onto the native CR: upjet keys that callback off the mapped
resource, so the name reaching `saveCallback` is `/TF-<cr-name>` and never
resolves the native CR. That defect predates this migration, since the CLI client
keys it the same way, and needs its own fix.

## Test plan

- [x] `make reviewable` with zero generated drift
- [x] Unit tests: framework-vs-CLI routing, disjoint external-name maps, the
      `QueryAsyncData` completion gate, tf-state identity recovery, and
      invalid-UUID rejection including the deleting case
- [ ] `Test_ServiceInstance_External_Name` green on Custom Upgrade Test Workflow

Closes #691


xdj2e9 (approved):
LGTM

I also ran the `Test_ServiceInstance_External_Name` upgrade test locally and it succeeded.

Issue #907, opened 2026-08-14
Adjust end user material for Destinations support


mpechkurov:
## Context
Part of epic #282 — Support Destinations TF Resource.

Update end user documentation to cover the new destination resources and guide users migrating from the standalone destination provider.

## Acceptance Criteria
- [ ] Update open source documentation
- [ ] Add migration guide from the destination provider to the new destination resources

GenosseOtt:
https://github.com/SAP/crossplane-provider-btp/pull/1007 ;-)

PR #1007, opened 2026-09-16
docs: add end-user guide for destinations and destination certificates

GenosseOtt:
## Summary

- Adds `docs/end-user-guides/account/destinations.mdx` — an end-user guide for the `SubaccountDestination` and `SubaccountDestinationCertificate` resources, which had no end-user documentation.
- Walks through the full Destination service dependency chain (Entitlement → ServiceInstance `offeringName: destination` → ServiceBinding secret) so users have the OAuth2 credentials the destination resources require.
- Documents creating an HTTP destination (with `additionalProperties` and the secret-based `additionalConfigurationSecretRefs`) and a destination certificate (base64 content secret, file-extension-driven format, FIPS caveat).
- Registers in the sidebar via `sidebar_position: 2` under **Account** (no `_category_.json` or index change needed); links to the existing external-name section for adoption rather than duplicating it.

Follows the existing guide conventions (prerequisites, save-then-apply steps, admonitions, CRD-reference buttons, FAQs).


xdj2e9:
@mpechkurov FYI. (partly) covers https://github.com/SAP/crossplane-provider-btp/issues/907

NVM just saw the comment on the issue 🧑‍🚀

SatabdiG (approved):

Issue #1006, opened 2026-09-16
Publish `v2.2.0`


xdj2e9:
## Payload since v2.1.0

- ServiceInstance migrated to the no-fork plugin-framework client (#1002)
- ServiceBinding migrated to the no-fork plugin-framework client (#1000)
- ServiceBinding external-name now fully ADR-compliant (#987)
- native Go implementation of SubaccountDestinationCertificate (#974)
- SAP/btp bumped to v1.26.0 (#993)

## Pre-release gate

- [x] Upgrade tests from [v2.1.0](https://github.com/SAP/crossplane-provider-btp/releases/tag/v2.1.0) pass on `main` ([run 35712389684](https://github.com/SAP/crossplane-provider-btp/actions/runs/35712389684))
- [x] Long / e2e tests pass on `main` ([run 35712376369](https://github.com/SAP/crossplane-provider-btp/actions/runs/35712376369))

## Publish

- [x] Run `Release` workflow against `main`, `nextVersion = v2.2.0`
- [x] Add to release channel

PR #1024, opened 2026-09-21
fix(servicebinding): if rotation is enabled persist the adapted name with suffix to status at create

xdj2e9:
Regression from #1000

## Root cause

When rotation is enabled we apply a random suffix to the binding name to keep each rotation unique. That means the `spec.name` (without suffix) will be different from the remote name (with suffix). Non-rotated bindings take the spec name as-is, so they never diff.

Before #1000 the "fork mode on disk state file" contained the last applied name with suffix so that a following plan operation produced no diff against the remote name and the resource reconciled normally.

After #1000 the actual `spec.name` is compared to the remote name without any knowledge of what has been applied thus producing a permanent plan diff so that the resource never can become ready.

## How was the regression merged?

`ServiceBinding` with rotation enabled is only tested by `TestServiceBinding_RotationLifecycle` which is part of the long E2E tests. They are only executed on main (daily) and before doing a release. That is how it was caught when preparing #1006.

## The fix

When rotation is enabled, persist the suffixed name directly during `Create` into `Status.AtProvider.Name`. On the following `Connect` it is read again and produces a plan without diff.

## Verified

Local run of `TestServiceBinding_RotationLifecycle` was successful.


internal/controller/account/servicebinding/servicebinding.go
```diff
@@ -285,6 +285,17 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
 		return managed.ExternalCreation{}, errors.Wrap(err, errCreateBinding)
 	}
 
+	// The rotated name carries a random suffix and the pending annotation is cleared above.
+	// Persist it to status.atProvider.name, the only durable record Connect and the key rotator read.
+	if e.isRotationEnabled(cr) {
```
Kukkerem: Isn't this evaluated after meta.RemoveAnnotations stripped ForceRotationKey, so it's effectively Spec.Rotation != nil? Maybe `if name != cr.Spec.ForProvider.Name` would be better.

Also the write runs after the pending-name annotation was cleared and persisted, so a crash in between, or a failed status update (retries only cover IsConflict) leaves the name recorded nowhere and Create never reruns. Shouldn't we move the block before SetExternalName?

internal/controller/account/servicebinding/servicebinding.go
```diff
@@ -285,6 +285,17 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
...
+		if err := reconcilerutil.UpdateStatusWithRetry(ctx, e.kube, cr, 3, func(cr *v1alpha1.ServiceBinding) error {
```
SatabdiG: nit: Should we do something about `3` retries. I see it also repeated in other places that the pr doesn't touch?

internal/controller/account/servicebinding/servicebinding.go
SatabdiG: I agree but maybe one thing worth covering in the same pass: `resolveCreateName's` adoption branch (adopted=true) returns out of Create() before this block runs at all, so a name persist there would need its own explicit write regardless of where we move this one — reordering alone won't close that path, would it?

xdj2e9: Good point. I moved it as suggested. Keeping the `isRotationEnabled` function as it clearly shows the intent.

internal/controller/account/servicebinding/servicebinding.go
xdj2e9: Yes I agree these magic numbers should be replaced. But I would like to keep it out of scope for this PR.

internal/controller/account/servicebinding/servicebinding.go
xdj2e9: > I agree but maybe one thing worth covering in the same pass: `resolveCreateName's` adoption branch (adopted=true) returns out of Create() before this block runs at all, so a name persist there would need its own explicit write regardless of where we move this one — reordering alone won't close that path, would it?

👍 

Nice find. Moving the write into `resolveCreateName` is even the better solution:
- it is explicitely for the rotation case
- it covers both paths automatically

As this method already had several side effects (annotation write + delete, external name write), I renamed it to reflect that better.

SatabdiG (approved):
lgtm 🔥

Kukkerem (approved):
LGTM

PR #1026, opened 2026-09-21
ci: refer to shared workflows repo

n-losse:
refer to shared workflows in https://github.com/SAP/crossplane-shared-workflows 

use main instead of pinning for now in order to keep maintenance effort low - can revisit this decision later if it becomes problematic.


Kukkerem (approved):
LGTM

SatabdiG (approved):

